PR #136, opened 2017-09-08
Utilities app correctly done?

nanbul98:
Just want to have you double check and see if I completed this correctly?

### DIFF
--- a/src/hackerspace_online/settings/base.py
+++ b/src/hackerspace_online/settings/base.py
@@ -99,6 +99,7 @@ INSTALLED_APPS = (
     'suggestions',
     'djcytoscape',
     'portfolios',
+    'utilities'
     # 'tours',
 )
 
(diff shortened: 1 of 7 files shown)


tylerecouture:
I don't think you committed the files for the Utilities app! only the models.py for quests is included here....?

tylerecouture:
Also, don't you need to update settings'py to include the new app?  Have you tested this? does it work?

tylerecouture:
Where's the utilities app itself though? as in :`src/utitlies` and `src/utitlies/models.py' etc etc? Those weren't added in any of these commits.

nanbul98:
is this good?

tylerecouture:
Sorry it took me so long to look at this =(.   The Quest model is an extension of the XPItem model which already has an icon attribute, so perhaps revert the Quest model and just leave that for now so I can accept the utilities app by itself and think about how to switch from the current icons to the new Utilities.icon model.

PR #147, opened 2017-10-22
reverting changes but adding utilities app

nanbul98:


tylerecouture:
Looking good. Can add back in your Utilities app's Icon model and do a new PR?

Issue #300, opened 2020-03-28
Proper authentication in postgres container


tylerecouture:
Currently using `POSTGRES_HOST_AUTH_METHOD=trust` to overcome the breaking change in the docker image: http://support.divio.com/en/articles/3719228-database-is-uninitialized-and-superuser-password-is-not-specified

But should do it properly.

tylerecouture:
Production should not be accepting all possible hosts: `ALLOWED_HOSTS = ['*']`
https://docs.djangoproject.com/en/3.0/ref/settings/#allowed-hosts

Should be `[.bytedeck.com]`

scssconsulting:
Please do that on prod settings on server if you need to. * is only on localhost

tylerecouture:
@scssconsulting can you explain that to me?  Django says "A value of '*' will match anything; "  Thanks!

tylerecouture:
Thanks, what about the main point of this issue in the first post?
```
    environment:
      POSTGRES_HOST_AUTH_METHOD: "trust"  # http://support.divio.com/en/articles/3719228-database-is-uninitialized-and-superuser-password-is-not-specified
```

scssconsulting:
@tylerecouture I am not too sure. This can give you more info: https://stackoverflow.com/questions/60262635/ci-cd-pipeline-with-postgresql-failed-with-database-is-uninitialized-and-superu

tylerecouture:
"ALLOWED_HOSTS = ['*']  This value is not safe for production usage." - from https://docs.docker.com/compose/django/

Changed on server.

PR #321, opened 2020-03-29
Slight adjustment

jeffdico:
adjusted the docker compose file and the environment variables to ensure that access to the db is not based on trust, but enforces authentication.
Once the db volume is recreated this will be enforced by default.
#300


tylerecouture:
Thanks for this!  Can you provide a suggestion (in code comments), as to how production environment should be set up to provide the pw?

tylerecouture:
Appreciate the work here.  Was this a one time contribution?  Major overhaul of app in the last week.

PR #322, opened 2020-03-29
Require access code signup

hafizSiddiq7675:
For course code in sign up form.


tylerecouture:
Do tests locally before PR please.
 `manage.py test src && flake8 src`

PR #323, opened 2020-03-30
Fix page reload when clicking permalink

### LINKED ISSUE #284: Fix awkward behaviour when clicking announcement permalink
Should just copy link to clipboard.  Find a JS library?

yujinyuz:
- [x] Changed behavior to copy link to clipboard
- [x] Use `build_absolute_uri` or something similar

Fixes #284


tylerecouture:
Thanks!  Can you update the icon to a copy icon instead of a link icon?

PR #328, opened 2020-03-31
Add Basic and Advanced sections in QuestForm

### LINKED ISSUE #326: Create "Advanced" section in the Quest edit/create form
Currently, there are too many fields on the form, and many of them are only used in special cases.

There should be a hidden, expandable "Advanced" section.  (Can Crispy Forms handle this already?)

Fields on the form should be in this order:

- Name
- XP
- Visible
- Verification Requried
- Icon
- Short Description
- Quest Details
- Submission Instructions
- Instructor Notes
- Campaign
- Common Data
- Max Repeats
- Hours Between repeats

--== ADVANCED SECTION ==--
All remaining fields currently on form

yujinyuz:
Couldn't find other alternatives for separating sections in the
django-crispy-forms documentation. Tabs were the closest thing I could
find.

Still needs some fixes in styling

Screen Shot 2020-03-31 at 9 47 34 PM

Screen Shot 2020-03-31 at 9 47 09 PM

What do you think @tylerecouture ?

Closes #326

### DIFF
--- a/src/quest_manager/forms.py
+++ b/src/quest_manager/forms.py
@@ -1,7 +1,7 @@
 from bootstrap_datepicker_plus import DatePickerInput, TimePickerInput
-from crispy_forms.bootstrap import Tab, TabHolder
+from crispy_forms.bootstrap import Accordion, AccordionGroup
 from crispy_forms.helper import FormHelper
-from crispy_forms.layout import Layout
+from crispy_forms.layout import Layout, Div
 from django import forms
 from django_select2.forms import ModelSelect2MultipleWidget
 from django_summernote.widgets import SummernoteInplaceWidget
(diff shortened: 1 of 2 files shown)


yujinyuz:
> I would prefer the Accordion & AccordionGroup layout. Can you give that a try?

I did see that but I thought it was just the same with fieldset. I didn't know it was clickable! Anyways, I've made the changes

tylerecouture:
Yeah! That's great.  Can you make it so that only the Advanced section is an Accordian?  The Basic section should not be hideable, and preferable doesn't have a header at all, its just like it is now, and the Advanced section looks like how you've got it in the screencap

yujinyuz:
Kapture 2020-04-01 at 0 42 53

PR #341, opened 2020-04-02
Fix incorrect copy link and change copy icon

### LINKED ISSUE #284: Fix awkward behaviour when clicking announcement permalink

yujinyuz:
* Used trigger.origin since it returns the base path of the URL which is what we need
  since `object.get_absolute_url` only returns `/path` instead of
  complete path
* Changed icon from fa-copy to fa-clipboard

Closes #284

### DIFF
--- a/src/announcements/templates/announcements/list.html
+++ b/src/announcements/templates/announcements/list.html
@@ -20,7 +20,12 @@
       });
     });
 
-    var clipboard = new ClipboardJS(permalinks);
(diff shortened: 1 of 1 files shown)


yujinyuz:
Not sure what you were experiencing but it's working fine on my end. I'm currently checked out in  beta-features-pre-multitenant branch

Kapture 2020-04-06 at 23 54 47

PR #351, opened 2020-04-05
Use timezone aware date widgets

### LINKED ISSUE #261: Use time zone aware date widget instead of local date
Currently using local dates `datetime.date.today()` instead of UTC dates `timezone.now().date()` but this doesn't work with current date widget, because date widget doesn't convert UTC date to local time, causing confusion.

Ideally, widget would display/read local date and convert to UTC before saving.

#89 and #260

yujinyuz:
Tried passing `options` to form widgets with `defaultDate` (based on
bootstrap-datepicker-plus docs) but `defaultDate` is being ignored when
`default` parameter is set in models.DateTimeField

Opted to override initial values of DateTime fields when form is instantiated

Closes #261

### DIFF
--- a/src/announcements/forms.py
+++ b/src/announcements/forms.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from bootstrap_datepicker_plus import DateTimePickerInput
 from django import forms
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


tylerecouture:
Can you change your upstream branch to ` beta-features-pre-multitenant `? We're going to use this branch for now until some major changes are done on the develop branch.

PR #357, opened 2020-04-06
deployment changes

arneec:


src/hackerspace_online/settings/base.py
```diff
@@ -546,3 +546,7 @@ def POSTMAN_NOTIFICATION_APPROVAL(u): return u.profile.get_messages_by_email
 TENANT_MODEL = "tenant.Tenant"
 
 DEFAULT_FILE_STORAGE = 'tenant_schemas.storage.TenantFileSystemStorage'
+
+
+DEFAULT_TENANT_SUPERUSER_USERNAME = 'admin'
+DEFAULT_TENANT_SUPERUSER_PASSWORD = 'admin1234'
```
tylerecouture, line 552: Why is this in the base settings file?  This should be in the local settings file as it is specific to local development and will definitely not be common between local vs production.

tylerecouture: Also, settings should begin with their qpp name, so this should be TENANT_DEFAULT_SUPERUSER_USERNAME for example.

SERVER-README.md
```diff
@@ -18,4 +18,9 @@
 ### Workflow
 - ``uwsgi`` is used to run the django application inside docker container
 - ``nginx`` reverse proxy is configured to point the docker django application mapped port
-- static and media files are served by ``nginx`` itself by pointing the url to the static and media directory of the application
+- static and media files are served by ``nginx`` itself by pointing the url to the static and media directory of the application ie. ``/src/static_in_project/static_root`` and ``/src/static_in_project/media_root``
```
tylerecouture, line 21: Static and media files should not be served from the project directly.  They should be somewhere else.  /var/www/bytedeck/media, or something.  This should be indicated in the production settings file.  Collectstatic will place the files where they need to go, right?

src/initial_data.json
```diff
@@ -3,8 +3,8 @@
     "model": "sites.site",
     "pk": 1,
     "fields": {
-      "domain": "localhost",
-      "name": "localhost"
+      "domain": "bytedeck.com",
+      "name": "bytedeck.com"
```
tylerecouture, line 7: This should stay as localhost?  I'm not worried about setting up other production environments, the app should work as close as possible "out of the box" to set up a development environment.  The production environment is the special case where we can manually change values, the defaults should be for development environment and then overridden in prorduction (via the production settings file, .env, or otherwise)

tylerecouture: This will never get used again on this production environment, once it is run the first time, right?

src/tenant/fixtures/tenants.json
```diff
@@ -3,7 +3,7 @@
     "model": "tenant.tenant",
     "pk": 1,
     "fields": {
-      "domain_url": "localhost",
+      "domain_url": "bytedeck.com",
```
tylerecouture, line 6: leave as localhost or even better, 127.0.0.1, because that's easier for developers, and also where django tells you to go when you runserver

SERVER-README.md
```diff
@@ -0,0 +1,21 @@
...
+
+### Stack
+- nginx
+- uwsgi
+- docker (docker-compose)
+- git
+
+### Steps
+- configure nginx in ``/usr/share/nginx/hackerspace.conf``
+- go to application directory ``/home/xillar/webapp/hackerspace``
```
tylerecouture: wrong directory?

src/hackerspace_online/settings/local.py
```diff
@@ -63,12 +63,12 @@
 
 # The absolute path to the directory where collectstatic will collect static files for deployment.
 # Set in production settings for deployment
-#STATIC_ROOT = os.path.join(BASE_DIR, "static_in_project",  "static_root")
+STATIC_ROOT = os.path.join(BASE_DIR, "static_in_project",  "static_root")
```
tylerecouture, line 66: Why changing my development set up?

src/hackerspace_online/urls.py
```diff
@@ -77,8 +77,8 @@
 ]
 
 if settings.DEBUG:
-    urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
-    urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
+    # urlpatterns += static(settings.STATIC_URL, document_root=settings.STATIC_ROOT)
+    # urlpatterns += static(settings.MEDIA_URL, document_root=settings.MEDIA_ROOT)
```
tylerecouture, line 80: Why this change?

Dockerfile
```diff
@@ -62,5 +66,5 @@ ENV SHELL /bin/bash
 
 #######################################################################
 
-# RUN adduser --disabled-password appuser
-# USER appuser
+RUN useradd -M xillar
+USER xillar
```
tylerecouture: why adding this user?

tylerecouture:
I left some review comments.  The repo shouldn't have any reference to my specific deployment of this app.  Any deployment-specific settings need to be on a file on the production server and read from there (.env or the production settings file).

the base.py settings file should only include settings that are common to local development and the production settings file, otherwise they should be broken apart between them.

My suggestions anyway, if there is a specific reason to do it another way, please discuss with me. Thanks!

tylerecouture:
In fact, should we remove the production settings file "template" in the repo to avoid confusion?

arneec:
> I left some review comments. The repo shouldn't have any reference to my specific deployment of this app. Any deployment-specific settings need to be on a file on the production server and read from there (.env or the production settings file).
> 
> the base.py settings file should only include settings that are common to local development and the production settings file, otherwise they should be broken apart between them.
> 
> My suggestions anyway, if there is a specific reason to do it another way, please discuss with me. Thanks!

I have fixed everything you have pointed out.

arneec:
> In fact, should we remove the production settings file "template" in the repo to avoid confusion?

It would be easier to copy and modify it if its there. You can rename it to production.template.py something like this.

PR #358, opened 2020-04-06
Fix students cant drop submission

### LINKED ISSUE #254: Student's can't drop submissions from quests that have become not visible
https://hackerspace.sd72.bc.ca/media/attachments/postman_message/316/Screenshot_from_2019-12-12_12-34-50.png

If a student has submitted a quest, and then that quest is made not visible, the submission stull appears for them, but clicking it leads to error page

yujinyuz:
Closes #254

### DIFF
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -162,7 +162,6 @@ class SubmissionViewTests(TestCase):
         self.assertTrue(success)
 
         s1_pk = self.sub1.pk
-        s2_pk = self.sub2.pk
 
         # Student's own submission
         self.assertEqual(self.client.get(reverse('quests:submission', args=[s1_pk])).status_code, 200)
@@ -253,7 +252,6 @@ class SubmissionViewTests(TestCase):
 
         # Should be able to see own submission even when quest is closed
         self.assertEqual(self.client.get(reverse('quests:submission', args=[self.sub1.pk])).status_code, 200)
-
 
     def test_ajax_save_draft(self):
         # loging required for this view
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -120,6 +120,7 @@ class SubmissionViewTests(TestCase):
 
     def setUp(self):
         User = get_user_model()
+        self.sem = SiteConfig.get().active_semester
 
         # need a teacher and a student with known password so tests can log in as each, or could use force_login()?
         self.test_password = "password"
@@ -132,9 +133,27 @@ class SubmissionViewTests(TestCase):
         self.quest1 = mommy.make(Quest)
         self.quest2 = mommy.make(Quest)
 
-        self.sub1 = mommy.make(QuestSubmission, user=self.test_student1, quest=self.quest1)
-        self.sub2 = mommy.make(QuestSubmission, quest=self.quest1)
-        self.sub3 = mommy.make(QuestSubmission, quest=self.quest2)
+        self.sub1 = mommy.make(
+            QuestSubmission,
+            user=self.test_student1,
+            quest=self.quest1,
+            is_completed=True,
+            semester=self.sem
+        )
+
+        self.sub2 = mommy.make(
+            QuestSubmission,
+            quest=self.quest1,
+            is_completed=True,
+            semester=self.sem
+        )
+
+        self.sub3 = mommy.make(
+            QuestSubmission,
+            quest=self.quest2,
+            semester=self.sem,
+            is_completed=False
+        )
 
     def test_all_submission_page_status_codes_for_students(self):
 
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -132,27 +132,17 @@ class SubmissionViewTests(TestCase):
 
         self.quest1 = mommy.make(Quest)
         self.quest2 = mommy.make(Quest)
+        self.quest3 = mommy.make(Quest)
 
-        self.sub1 = mommy.make(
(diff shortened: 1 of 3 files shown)


src/quest_manager/tests/test_views.py
```diff
@@ -161,13 +179,6 @@ def test_all_submission_page_status_codes_for_students(self):
         self.assertEqual(self.client.get(reverse('quests:unflag', args=[s1_pk])).status_code, 302)
         self.assertEqual(self.client.get(reverse('quests:complete', args=[s1_pk])).status_code, 404)
 
-        # Not this student's submission
-        self.assertEqual(self.client.get(reverse('quests:submission', args=[s2_pk])).status_code, 302)
-        self.assertEqual(self.client.get(reverse('quests:drop', args=[s2_pk])).status_code, 302)
-        self.assertEqual(self.client.get(reverse('quests:skip', args=[s2_pk])).status_code, 404)
-        self.assertEqual(self.client.get(reverse('quests:submission_past', args=[s2_pk])).status_code, 302)
-        self.assertEqual(self.client.get(reverse('quests:complete', args=[s2_pk])).status_code, 404)
-
```
tylerecouture, line 170: Why deleting all these tests?

src/quest_manager/tests/test_views.py
```diff
@@ -132,9 +133,27 @@ def setUp(self):
...
+            is_completed=True,
+            semester=self.sem
+        )
+
+        self.sub3 = mommy.make(
+            QuestSubmission,
+            quest=self.quest2,
+            semester=self.sem,
+            is_completed=False
+        )
```
tylerecouture, line 156: I don't think you should edit the general setup for one specific test?  Is this why you needed to delete all those other tests, because changing the setup causes the tests to fail?  In general, never mess with existing tests.  Just add, add, add.  There is rarely a need to REMOVE a working test.

src/quest_manager/views.py
```diff
@@ -930,17 +931,20 @@ def drop(request, submission_id):
...
+        CustomSubmissionForm = SubmissionFormStaff
     else:
-        initial = {'comment_text': sub.draft_text}
-        main_comment_form = SubmissionForm(request.POST or None, initial=initial)
+        qs = QuestSubmission.objects.all_completed(request.user)
+        sub = get_object_or_404(qs, pk=submission_id)
+        initial['comment_text'] = sub.draft_text
+        CustomSubmissionForm = SubmissionForm
+
+    main_comment_form = CustomSubmissionForm(request.POST or None, initial=initial)
```
tylerecouture, line 947: Please add comments explaining these code changes.  It's not obvious to me what's going on here and I don't want to waste tons of time trying to parse the code.  More code comments everywhere please!

src/quest_manager/tests/test_views.py
yujinyuz: Sorry about that. Though I added comments inside commit's description.

https://github.com/timberline-secondary/hackerspace/pull/358/commits/5a3df9d2001f5eccce9126904918e8d5f2637821

Screen Shot 2020-04-07 at 8 10 09 AM

I deleted these lines since QuestSubmissions are already filtered by all submissions of that student via `qs = QuestSubmission.objects.all_completed(request.user)`

```python
@login_required
def submission(request, submission_id=None, quest_id=None):


    initial = {}


    if request.user.is_staff:
        qs = QuestSubmission.objects.all_completed()
        sub = get_object_or_404(qs, pk=submission_id)
        CustomSubmissionForm = SubmissionFormStaff
    else:
        qs = QuestSubmission.objects.all_completed(request.user)
        sub = get_object_or_404(qs, pk=submission_id)
        initial['comment_text'] = sub.draft_text
        CustomSubmissionForm = SubmissionForm
```

It was causing the test to fail since it was expecting a redirect to available quests. 

```python
    if sub.user != request.user and not request.user.is_staff:
        return redirect('quests:quests')
```

The current implementation throws a 404. I also deleted the line above since the check wasn't necessary anymore

src/quest_manager/tests/test_views.py
yujinyuz: I changed the general setup coz `QuestSubmission.objects.all_completed(user)` doesn't return anything as it wasn't assigned to a semester and the quest wasn't complete. Although I've only used `self.sub1` in the test cases.

Should I revert it back and just add another QuestSubmission dummy data?

src/quest_manager/views.py
yujinyuz: I'll add comments to it but what I was trying to do here was that I'm pre-filtering quest submissions. If it's a staff then it should fetch all completed submissions. And it should also use the SubmissionFormStaff form that will be used for instantiating later.

I did that instead of having this condition

```python
if request.user.is_staff:
    main_comment_form = SubmissionFormStaff(request.POST or None)
else:
    main_comment_form = SubmissionForm(request.POST or None, initial=initial)
```

src/quest_manager/tests/test_views.py
tylerecouture: Yes, I would prefer that.  Add your own test methods as needed and leave those that are already there and working (not broken)

tylerecouture: Also, don't change the setup, as all the other tests assume that data is set up the way it is.  It might not be a problem, but I'd rather not change working tests at all, just add to them with your own. and don't edit setup.  If you want a different setup, then make a new test class with your own setup.

src/quest_manager/tests/test_views.py
tylerecouture: Ok, just change the test to check for 404 then, that's ok.

PR #410, opened 2020-04-10
Update admin site heading

### LINKED ISSUE #370: Fix admin site heading so it doesn't say Hackerspace
Should use the name of the current tenant's site (from SiteConfig?)

yujinyuz:
Closes #370


tylerecouture:
Oh shoot, can you fix this?  don't use `SiteConfig.objects.first()`

Use `SiteConfig.get()` This is a class method I added that does the same thing, but also checks if the config exists yet if it doesn't, creates it (also, if we always get the config from there it will be easier for us to update later)

yujinyuz:
@tylerecouture Sure. Will use that. I didn't see that tho

PR #413, opened 2020-04-11
Server read me updated

scssconsulting:


tylerecouture:
Is this what I should expect?
````
$ export WUID=33
$ export WGID=33
$ sudo docker-compose up
WARNING: The WUID variable is not set. Defaulting to a blank string.
WARNING: The WGID variable is not set. Defaulting to a blank string.
Creating network "hackerspace_default" with the default driver
Creating hackerspace_redis_1 ... 
Creating db ... 
````

PR #414, opened 2020-04-11
Use Site.get() instead of Site.objects.first()

yujinyuz:
Sorry about the unnecessary changes. Just tried using a linter


tylerecouture:
No prob! What linter are you using?

yujinyuz:
I think it was autopep8

PR #415, opened 2020-04-11
add dev.yml and fix celery issue

it-avenger:
You should change postgres db name. Please check .env file.
Please follow the stesp.
1. `docker-compose -f dev.yml up db redis web`
2. `docker-compose -f dev.yml up`


tylerecouture:
Fails tests.

PR #448, opened 2020-04-17
Move .flake8 file to root folder

yujinyuz:
autopep8 and other linters looks for files under the root folder.


tylerecouture:
Looks like we need to adjust travis.yml to find it now.

PR #463, opened 2020-04-22
Redirect student if quest is in progress

### LINKED ISSUE #455: Redirect student to in-progress quest if they try to start it again
There is a user error that is common enough that we should handle it:

A student starts a quest, then for some reason hits back on their browser,  and then tries to start the quest again.  But since it's already been started (and appearing in their in progress quests), they get a 404 error.

Instead, if they try to start (via "/quests/23/start/" for example) a quest a second time, instead it should just redirect them to the already in-progress quest.

yujinyuz:
Closes #455

### DIFF
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -772,6 +772,12 @@ def complete(request, submission_id):
 @login_required
 def start(request, quest_id):
     quest = get_object_or_404(Quest, pk=quest_id)
+    quest_submission = (QuestSubmission.objects.all_not_completed(user=request.user)
+                        .filter(quest_id=quest_id)).first()
+
+    if quest_submission:
+        return redirect(reverse('quests:submission',
+                                kwargs={'submission_id': quest_submission.id}))
 
     if not quest.is_available(request.user):
         raise Http404


yujinyuz:
@tylerecouture Yes. I am adding tests. I'm still trying to figure out how to add a quest in the current semester since `QuestSubmission.objects.not_all_completed` returns an empty queryset.

tylerecouture:
Under what conditions does that happen?

yujinyuz:
> Under what conditions does that happen? Do you mean during a test?

Yes during test.  I think a 404 happens in the test case since a quest isn't assigned in a particular semester

tylerecouture:
Ah, OK.  How about I'll figure out the test for this one and you can use it for future reference.

yujinyuz:
Sure @tylerecouture thanks!

tylerecouture:
@yujinyuz 

A few notes for you on this PR:
1. [redirect](https://docs.djangoproject.com/en/3.0/topics/http/shortcuts/#redirect) accepts a model object, so you don't have to go through all that convoluted code to get redirect to work! haha, just go `redirect(quest_submission)` and it will work as long as `get_absolute_url()` is defined on the model, which it is.  I probably have some of that convoluted code elsewhere myself, but it's not necessary (feel free to fix if you find!)... I even do it slightly further down in the same method!
2. If you look further down the code in this method, you can see that there was already an attempt to solve this problem (apparently from me 5 years ago according to git blame!)
```
    new_sub = QuestSubmission.objects.create_submission(request.user, quest)
    # if new_sub is None:
    #     print("This quest is not available, why is it showing up?")
    #     raise Http404 #shouldn't get here
    # should just do this, but the redirect is screwing up with the tour...
    # return redirect('quests:submission', submission_id = new_sub.id)

    if new_sub is None:  # might be because quest was already started
        # so try to get the started Quest
        sub = QuestSubmission.objects.all_for_user_quest(request.user, quest, True).last()
        if sub is None:
            raise Http404
    else:
        sub = new_sub
```

And in the end, should redirect to that submission anyway.  So why doesn't it do it?  I'll fix it up as I write the test for this method.  I  don't have a thorough test suite for the views that cehcks all the possible logic paths, so this will be the start!  I'm going to comment it thoroughly so hopefully you can more easily start adding appropriate tests for each of your PRs

PR #465, opened 2020-04-22
Fix error when commenting on flagged comments

### LINKED ISSUE #459: Trying to reply after a comment is flagged is banned gets error
> I get the error (image attached) when I am logged in as a student trying to reply to the instructor who flagged my comment, asking why it was flagged.

image

yujinyuz:
It seems that this problem has been existing for years now. Took me a while to figure out, too!

Closes #459


tylerecouture:
Tests! =D

PR #478, opened 2020-04-25
Use apply_async for calculating auto-approved

### LINKED ISSUE #473: Submitting a quest that is auto-approved doesn't use celery
Need to figure out a better way to handle auto-approved quests (`Quest.verification_required == False`)

They are set to NOT use celery, because students are expecting an immediate update to their available quests, based on this one being accepted.

If we use celery, then it will show as submitted, but they'll have to wait for the asynch celery task to finish before the new quest appears, which is confusing.  We NEED to use celery, because it takes too long otherwise.

Here's what I think we should do:
In the "successfully sent message" that appears at the top of the next page that loads, add this additional information:

"This quest was automatically approved, please give me a moment to calculate what new quests this should make available to you!  Try refreshing your browser in a few moments. Thanks! - {{ config.deck_ai }}"  (not sure that's the right way to call the config.deck_ai)

Then, set these quests to use celery just like all the other quests.

yujinyuz:
Fixes #473

Screen Shot 2020-04-25 at 6 40 31 PM

### DIFF
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -708,6 +708,9 @@ def complete(request, submission_id):
                     note_verb += ", awaiting approval."
                 else:
                     note_verb += " and automatically approved."
+                    note_verb += " Please give me a moment to calculate what new quests this should make available to you."
+                    note_verb += " Try refreshing your browser in a few moments. Thanks! <br>&mdash;{deck_ai}"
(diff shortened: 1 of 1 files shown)


src/quest_manager/views.py
```diff
@@ -752,6 +755,7 @@ def complete(request, submission_id):
                 verb=note_verb,
                 icon=icon,
             )
+            print(note_verb)
```
tylerecouture, line 758: remove print()

PR #480, opened 2020-04-25
Send comment notif when quest is auto-approved

### LINKED ISSUE #180: Notify teacher when other teacher comments
When another teacher comments on a student quest, notify the student's main teacher

### LINKED ISSUE #476: Teacher should get notification when a student comments on an auto-approved quest
Currently, when a student submits a quest that does not require approval, the teacher is not notified.

If the student types something, the teacher should receive a notification about the comment like any other quest.

If the student does not comment on an auto-approved quest, then there is no need for a notification.

yujinyuz:
Fixes #476 
Fixes #180

### DIFF
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -724,8 +724,8 @@ def complete(request, submission_id):
                         and submission.quest.specific_teacher_to_notify not in request.user.profile.current_teachers():
                     affected_users = [submission.quest.specific_teacher_to_notify, ]
                 else:
-                    # Send notification to current teacher when a quest is auto-approved
-                    if comment_text:
+                    # Send notification to current teacher when a quest is auto-approved. Don't send when there are no comments
+                    if comment_text != '(submitted without comment)':
                         affected_users = [request.user.coursestudent_set.first().block.current_teacher]
 
                 submission.mark_completed()
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -492,6 +492,12 @@ def approve(request, submission_id):
                 action = comment_new
 
             affected_users = [submission.user, ]
+            main_teacher = submission.user.coursestudent_set.first().block.current_teacher
+
+            # Only notify main teacher if the user who commented is another teacher
+            if 'comment_button' in request.POST and request.user != main_teacher:
+                affected_users.append(main_teacher)
+
             notify.send(
                 request.user,
                 action=action,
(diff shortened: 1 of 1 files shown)


yujinyuz:
Should merge this after I figure out https://github.com/timberline-secondary/hackerspace/issues/180

yujinyuz:
Would it be better to write tests for this?

If so, how do I test for notifications?

src/quest_manager/views.py
```diff
@@ -716,7 +724,10 @@ def complete(request, submission_id):
                         and submission.quest.specific_teacher_to_notify not in request.user.profile.current_teachers():
                     affected_users = [submission.quest.specific_teacher_to_notify, ]
                 else:
-                    affected_users = None
+                    # Send notification to current teacher when a quest is auto-approved
+                    if comment_text:
```
tylerecouture, line 728: Did you manually test this?  Because I think our summernote widget (the comment_text field widget) automatically adds `<p></br></p>` to every one of its fields (at least that's what you see when you are start editing text in one of the widget in html view.  Myabe summernote strips that if nothing is entered?  Just want to make sure before I accept the PR, sinse we don't have a unit test yet for this.

yujinyuz: You're right! `comment_text` always has a value of `(submitted without comment)`. I was wondering if it would be better to just check for the `form.cleaned_data['comment_text']` since it returns an empty string when there are no comments. Coz what I did what I just compared `comment_text != '(submitted without comment)'`

yujinyuz:
django-summernote handles the empty strings

https://github.com/summernote/django-summernote/blob/06d01388132b038089bf73891ea4ec87bb3b3aa7/django_summernote/apps.py#L24-L25

But we can't check if the user has just entered empty lines. It will result to something like this

```
'<p><br></p><p><br></p><p><br></p><p><br></p><p><br></p><p><br></p><p><br></p><p><br></p>'
```

And it's quite complicated

yujinyuz:
@tylerecouture Was there any reason most of the stuffs in `notifications/tests/test_views.py` are commented out?

I was about to add the tests there but I thought it might not be the right place since most of them were commented out

tylerecouture:
Ok, forget about unit tests (I'll look into that and fix the notification tests)

Let's just handle the basic case for now, where a student enters nothing.

tylerecouture:
Haha, looks like I just copy and pasted the tests from the badges app and never got around to doing them =D

Issue #405, opened 2020-04-10
django-tenant-schemas vs django-tenants?


yujinyuz:
django-tenants seems to be more actively maintained compared to django-tenant-schemas.

Plus, django-tenants is a fork of django-tenant-schemas, looking at their documentation credits, it says:

> I would like to thank the original author of this project Bernardo Pires Carneiro under the name django-tenant-schemas. I forked this project as I wanted to add enhancements to make it work better with Django 1.8
If you are using Django 1.7 or below please use his project.

Should we consider using `django-tenants` instead?

scssconsulting:
django-tenant-schemas is still an active project. It has more stars than django-tenants. django-tenant-schemas supports django 2.0+ as well. I wouldn't suggest jumping to django-tenants unless you have any specific reason. What are the features that you can't find in django-tenant-schemas but on django-tenant?

yujinyuz:
I didn't see anything special in django-tenants. It was just that django-tenant-schemas' last commit was November.

Anyway, can close this since

tylerecouture:
https://github.com/tomturner/django-tenants/issues/372

tylerecouture:
Closed by #906

PR #486, opened 2020-05-03
[WIP] Switch to django-tenants

yujinyuz:
@tylerecouture I've managed to make it work. The `django_cache` problem came from the `select2` key in `CACHES`. I don't have any ideas why but commenting it out or replacing it to use redis made the test suite work again.

Tests aren't passing right now. `django-tenants` seem to have a different structure. A `Tenant` can have one or multiple `Domains`.

In order to get `domain_url`, we'd have to change it to `self.domains.first().domain`.
https://github.com/timberline-secondary/hackerspace/blob/f4d06dba1b6112eeba9fd30569ae6849dd270394/src/tenant/models.py#L33-L36

 I didn't do that yet as I also want to hear your thoughts about this.

Also, the `tenant/tests/test_admin.py`, I'm not quite sure how to test it given that domain_url isn't part of the `Tenant` model anymore.


tylerecouture:
#405

PR #493, opened 2020-05-06
Use docker-compose run instead of exec

yujinyuz:
When following the documentation under Running Code > Initial Setup
Number 6, running bash init_public_schema.sh gives an error:

```
ERROR: No container found for web_1
```

It seems that exec command requires the `web_1` to be run before we can
use it.


yujinyuz:
While I was doing the README, I've encountered an error so I just made a PR instead in case someone else would work on the project

PR #501, opened 2020-05-09
Fix/redirect to login page instead of home

### LINKED ISSUE #489: Redirect to login page instead of home or admin when you are not logged in
Currently redirects to home page when trying to access a view that requires you to be logged in, should redirect to login page instead.

Also, when accessing a staff-only view, it currently redirects to admin page, should also just go to standard login page instead.

Tests that use RedirectsHome should be replaced with a new utility test method `RedirectsLogin`

yujinyuz:
Fixes #489

### DIFF
--- a/src/badges/views.py
+++ b/src/badges/views.py
@@ -2,15 +2,16 @@ from django.contrib import messages
 from django.contrib.admin.views.decorators import staff_member_required
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.models import User
-from django.shortcuts import render, get_object_or_404, redirect
+from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse_lazy
 from django.utils.decorators import method_decorator
 from django.views.generic.edit import DeleteView, UpdateView
 
 from notifications.signals import notify
-from tenant.views import allow_non_public_view, AllowNonPublicViewMixin
-from .forms import BadgeForm, BadgeAssertionForm, BulkBadgeAssertionForm
-from .models import Badge, BadgeType, BadgeAssertion
+from tenant.views import AllowNonPublicViewMixin, allow_non_public_view
+
+from .forms import BadgeAssertionForm, BadgeForm, BulkBadgeAssertionForm
+from .models import Badge, BadgeAssertion, BadgeType
 
 
 @allow_non_public_view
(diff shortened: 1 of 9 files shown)


src/badges/views.py
```diff
@@ -61,9 +62,22 @@ def dispatch(self, *args, **kwargs):
...
 def badge_create(request):
+
+    # Using `@staff_member_required(login_url='/accounts/login/')` causes a RedirectCycleErrror.
+    # For example, a student that is already logged in tries to access this page,
+    # They will get redirect to `/accounts/login/?next=/achievements/create/`.
+    # And since they are already logged in, they will be redirected to `/achievements/create/`
+    # Causing again to redirect since they are not a staff.
+
+    # We could use `@staff_member_required(login_url='/')` but this breaks the tests
+    # which requires views to be redirected to /accounts/login
```
tylerecouture, line 75: What happens if we just use `@staff_member_required`?  What are we doing elsewhere when we use this decorator?  This view should act the same as everywhere else that uses the same decorator.  And `@login_required` won't work because then students can access this view.

src/hackerspace_online/tests/utils.py
```diff
@@ -28,6 +29,12 @@ def assertRedirectsHome(self, url_name, *args, **kwargs):
             expected_url='%s?next=%s' % (reverse('home'), reverse(url_name, *args, **kwargs)),
         )
 
+    def assertRedirectsLogin(self, url_name, *args, **kwargs):
+        self.assertRedirects(
```
tylerecouture, line 33: Please add docstring to the method (see other methods in this class)

src/badges/views.py
yujinyuz: With `@staff_member_required`, students that are already logged in are redirected to `/accounts/login/?next=/achievements/create/`, and since they are already logged in, they get redirected to `/achievements/create/` which is has the `@staff_member_required`, so they would get redirected again to `/accounts/login/?next=/achievements/create/`. And it creates a loop...

tylerecouture: Won't that be a problem for ALL @staff_member_required views?  Maybe what we need is our own decorator that checks if they are already logged in as a student, then they get 403 Permission Denied, and if they are NOT logged in, then redirect to login page.

tylerecouture: For now, let's just leave it to redirect to django admin log in page.

yujinyuz: > Won't that be a problem for ALL @staff_member_required views?

For the case of creating badges, I don't think it is. It's the only case that I saw that redirects to `/`, other @staff_member_required redirects to admin.

> Maybe what we need is our own decorator that checks if they are already logged in as a student, then they get 403 Permission Denied, and if they are NOT logged in, then redirect to login page.

I agree and I think that would be better to make a custom decorator. I do think that @staff_member_required should only be used for things that can only be done via admin.

PR #578, opened 2020-07-30
Django tenants refactor

stanvanrooy:
This moves the application from the no longer maintained django-tenants-schemas to django_tenants. The main difference is that the `domain_url` property moved to the new `Domain` model. One tenant can have multiple domains associated.

I've added a getter/setter to the `Tenant` model for the domain url, for this I've assumed each tenant will have only one domain. @tylerecouture is this correct?


tylerecouture:
As discussed, please squash some of those commits and add in the migration or handling of existing tenants. Thanks!

PR #590, opened 2020-08-05
Add condition for displaying marks

### LINKED ISSUE #544: Change how current mark displays
See screenshot for how current mark displays.
Change so that default is to show XP only.
When mark calculation is turned on, show XP and % mark.
quest_marks

yujinyuz:
Closes #544

### DIFF
--- a/src/comments/templates/comments/comments.html
+++ b/src/comments/templates/comments/comments.html
@@ -27,8 +27,8 @@
         <h3 class="comment-heading">
           {% if request.user.is_staff %}
           <a href="{% url 'profiles:profile_detail' comment.user.profile.id %}">{{comment.user}}</a>
-          <small>({{comment.user.profile.preferred_full_name}}
-          Current Mark: {{comment.user.profile.mark|floatformat:0}}%)
+          <small>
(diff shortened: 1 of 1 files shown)


src/comments/templates/comments/comments.html
```diff
@@ -27,8 +27,8 @@
         <h3 class="comment-heading">
           {% if request.user.is_staff %}
           <a href="{% url 'profiles:profile_detail' comment.user.profile.id %}">{{comment.user}}</a>
-          <small>({{comment.user.profile.preferred_full_name}}
-          Current Mark: {{comment.user.profile.mark|floatformat:0}}%)
+          <small>
+            ({{comment.user.profile.preferred_full_name}}{% if config.display_marks_calculation %} Current Mark: {{comment.user.profile.mark|floatformat:0}}%{% endif %})
```
yujinyuz, line 31: Combined into a single line since it creates an extra whitespace
Screen Shot 2020-08-06 at 7 53 00 AM

tylerecouture: There should be an `else` block that displays `profile.xp_per_course` otherwise.

PR #648, opened 2020-10-07
Migrate django-tenant-schemas to django-tenants

yujinyuz:
* FIxed tests to be compatible with django-tenants
* Migrate `Tenant.domain_url` to `TenantDomain.domain`

Closes #405 
Closes #384

PR #906, opened 2022-01-27
Migrate django-tenant-schemas to django-tenants

yujinyuz:
Original PR from #648


yujinyuz:
@tylerecouture Based on the previous comments, the only blockers that the previous PR had before was a proper staging environment.

Do we have that now?

tylerecouture:
I started setting that up this week. Hoping to have it completed in a couple days.  But I'm going to accept this so you can carry on with the django 3.2 work with hopefully less conflicts.

PR #947, opened 2022-03-07
Prevent student from starting a quest without course

### LINKED ISSUE #892: New students can start Welcome quest before joining a course by accessing via maps
See example here:
Peek 2022-01-06 22-03

"Welcome to Bytedeck" has NO prerequisites, but it still should not be startable/doable by students who are not registered in a course.

Maybe all quests that are not "Available outside course" should just redirect to the main quests page if the user is not registered in a course.

yujinyuz:
Closes #892

### DIFF
--- a/src/quest_manager/models.py
+++ b/src/quest_manager/models.py
@@ -24,7 +24,7 @@ from prerequisites.models import Prereq, IsAPrereqMixin, HasPrereqsMixin, Prereq
 
 
 class Category(IsAPrereqMixin, models.Model):
-    """ Used to group quests into 'Campaigns' 
+    """ Used to group quests into 'Campaigns'
     """
     title = models.CharField(max_length=50, unique=True)
     icon = models.ImageField(upload_to='icons/', null=True, blank=True)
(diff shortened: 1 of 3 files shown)


tylerecouture:
@yujinyuz Tests are failing. But this also needs its own test.

tylerecouture (changes requested):
Students shouldn't be able get to a page with a Start button, so this view is too far down.  I think what needs to happen is that students who are not in a course should only have access to "preview" versions of quests (which don't have any buttons), unless that quest is "available outside a course".

tylerecouture:
This might be more easily managed by adding a condition to the `QuestManager.get_available()` method.  Everything else should be taken care of.

tylerecouture:
For some reason the active course check
`def get_available_without_course(self, user):`

is seperate from `def get_available(self, user):`

So can either move the course check into `get_available()`

OR

Can add a check of `available_without_course` on the quest detail view, wherever it determines to show a preview (unavailable) or the start buttons (available)

yujinyuz:
@tylerecouture This is the current code for the quest detail view.. It doesn't use the `get_available_without_course` or `get_available` logic.. I can't seem to figure out how do the methods you mentioned fit in the quest detail view.

https://github.com/bytedeck/bytedeck/blob/ec53cbe92a40cc3c4bf5ec866f60df32c231e5e4/src/quest_manager/views.py#L487-L518

tylerecouture:
You're right, here's the line I was thinking of:

`if q.is_available(request.user)`

This brings up a bigger problem, because the `is_available()` method should use the same criteria as `get_available()` but it doesn't...argh.  Problem for another day.

For now, I think we can just add to the course logic to is_available()

So, if the user isn't registered in a course AND the self.available_outside_course is False, then return False.

There should be a method in the Profile model to check for current courses.

At the end of this, when you click on a quest in the maps, and you don't have access to that quest, it should only show you a preview with no buttons to start it.  This is what currently happens to all quests EXCEPT the first one, (because it has no prereqs, so the app thinks it should be accessible because it doesn't look at courses currently)

tylerecouture (approved):

PR #951, opened 2022-03-08
Use xp entered by students when quest when viewing submission

### LINKED ISSUE #915: Returned quest returns to default xp when Xp can be entered by students
If xp can be entered by a student, and the quest is returned, it should not revert back to the default xp value, but instead maintain the value entered by the student.

yujinyuz:
Closes #915


tylerecouture (changes requested):
Test!

PR #954, opened 2022-03-09
Fix case insensitive username

### LINKED ISSUE #858: Make usernames case insensitive
When signing up on mobile devices, the auto capitalization that phones do is annoying and can cause sign in problems if they don't remember whether they used a cap or not.  So just make sign-in case insensitive!

Will also need to be unique on case-insensitive usernames too.

yujinyuz:
Closes #858

### DIFF
--- a/src/hackerspace_online/management/commands/lowercase_usernames.py
+++ b/src/hackerspace_online/management/commands/lowercase_usernames.py
@@ -22,7 +22,7 @@ class Command(BaseCommand):
 
         # Update their usernaems to lowercase
         count = User.objects.filter(username__in=unique_usernames).update(username=Lower('username'))
-        self.stdout.write(self.style.SUCCESS(f'Updated {count} users!\n'))
+        self.stdout.write(self.style.SUCCESS(f'Updated {count} users!'))
 
         # Display users that have multiple usernames e.g. Student vs studenT vs STudenT
         # Since we have to take action for those accounts that have different cases
@@ -30,4 +30,4 @@ class Command(BaseCommand):
         if non_unique_users:
             self.stdout.write(self.style.NOTICE('List of users with multiple usernames: \n'))
             for user in non_unique_users:
-                print(user, user.id, user.username)
+                self.stdout.write(f"{user}, {user.id}, {user.username}")
--- a/src/hackerspace_online/management/commands/lowercase_usernames.py
+++ b/src/hackerspace_online/management/commands/lowercase_usernames.py
@@ -1,0 +1,33 @@
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand
+from django.db import transaction
+
+from django.db import models
+from django.db.models.functions import Lower
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+
+    help = 'Used to lowercase usernames'
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+
+        # First let's take all users that have 0 duplicates
+        users = User.objects.values('username').annotate(existing_count=models.Count(Lower('username')))
+
+        unique_usernames = users.filter(existing_count=1).values_list('username', flat=True)
+
+        # Update their usernaems to lowercase
+        count = User.objects.filter(username__in=unique_usernames).update(username=Lower('username'))
+        self.stdout.write(self.style.SUCCESS(f'Updated {count} users!\n'))
+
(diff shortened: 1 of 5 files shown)


yujinyuz:
Do I still need to add tests for this? Since it's already being handled by django-allauth

src/hackerspace_online/management/commands/lowercase_usernames.py
```diff
@@ -0,0 +1,33 @@
...
+    def handle(self, *args, **options):
+
+        # First let's take all users that have 0 duplicates
+        users = User.objects.values('username').annotate(existing_count=models.Count(Lower('username')))
+
+        unique_usernames = users.filter(existing_count=1).values_list('username', flat=True)
+
+        # Update their usernaems to lowercase
+        count = User.objects.filter(username__in=unique_usernames).update(username=Lower('username'))
+        self.stdout.write(self.style.SUCCESS(f'Updated {count} users!\n'))
```
tylerecouture, line 25: Let's stick with print() for this project, self.stdout.write adds nothing and is ugly =)

src/hackerspace_online/management/commands/lowercase_usernames.py
```diff
@@ -0,0 +1,33 @@
...
+
+from django.db import models
+from django.db.models.functions import Lower
+
+User = get_user_model()
+
+
+class Command(BaseCommand):
+
+    help = 'Used to lowercase usernames'
```
tylerecouture, line 13: Can we make this loop through all tenants?  Otherwise we have to run in manually for all tenants, there are too many tenants in production...

tylerecouture (changes requested):
A few notes:  

1. Please stick with `print()` method in this project
2. Management command to loop through all tenants
3. Simple tests: 
    1. Create a user "TestUser" with capitals and verify the username was saved as lowercase
    2. Try to sign in as the "TestUser" (including caps) and verify it works.
    3. Add help_text to the signup and login forms saying that the username is not case sensitive.

src/hackerspace_online/management/commands/lowercase_usernames.py
yujinyuz: https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/

It's being recommended from the django docs.

> Note
> 
> When you are using management commands and wish to provide console output, you should write to self.stdout and self.stderr, instead of printing to stdout and stderr directly. By using these proxies, it becomes much easier to test your custom command. Note also that you don’t need to end messages with a newline character, it will be added automatically, unless you specify the ending parameter:


Should we just use print then?

src/hackerspace_online/management/commands/lowercase_usernames.py
yujinyuz: Okay. I thought we are already using `django-tenants` and they provide a custom command

https://django-tenants.readthedocs.io/en/latest/use.html#all-tenants-command

So we could just do

```
python src/manage.py all_tenands_command lowercase_usernames
```

yujinyuz: Forgot that we still use django-tenant-schemas

tylerecouture: You are right, we ARE using django-tenants now!  I didn't know about `all_tenands_command` !!

src/hackerspace_online/management/commands/lowercase_usernames.py
tylerecouture: Nope, good catch.  Let's stick with the recommendation in the docs.  In that case, you still have one `print()` in there that should be changed to self.stdout.write! =D

tylerecouture (approved):

tylerecouture:
@yujinyuz I tried running this and it isn't handling the conflicting usernames gracefully:
```
(venv) tylere@sparagus:~/Developer/bytedeck$ ./src/manage.py tenant_command lowercase_usernames
Enter Tenant Schema ('?' to list schemas): hackerspace
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
psycopg2.errors.UniqueViolation: duplicate key value violates unique constraint "auth_user_username_key"
DETAIL:  Key (username)=(tylerecouture) already exists.
```

PR #973, opened 2022-05-25
Fixed an issue where there could be multiple primary quest maps/scape…

### LINKED ISSUE #963: Changing primary map doesn't turn off previous one
image

It's a lie though!  If you tick the box on another scape, then BOTH will be primary resulting in an error.

Dylan-RG-Patterson:
…s; Closes #963


tylerecouture (changes requested):
Looking good.  As discussed last night, please add a test (or tests) for the save() method to cover this bug.

Issue #939, opened 2022-02-17
Mark Calculations Chart not matching Mark Ranges


aluxenburg:
<img width="881" alt="Screenshot chart for mark ranges" src="https://user-images.githubusercontent.com/95307054/154538336-eddae00c-67a9-4648-a9b6-40acf02ded36.png">
I have just released mark ranges to the students.
I noticed that the chart, at the top, shows 50% for a pass, 73% for a B...
My mark ranges are different than this.  You need a 64% for a pass in our program.
Question: Is there some setting I have missed?

aluxenburg:
Any ideas regarding where we are at with this one? What does "Beta Release" in your comment mean?

tylerecouture:
I haven't had any time to do development in the past month or so, so no progress.  Beta Release is a milestone.  A while ago a roughly created 3 milestones, Beta Release, Version 2, and Version Future, just to help prioritize issues.  Our goal is to complete all the Beta Release issues before September.  My preference is to have every issue attach to one of the milestones.  Of course if we can secure funding, we can start eating into the next milestone too.  So, feel free to attach milestone to any issue!

tylerecouture:
Todo:

- [ ] Chart line name match Mark Range name field
- [ ] Chart line color matches MarkRange color (both for dark theme and for light theme!)
- [ ] Only show on chart if mark range is Active "Active"

PR #977, opened 2022-05-26
Added ability to change what grade ranges are for pass, b, and a; Clo…

9921622:
…ses #939

Grade ranges for pass, b, and a can be changed by adding them (with their respective names) in admin MarkRanges
image
image


9921622:
flake8 failed will resubmit

Issue #976, opened 2022-05-25
Rounded corners on campaign icons


tylerecouture:
In list view the campaign icons should havd rounded corners that match to every where else on the site.  Currently they do not:
https://hackerspace.localhost:8000/quests/campaigns/
![image](https://user-images.githubusercontent.com/10604391/170182111-b5b396bd-9c46-41ce-80c7-e2e377dba132.png)

PR #980, opened 2022-05-27
rounded campaign icons to match every other icon on the website; clos…

Dylan-RG-Patterson:
…es #976
image

### DIFF
--- a/src/quest_manager/templates/quest_manager/category_list.html
+++ b/src/quest_manager/templates/quest_manager/category_list.html
@@ -21,7 +21,7 @@
   {% for object in object_list %}
   <tr>
     <td>{{ object.title }}</td>
-    <td>{% if object.icon %}<img src="{{ object.icon.url }}" width="50" height="50">{% endif %}</td>
+    <td>{% if object.icon %}<img class="img-responsive panel-title-img img-rounded" src="{{ object.icon.url }}">{% endif %}</td>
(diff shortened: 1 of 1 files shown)


src/quest_manager/templates/quest_manager/category_list.html
```diff
@@ -21,7 +21,7 @@
   {% for object in object_list %}
   <tr>
     <td>{{ object.title }}</td>
-    <td>{% if object.icon %}<img src="{{ object.icon.url }}" width="50" height="50">{% endif %}</td>
+    <td>{% if object.icon %}<img class="img-responsive panel-title-img img-rounded" src="{{ object.icon.url }}" width="50" height="50">{% endif %}</td>
```
tylerecouture, line 24: What does "panel-title-img" do here?  I think that css sets the image size, but we've already hard coded the image size on this line.... hmm. 

I think we can use the panel-title-img class and remove the width and height attributes?  We should be using css anyway and not hardcoding.

tylerecouture (changes requested):
See inline comments

tylerecouture:
@Dylan-RG-Patterson there's no test, which is fine since none of the other initialization stuff is tested... but can you verify that you've manually tested it by creating a new tenant and checking these new defaults show up?  Thanks!

Dylan-RG-Patterson:
hopefully this is what was in mind.

PR #984, opened 2022-05-27
Create badge rarities upon initialization; closes #981

### LINKED ISSUE #981: New decks have no preset badge rarities
image
badges don't have rarities before they're assigned manually in a new deck.
steal the rarity system from the live hackerspace page and have them automatically generated at the conception of a new deck.
the code for this should be written in tenant/initializations.py.

Dylan-RG-Patterson:
creates 6 default rarities when a new tenant is made in tenant/initialization.py
image

### DIFF
--- a/src/badges/tests/test_models.py
+++ b/src/badges/tests/test_models.py
@@ -21,15 +21,15 @@ class BadgeRarityTestModel(TenantTestCase):
         self.assertEqual(str(self.common), self.common.name)
 
     def test_get_rarity(self):
-        self.common = baker.make(BadgeRarity, percentile=100.0)
-        self.rare = baker.make(BadgeRarity, percentile=50.0)
-        self.ultrarare = baker.make(BadgeRarity, percentile=1.0)
+        """rarity values used are chosen to avoid conflicts with defaults"""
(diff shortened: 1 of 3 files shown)


src/badges/tests/test_models.py
```diff
@@ -19,6 +19,15 @@ def setUp(self):
     def test_badge_rarity_creation(self):
         self.assertIsInstance(self.common, BadgeRarity)
         self.assertEqual(str(self.common), self.common.name)
+    
+    def test_default_badge_rarities_created(self):
```
tylerecouture, line 23: This test should be moved to the app that has the code being tested.  Can you move this to a new `test_initialization.py` in the tenant app?

tylerecouture: Also the comment here is wrong. this isn't a "migration" (that means a specific thing in databases and django).  Comment should be something like:
"""" When a new tenant is created, the initialization script should create 6 default BadgeRarity objects.""""

However, you'll need to make a new test class in the new test_initialization.py, and make sure it inherits from `TenantTestCase` (like almost all other test class in the project)

tylerecouture (changes requested):

PR #990, opened 2022-05-28
Teacher and student get different text in courses tab in profile detail; Closes #953

### LINKED ISSUE #953: Staff shouldn't be prompted to join a course
On the Profile page of Staff:
image

The "join a course" message and button should be removed.

9921622:
Teacher with no courses:
image

Teacher with courses (via /admin/)
image

Student with no courses
image

### DIFF
--- a/src/profile_manager/templates/profile_manager/profile_detail.html
+++ b/src/profile_manager/templates/profile_manager/profile_detail.html
@@ -122,10 +122,12 @@
                  </small>
                </li>
              {% endfor %}
-         {% else %}
+         {% elif not request.user.is_staff %}
            <li class="list-group-item">You have not joined a course yet for this semester.
              <a href="{% url 'courses:create' %}" class="btn btn-info" role="button">Join a Course</a>
            </li>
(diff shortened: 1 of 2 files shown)


tylerecouture (changes requested):
Great PR.  Your test has a great docstring and is awesomely commented.  This is exactly what I'd like to see.

However, seeing the result I would like two changes:

1. Even if a staff member DOES have a course, it should still say "Not applicable to staff users".
2. Please remove the edit button in these cases also

9921622:
applied the changes

staff
image
student
image

tylerecouture:
@9921622 can you confirm that the "Add" button still appears when a staff member is viewing a student's profile?  I think so, but want to make sure because it isn't in your tests.

tylerecouture (commented):
Can you confirm that the "Add" button still appears when a staff member is viewing a student's profile? I think so, but want to make sure because it isn't in your tests.

9921622:
Add button still appears when viewing student profile with staff account
image

PR #992, opened 2022-05-28
Changed "Discussion Forum" to "ByteDeck Forum" in Admin drop down; Closes #943

### LINKED ISSUE #943: Contact Change to Discuss forum may cause confusion
Students seeing "Discussion Forum" will think that it is a course-based discussion.
Can we come up with a different name for that in the dropdown? "ByteDeck Forum"?

9921622:
The first two commits should do anything since its a commit + revert

image

### DIFF
--- a/src/courses/templates/courses/chart_xp_progress.html
+++ b/src/courses/templates/courses/chart_xp_progress.html
@@ -61,7 +61,7 @@
 
   var GradeAData = {
       type: 'line',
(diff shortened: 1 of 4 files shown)


tylerecouture (changes requested):
this still has those 2 ghost commits for some reason. Good practice for you =) Can resubmit this without them?

Issue #964, opened 2022-05-20
Can delete a course that has a student registered


tylerecouture:
A course can be deleted even if it has a student currently registered in that course.  This will cause the student list to raise an error (and probably other places)
 
Need to prevent the removal of courses that have student registered in them (i.e. "CourseStudent" model object).  Perhaps add an archive option to courses so they no longer appear in the list of options for new students?

tylerecouture:
This will need some robust tests

PR #1004, opened 2022-06-03
fix: now cant delete course with student; Closes: #964

9921622:
all i did was add models.PROTECT + quick test for that


src/courses/tests/test_models.py
```diff
@@ -323,6 +323,12 @@ def test_xp_per_day_ave(self, days_so_far):
         xp_per_day = self.course_student.xp_per_day_ave()
         self.assertEqual(xp_per_day, 0)
 
+    def test_cascade_protect(self):
+        """ 
+            quick test to see if cascade
+        """ 
+        self.assertRaises(Exception, self.course.delete)
+
```
tylerecouture, line 331: This is in the wrong test class (CourseStudent instead of Course, and the docstring text appears to be cutoff.

tylerecouture (changes requested):
See my comment about test location.  Also, what's the user experience if a teacher tries to do this? Is it an error page? or what happens?  It needs to be informative to the user what's going one or why they can't delete the course.  Can you try it and see what happens?

src/courses/tests/test_models.py
9921622: Moved the test to CourseModelTest

9921622:
Added:
- text when user tries to delete course with students
- prevent any forced post request on deleteview if it has any students
- added tests on course views

src/courses/tests/test_models.py
```diff
@@ -277,6 +277,17 @@ def test_default_object_created(self):
         """ A data migration should make a default object for this model """
         self.assertTrue(Course.objects.filter(title="Default").exists())
 
+    def test_model_protection(self):
+        """ 
+            quick test to see if model deleteion is protected when trying to delete progmatically
```
tylerecouture, line 282: Two typos here (is spell check turned on in your IDE? Also, be more specific about what you are actually testing.  It's not clear to me from the docstring what you are testing specifically, which model? what's being deleted? what should happen?

src/courses/views.py
```diff
@@ -128,6 +128,24 @@ class CourseDelete(NonPublicOnlyViewMixin, DeleteView):
...
+        context = super().get_context_data(**kwargs)
+        context["course"] = self.get_object()
+
+        course_student_qs = self.get_object().coursestudent_set
+        context["population"] = course_student_qs.count()
+        context["populated"] = course_student_qs.exists()
+        return context
+
+    def post(self, request, *args, **kwargs):
+        # make sure no data can be pased in POST if populated
```
tylerecouture, line 141: typos?  "pased"

src/courses/templates/courses/course_confirm_delete.html
```diff
@@ -4,14 +4,21 @@
...
+    <p>Title: {{course.title}}</p>
+    <p>XP for 100%: {{course.xp_for_100_percent}}</p>
   </div>
   <a href="{% url 'courses:course_list' %}" role="button" class="btn btn-info">Cancel</a>
   <input type="submit" value="Delete" class="btn btn-danger" />
 </form>
+{% else %}
+  <p>Unable to delete '{{ course.title }}' as it still has students registered. Perhaps try to disable the <a href="{% url 'courses:course_update' pk=course.pk %}">active</a> field?</p>
+  <div class="well">
+    <p>Registed Student(s): {{ population }}</p>
```
tylerecouture, line 21: no need for the (s).  "Registered Students: 1" makes perfect sense

src/courses/templates/courses/course_confirm_delete.html
```diff
@@ -4,14 +4,21 @@
...
-    <p>XP for 100%: {{object.xp_for_100_percent}}</p>
+    <p>Course ID: {{course.id}}</p>
+    <p>Title: {{course.title}}</p>
+    <p>XP for 100%: {{course.xp_for_100_percent}}</p>
   </div>
   <a href="{% url 'courses:course_list' %}" role="button" class="btn btn-info">Cancel</a>
   <input type="submit" value="Delete" class="btn btn-danger" />
 </form>
+{% else %}
+  <p>Unable to delete '{{ course.title }}' as it still has students registered. Perhaps try to disable the <a href="{% url 'courses:course_update' pk=course.pk %}">active</a> field?</p>
```
tylerecouture, line 19: Change to: `Consider disabling the course by toggling the <a href="{% url 'courses:course_update' pk=course.pk %}">active</a> field instead.`

tylerecouture (changes requested):
A few minor typos, and text update, then should be good!

src/courses/tests/test_models.py
9921622: fixed

9921622: didnt have a spell check will install one

src/courses/views.py
9921622: fixed

src/courses/templates/courses/course_confirm_delete.html
9921622: tweaked

src/courses/templates/courses/course_confirm_delete.html
9921622: changed

9921622:
Added requested changes

PR #1005, opened 2022-06-04

### LINKED ISSUE #895: Add detailed view for Campaigns

Detail view itself:

Dylan-RG-Patterson:


tylerecouture (changes requested):
Looking good! Can you add some screenshots of the new pages in the PR comments?  In the future, any changes that affect the front end should include screenshots. Thanks!

tylerecouture:
I checked out your PR branch, and I don't see a link to the detailed view in the Campaigns list, which is the first item in issue #895, so I'm not even sure how to view the details of a campaign without looking in the code. ew ;)
I'm going to add a couple more items to that issues now that's I've had a chance to think about it more.

src/quest_manager/views.py
```diff
@@ -214,6 +214,18 @@ def get_context_data(self, **kwargs):
         return context
 
 
+def categorydetail(request, category_id):
+
```
tylerecouture, line 218: This should be a CBV (Class based view) and it should be with the rest of this model's views (CategoryUpdate, CategoryDelete, CategoryCreate... etc)

tylerecouture (changes requested):
Also tests!  At least basic tests, like who can access the view (see existing category/campaign test methods in test_views.py: "test_all_page_status_codes for anonymous"  "...for_students" and "...for_staff", and just add the new view in those existing tests.

PR #1009, opened 2022-06-07

### LINKED ISSUE #895: Add detailed view for Campaigns

Dylan-RG-Patterson:
Added a new detail view for campaigns that lets you look at every individual quest in them.

image

You can also view and edit a quest directly from this view.

image

### DIFF
--- a/src/quest_manager/templates/quest_manager/category_detail.html
+++ b/src/quest_manager/templates/quest_manager/category_detail.html
@@ -1,0 +1,11 @@
+{% extends "quest_manager/base.html" %}
+{% load crispy_forms_tags %}
+
+{% block heading_inner %} {{ heading }}{% endblock %}
+
+{% block content %}
+    {% with c as category %}
+        {% include "quest_manager/category_detail_content.html" %}
+    {% endwith %}
+
+{% endblock %}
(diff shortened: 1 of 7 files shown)


src/quest_manager/views.py
```diff
@@ -51,6 +51,18 @@ class CategoryList(NonPublicOnlyViewMixin, LoginRequiredMixin, ListView):
     model = Category
 
 
+@method_decorator(staff_member_required, name='dispatch')
+class CategoryDetail(NonPublicOnlyViewMixin, DetailView):
+    model = Category
+
+    def get_context_data(self, **kwargs):
+
```
tylerecouture, line 59: I don't think you need this method.  `object` is already passed to the template by the default implementation of this method, so you can use `object` directly in your template with `object.title` and `object` instead of `c` (also, don't use uninformative variable names if you can help it)

src/quest_manager/templates/quest_manager/category_detail.html
```diff
@@ -0,0 +1,11 @@
+{% extends "quest_manager/base.html" %}
+{% load crispy_forms_tags %}
+
+{% block heading_inner %} {{ heading }}{% endblock %}
+
+{% block content %}
+    {% with c as category %}
```
tylerecouture, line 7: use `with object as category` and you can remove the extra context data and method from the view

tylerecouture (changes requested):
This looks great! I requested a few tweaks in my code review

PR #1013, opened 2022-06-08

### LINKED ISSUE #895: Add detailed view for Campaigns

Dylan-RG-Patterson:
Added a detail view for campaigns, from which you can view every quest in them.

image

You can also preview and view/edit/etc. quests directly from this view.

image

### DIFF
--- a/src/quest_manager/models.py
+++ b/src/quest_manager/models.py
@@ -37,6 +37,15 @@ class Category(IsAPrereqMixin, models.Model):
     def __str__(self):
         return self.title
 
+    def get_absolute_url(self):
+        return reverse('quests:category_detail', kwargs={'pk': self.id})
+
+    def get_icon_url(self):
+        if self.icon and hasattr(self.icon, 'url'):
+            return self.icon.url
+        else:
+            return SiteConfig.get().get_default_icon_url()
+
     def xp_sum(self):
         """ Returns the total XP available from completing all visible quests in this campaign.
         Repeating quests are only counted once."""
--- a/src/quest_manager/templates/quest_manager/category_detail.html
+++ b/src/quest_manager/templates/quest_manager/category_detail.html
@@ -1,0 +1,22 @@
+{% extends "quest_manager/base.html" %}
+{% load crispy_forms_tags %}
+
+{% block heading_inner %} {{ object.title }} Campaign{% endblock %}
+
+{% block content %}
+    <div class="pull-right">
+        <a class="btn btn-warning" href="{% url 'quests:category_update' object.id %}" role="button"
+            title = "Edit this campaign" >
+        <i class="fa fa-edit"></i>
+        </a>
+        <a class="btn btn-danger" href="{% url 'quests:category_delete' object.id %}" role="button"
+            title = "Delete this campaign" >
+        <i class="fa fa-trash-o"></i>
+        </a>
+    </div>
+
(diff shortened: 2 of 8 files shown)


src/quest_manager/views.py
```diff
@@ -51,6 +51,18 @@ class CategoryList(NonPublicOnlyViewMixin, LoginRequiredMixin, ListView):
...
 
 
+@method_decorator(staff_member_required, name='dispatch')
+class CategoryDetail(NonPublicOnlyViewMixin, DetailView):
+    model = Category
+
+    def get_context_data(self, **kwargs):
+
+        kwargs['heading'] = self.object.title
+        kwargs['c'] = self.object
```
tylerecouture, line 61: Is the c variable used anywhere?  I think you are using the `object` that is passed by default to the template now, correct? So you should be able to remove this.  And you can also remove ` kwargs['heading']` and use {{object.title}} directly in the template, so you can remove the ` get_context_data` method completely, which simplifies things a bit.

src/quest_manager/models.py
```diff
@@ -37,6 +37,15 @@ class Meta:
     def __str__(self):
         return self.title
 
+    def get_absolute_url(self):
```
tylerecouture, line 40: All new methods should have tests.  These ones should be easy, you should be able to almost directly copy the same method tests for badges, or quests.... or some other model that has these exact same methods.

src/quest_manager/templates/quest_manager/category_detail.html
```diff
@@ -0,0 +1,11 @@
+{% extends "quest_manager/base.html" %}
+{% load crispy_forms_tags %}
+
+{% block heading_inner %} {{ object.title }}{% endblock %}
```
tylerecouture, line 4: Seeing the result, please add "Campaign" after the title so it's more clear what this page is:
`{{ object.title }} Campaign`

src/quest_manager/views.py
tylerecouture: Actually I don't think you are using heading anymore already, so this can just be removed.

src/quest_manager/templates/quest_manager/category_detail.html
tylerecouture: Also, add buttons to the  edit, and delete views, similar to the Quest detail view (except you'll only have two buttons)

tylerecouture (changes requested):
A few tweaks, thank you for this!

src/quest_manager/models.py
Dylan-RG-Patterson: Test added. There isn't currently a test for the get_icon_url() method on any relevant model (quests, badges) so the methodology is a little unclear to me. The test objects don't generate a custom icon upon creation, so any time the method is called in a test, it returns the default url, which can't be found using (self.client.get(self.category.get_icon_url(), follow=True).

src/quest_manager/views.py
Dylan-RG-Patterson: Removed the `get_context_data` method.

src/quest_manager/templates/quest_manager/category_detail.html
Dylan-RG-Patterson: Added the additional "Campaign" line to the campaign detail view header, after `{{ object.title }}`.

PR #1016, opened 2022-06-09
Feat: MarkRange model shows on MarkRange graph; Closes: #939

9921622:
Todo:

 - [x] Chart line name match Mark Range name field
 - [x] Chart line color matches MarkRange color (both for dark theme and for light theme!)
 - [x] Only show on chart if mark range is Active "Active"
 - [x] Ensure default MarkRange data (in a new deck) make sense.

Only thing i didnt add are tests as i still dont know how to test the views when the graph is loaded via javascript. meaning i cant use assertContains to determine if mark labels are being properly rendered

image
image
image

### DIFF
--- a/src/courses/views.py
+++ b/src/courses/views.py
@@ -18,7 +18,7 @@ from siteconfig.models import SiteConfig
 from tenant.views import NonPublicOnlyViewMixin, non_public_only_view
 
 from .forms import BlockForm, CourseStudentForm, SemesterForm
-from .models import Block, Course, CourseStudent, Rank, Semester
+from .models import Block, Course, CourseStudent, Rank, Semester, MarkRange
 
 
 # Create your views here.
@@ -44,12 +44,21 @@ def mark_calculations(request, user_id=None):
     else:
         xp_per_course = None
 
+    # only show mark ranges where student is enrolled in and is also active
+    user_courses = user.profile.current_courses().values_list('course', flat=True)
+    assigned_ranges = MarkRange.objects.filter(active=True, courses__in=user_courses)
+    all_ranges = MarkRange.objects.filter(active=True, courses=None) 
+
+    # combine assigned_ranges and all_ranges, then order by min mark
+    markranges = (assigned_ranges | all_ranges).order_by('minimum_mark')
+
     context = {
         'user': user,
         'obj': course_student,
         'courses': courses,
         'xp_per_course': xp_per_course,
         'num_courses': num_courses,
+        'markranges': markranges,
     }
     return render(request, template_name, context)
 
--- a/src/tenant/initialization.py
+++ b/src/tenant/initialization.py
@@ -7,7 +7,7 @@ among other issues
 from django.conf import settings
 from django.contrib.auth import get_user_model
 
-from courses.models import Grade, Rank, Course, Block
+from courses.models import Grade, Rank, Course, Block, MarkRange
 from quest_manager.models import Quest, Category
 from badges.models import Badge, BadgeType, BadgeRarity
 from prerequisites.models import Prereq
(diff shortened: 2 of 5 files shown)


src/courses/views.py
```diff
@@ -44,12 +44,20 @@ def mark_calculations(request, user_id=None):
     else:
         xp_per_course = None
 
+    # only show mark ranges where student is enrolled in and is also active
+    user_courses = [sc.course for sc in user.coursestudent_set.all()]
```
tylerecouture, line 48: I believe there is already a method in the Profile model to get a student's current courses.  Since that will already be fully tested, better to use that, also easier to read the code:  `user.profile.current_courses()`.

src/courses/views.py
```diff
@@ -44,12 +44,20 @@ def mark_calculations(request, user_id=None):
...
+    assigned_ranges = MarkRange.objects.filter(active=True, courses__in=user_courses)  # filter for mark ranges with courses assigned
+    all_ranges = MarkRange.objects.filter(active=True, courses=None)  # filter for mark ranges with no courses (mark range for all courses)
+
+    markranges = (assigned_ranges | all_ranges).order_by('minimum_mark')
+
```
tylerecouture, line 53: Can you add a bit more commenting to the lines above? It's not clear to me why your doing why your doing.  Your current comments are basically restating the code, which is ok, but more important is why your doing what you're doing.

` (assigned_ranges | all_ranges)` This could use a comment. What's this doing? Why?  I could guess, but not familiar with it...  or?

src/tenant/initialization.py
```diff
@@ -55,6 +56,12 @@ def create_initial_blocks(user):
     Block.objects.create(block="Default", current_teacher=user)
 
 
+def create_initial_markranges():
+    MarkRange.objects.create(name="A", minimum_mark=85.5, color_light="#FFECB3", color_dark="#FFFF00")
+    MarkRange.objects.create(name="B", minimum_mark=72.5, color_light="#BEFFFA", color_dark="#337AB7")
+    MarkRange.objects.create(name="Pass", minimum_mark=49.5, color_light="#FFB3B3", color_dark="#FF0000")
+
```
tylerecouture, line 63: Please add a simple check to the existintg initialization test that 3 mark ranges have been created in a new deck (see existing test)

tylerecouture (changes requested):
Nicely done sir!  A few tweaks requested.

tylerecouture:
@9921622 oh, it's also failing tests for some weird reason.  Seems similar to what happened last time?

src/courses/views.py
9921622: seems like current_courses() only returns StudentCourses

so its still roughly the same

user_courses = [sc.course for sc in user.profile.current_courses()]

9921622: changed

src/courses/views.py
9921622: Removed redundant comments
Added comment for (assigned_ranges | all_ranges)

src/courses/views.py
tylerecouture: Ah yes, I see what you mean.  You should use [values_list](https://docs.djangoproject.com/en/4.0/ref/models/querysets/#values-list) on the queryset (much better habit to stick with queryset operations when you can, which are way more efficient):
`user.profile.current_courses().values_list('course', flat=True)` <-- gives you a list of the provided field.  End result should be identical to your list comprehension.

src/tenant/initialization.py
9921622: done

src/courses/views.py
9921622: tweaked

9921622:
Added requested changes

Should pass markrange model test now. Since what was failing was an assert that was assuming there were no default mark ranges

tylerecouture (approved):

Issue #755, opened 2020-12-06
Handle images in notifications more gracefully


tylerecouture:
When there is a reply that is only an image or includes an image, get something like this which is kinda ugly:

![image](https://user-images.githubusercontent.com/10604391/101295843-e3d65100-37d4-11eb-9769-e57fcc9967d1.png)

Try approving a quest with only an embedded image to reproduce.

tylerecouture:
Strip HTML for notifications with 
`from django.utils.html import strip_tags`

9921622:
gonna try to fix this one

tylerecouture:
Please focus on issues I've already assigned to you first.

PR #1024, opened 2022-06-13
Fix: Rendering images in notifications; Closes: #755

9921622:
Only added a test for Notification.__str__ method 

I'm not really sure if i should add test views since I'm already testing the __str__ method in the model_tests. The notification list view just uses the notification __str__ method so adding a test in views might be redundant 

image
image

### DIFF
--- a/src/notifications/models.py
+++ b/src/notifications/models.py
@@ -6,6 +6,7 @@ from django.db.models import Q
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.html import strip_tags
+from bs4 import BeautifulSoup
 
 from tenant.utils import get_root_url
 
@@ -124,6 +125,77 @@ class Notification(models.Model):
 
     objects = NotificationManager()
 
+    def html_strip(string, char_limit=50, tag_size=1, resize_image=True, image_height=20, **kwargs) -> str:
+        """  
+            Strips all html tags except img tags and imposes a length limit. Returns the input text without html tags save for img tag
+
+            tag_size: how many letters an image should count as
+
+            resize_image: enables image resizing
+            image_height: image height after resizing in pixels
+        """
+        if not string:
+            return ""
+        if type(string) is not str:
+            string = str(string)
+
+        limit_imposed = False
+        HTML_tags = []
+        tag_index = []
+
+        # store HTML tags and index to lists
+        cache_open_index = None
+        for index in range(len(string)):
+            char = string[index]
+
+            # check for open img
+            if char == "<" and string[index:].startswith("<img"):
+                cache_open_index = index
+            
+            # check for closed if open already found 
+            elif cache_open_index is not None and char == ">":
+                # position img tag would be without html tags
+                offset = cache_open_index - len(strip_tags(string[:cache_open_index]))
+
+                start = cache_open_index
+                end = index + 1
+
+                # save the html tag and index to list
+                HTML_tags.append(string[start:end])
+                tag_index.append(cache_open_index - offset)
+
+                cache_open_index = None
+
+        # dict of html tag and its position in the string
+        html_index = {HTML_tags[i]: tag_index[i] for i in range(len(HTML_tags))}
+
+        # index count of images that will be shown
+        shown_tag_count = len([index for _, index in html_index.items() if index < char_limit])
+        # tags should count towards the character limit
+        char_limit -= shown_tag_count * tag_size
+
+        # strip all the tags and apply char limit to TEXT
+        stripped = strip_tags(string)
+        text = stripped[:char_limit]
+        limit_imposed = len(stripped) > char_limit
+        
+        # reinsert tags with new stripped text
+        for tag, index in html_index.items():
+            if index < char_limit:
+                text = text[:index] + tag + text[index:]
+
+        # resizes all images in string
+        if resize_image:
+            soup = BeautifulSoup(text, features="html.parser")
+            for img in soup.findAll('img'):
+                img['style'] = ""  # remove style as it always comes with width/height modifiers
+                img['height'] = f"{image_height}px"
+                img['width'] = "auto"
+
+            text = str(soup)
+            
+        return text + ("..." if limit_imposed else "")
+
     def __str__(self):
         try:
             target_url = self.target_object.get_absolute_url()
@@ -136,20 +208,16 @@ class Notification(models.Model):
 
         action = self.action_object
 
-        if len(str(self.action_object)) > 50:
-            action = str(self.action_object)[:50] + "..."
-        else:
-            action = str(self.action_object)
+        # uses custom strip
+        action = Notification.html_strip(action)
 
-        action = strip_tags(action)
-
-        # absolute url needed for when notfication sare sent via email
+        # absolute url needed for when notifications are sent via email
         root_url = get_root_url()
         context = {
             "sender": self.sender_object,
             "verb": self.verb,
-            "action": action,
-            "target": self.target_object,
+            "action": action,  # notif text
+            "target": self.target_object,  # basically quest name
             "verify_read": "{}{}".format(root_url, reverse('notifications:read', kwargs={"id": self.id})),
             "target_url": target_url,
         }
@@ -161,7 +229,7 @@ class Notification(models.Model):
             else:
                 url = url_common_part + " <em>%(target)s</em></a>" % context
         else:
-            url = url_common_part + "</a>"
+            url = url_common_part + "</a>"  # this is for 'teacher returned/approved ...'
         return url
 
     def mark_read(self):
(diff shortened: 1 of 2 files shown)


9921622:
the html_strip() method might need to be moved to a utility file

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,90 @@ def new_notification(sender, **kwargs):
 
         notes_unread = Notification.objects.all_unread(self.student)
         self.assertEqual(notes_unread.count(), 1)
+
+    def test_notification_str(self):
+        """ 
+            Check if notification string method leaves behind un-stripped html tags except <img>.
+        """ 
+        # CONSTANTS
```
tylerecouture, line 73: These constants should be in a `setUp(self)` method, see how other test classes use setUp.  The setup method is run before each individual test and ensures a predictable state for each test, and prevents data from changing between tests, and so that test order shouldn't matter.

tylerecouture: Hmm... I didn't realize you are running methods inside methods.  I think this might be more standard and easier to read if you just created your own test class for testing this method.  The boiler plate helper method can be it's own method, or just run it or put it in setUp.  The goal is for each specific thing you are testing to have it's own test method, as precise as possible.  I know in this project there are many waaay to big tests... those are old and bad and need to be broken up. So best practices as best we can from here on in!

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,90 @@ def new_notification(sender, **kwargs):
...
+            """ 
+                Helper for tests. Just here to prevent extra lines of boilerplate code
+            """ 
+            notification = create_notification(text=test_case)
+            strip_count = len(notification_start_text(notification.pk))
+
+            # this is just to get the text without the enclosing <a> tag
+            notification_text = str(notification)[strip_count:-len(notification_end_text)]
+
+            self.assertEqual(notification_text, expected_case)
```
tylerecouture, line 130: Put this assert statement in each individual test.  That way when a test fails, we know exactly what's causing it.

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,90 @@ def new_notification(sender, **kwargs):
...
+
+        # Check with no html (no stripped tags)
+        test_case = "TEST CASE 1 NO STRIPPED TAGS"
+        expected_case = "TEST CASE 1 NO STRIPPED TAGS"
+
+        notification_test_helper(test_case, expected_case)
+        
+        # Check with html tags (no img tag)
+        test_case = "<p>TEST CASE 2</p> WITH <h1>HTML</h1> TAGS"
+        expected_case = "TEST CASE 2 WITH HTML TAGS"
```
tylerecouture, line 140: Each of these `#Check with....` sections that end with `notification_test_helper(test_case, expected_case)` should be their own test method

tylerecouture (changes requested):
Looks awesome.  I'd like you to refactor the test structure a bit, just make it a little more standard and more helpful in case any of the asserts ever fail in the future.

src/notifications/tests/test_models.py
9921622: put it in its own test class

src/notifications/tests/test_models.py
9921622: done

src/notifications/tests/test_models.py
9921622: done

9921622:
Added the requested changes

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,120 @@ def new_notification(sender, **kwargs):
...
+    def notification_text(self, test_case) -> str:
+        """ 
+            Helper for tests. Just here to prevent extra lines of boilerplate code.
+            This just gets the comment text from notifications model and returns it
+        """ 
+        notification = self.create_notification(text=test_case)
+        strip_count = len(self.notification_start_text(notification.pk))
+
+        # this is just to get the text without the enclosing <a> tag
+        notification_text = str(notification)[strip_count:-len('"</a>')]
```
tylerecouture, line 119: Why doing this?  Also, no reason to say "just" all over the place =)

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,120 @@ def new_notification(sender, **kwargs):
...
+        notification_url = reverse('notifications:read', args=[notification_id])
+        submission_url = reverse('quest_manager:submission', args=[self.quest_submission.pk])
+        chars = f"{self.admin.username}  <a href='{site_url}{notification_url}?next={submission_url}'> <em>{self.quest.name}</em> with \""
+
+        return chars
+
+    def notification_text(self, test_case) -> str:
+        """ 
+            Helper for tests. Just here to prevent extra lines of boilerplate code.
+            This just gets the comment text from notifications model and returns it
```
tylerecouture, line 113: "just x2"  no need to says just....

tylerecouture: Also this does more that "gets the comment text" doesn't it? seems like it's manipulating the text also?  And what do you mean by "comment" text?  Sorry, these notifications are confusing.  Write a novel with examples if you have to.

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,120 @@ def new_notification(sender, **kwargs):
...
+        self.quest_submission = baker.make("quest_manager.QuestSubmission", user=self.student, quest=self.quest)
+
+        self.comment_content_type = ContentType.objects.get(app_label="comments", model="comment")
+        self.quest_submission_content_type = ContentType.objects.get(app_label="quest_manager", model="questsubmission")
+
+    #  Check if notification string method leaves behind un-stripped html tags except <img>.
+    
+    def test_notification_str__check_with_no_html(self):
+        """ 
+            Check with no html (no stripped tags)
```
tylerecouture, line 140: Check what? What are stripped tags? Maybe provide examples in the docstring.
My guess:  "Check that a notification comment(?) without htmls tags is not changed by __str__()"  Is that what you're checking here? I'm not sure because you don't actually call `str(notification)` anywhere?

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,120 @@ def new_notification(sender, **kwargs):
...
+        expected_case = "TEST CASE 1 NO STRIPPED TAGS"
+
+        self.assertEqual(
+            self.notification_text(test_case), 
+            expected_case
+        )
+        
+    def test_notification_str__check_with_html(self):
+        """ 
+            Check with html tags (no img tag)
```
tylerecouture, line 152: Same as above.  I assume you mean:

 "Test that str(notification) strips out any html tags (excluding img) included in a notification comment(?). `` tags are not tested here.

Is that correct or am I confused?

src/notifications/tests/test_models.py
```diff
@@ -63,3 +65,120 @@ def new_notification(sender, **kwargs):
...
+
+    def notification_start_text(self, notification_id) -> str:
+        """  
+            Recreates string tag for testing purposes. Equivalent to: str(notification) up until action_object
+            '{notification_start_text(id)}{action_object}{notification_end_text}' <=> str(notification)
+        """
+        site_url = 'https://tenant.test.com'
+        notification_url = reverse('notifications:read', args=[notification_id])
+        submission_url = reverse('quest_manager:submission', args=[self.quest_submission.pk])
+        chars = f"{self.admin.username}  <a href='{site_url}{notification_url}?next={submission_url}'> <em>{self.quest.name}</em> with \""
```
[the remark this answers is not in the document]
tylerecouture: Seems weird to manually regenerate the string... why not just call str(notification) since that's what you're actually testing?  I don't understand the flow of these tests.

tylerecouture (changes requested):
Sorry, I'm super confused here by your tests.  Since all your new code is an "html_strip" method, wouldn't it be way easier to just write tests for that method?

Seems like none of this confusing stuff would be required if you just tested that your new html_strip method produces the output you are expecting.

src/notifications/models.py
```diff
@@ -124,6 +125,83 @@ class Notification(models.Model):
...
+        """  
+            Strips all html tags except img tags and imposes a length limit. Returns the input text without html tags save for img tag
+
+            kwargs:
+                tag_size
+                    - char size of img tag that is default 1
+                resize_image
+                    - resizes image to a default height of 20px
+                image_height
+                    - if resize_image; changes image height to var in px
```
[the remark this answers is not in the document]
9921622: added some clarity

src/notifications/tests/test_models.py
9921622: str(notif) is called in one of the helper functions
image

9921622: clarified doc string

src/notifications/tests/test_models.py
9921622: youre correct ill replace my doc string with your explaination

9921622:
Added clarity to test doc strings
test now is focused entirely on the html strip method instead of the __str__ method

PR #1029, opened 2022-06-14
change achievements url to badges and redirect old links; closes #997

### LINKED ISSUE #997: Change url for badges to /badges/
Currently the url is /achievements/ for the badges app.

Change the base url to /badges/, and redirect /achievements/ there (to prevent old links from breaking)

Make sure to add a test for that redirection.  Might need to update some existing tests if they are looking for /achievements/ in the final url destination.

Dylan-RG-Patterson:
image
"achievements" in url path to get to badge views has been changed to "badges".
links that go to "achievements" at any point will instead redirect to the badge list view.

### DIFF
--- a/src/hackerspace_online/tests/test_views.py
+++ b/src/hackerspace_online/tests/test_views.py
@@ -83,3 +83,15 @@ class ViewsTest(ViewTestUtilsMixin, TenantTestCase):
 
     def test_password_reset_view(self):
         self.assert200('account_reset_password')
+    
+    def test_achievements_redirect_to_badges_views(self):
+        # log in a teacher
+        staff_user = User.objects.create_user(username="test_staff_user", password="password", is_staff=True)
+        self.client.force_login(staff_user)
+
+        # assert (most) relevant badge views are redirected to from old urls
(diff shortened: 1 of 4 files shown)


src/hackerspace_online/tests/test_views.py
```diff
@@ -83,3 +83,9 @@ def test_favicon_public_tenant(self, mock_connection1, mock_connection2):
 
     def test_password_reset_view(self):
         self.assert200('account_reset_password')
+    
+    def test_achievements_redirect_to_badges_view(self):
+        user = User.objects.create_user(username="test_user", password="password")
+        self.client.force_login(user)
+        response = self.client.get('/achievements/')
+        self.assertRedirects(response, reverse("badges:list"))
```
tylerecouture, line 91: Can you add another test that going to "/achievements/1" (or the detail view for some other badge that exists) still works (either works directly ior redirects to "/badges/1"?  It's not clear to me from the redirect that it will work, might only redirect to list view, which mean broken links to any badge detail view... possibly.  Test would add confidence it work.

tylerecouture (changes requested):
One more small test requested, just to make sure.

tylerecouture (approved):
Beauty

Issue #855, opened 2021-09-01
Unable to Delete or Hide Blocks


fjbagley:
I have tried removing my All Day, C and D blocks from both the main site and the admin side, but get a 500 server error in both cases. Would be good to keep those old blocks hidden so no students get confused when registering.

tylerecouture:
TODO: 

- [ ] Add an `active` field to blocks (default=True)
- [ ] If trying to delete a block that has past students in it, have a page recommend making the block inactive instead of just giving a 500 error
- [ ] Add the active field to the Block edit/create form
- [ ] Only display active=True blocks to students when they are registering for a course

PR #1030, opened 2022-06-14
Feat: Added delete/hide blocks; Closes: #855

9921622:
- [x] Add an active field to blocks (default=True)
- [x]  If trying to delete a block that has past students in it, have a page recommend making the block inactive instead of just giving a 500 error
- [x] Add the active field to the Block edit/create form __(form already has an all tag)__
- [x] Only display active=True blocks to students when they are registering for a course


src/courses/forms.py
```diff
@@ -22,6 +22,9 @@ def __init__(self, *args, **kwargs):
         if courses_qs.count() == 1:
             self.fields['course'].empty_label = None
 
+        block_qs = Block.objects.filter(active=True)
+        self.fields['block'].queryset = block_qs
+
```
tylerecouture, line 27: I think this should be moved above the code that sets a default if there is only one option.  As is, it sets the default based on all Blocks, but then you change the queryset afterwards.  So move this up, and then change Block.objects.count() to block_qs.count()

src/courses/tests/test_models.py
```diff
@@ -348,6 +348,21 @@ def test_default_object_created(self):
...
+        """ 
+        # Setup
+        student = baker.make(User)
+        block = baker.make(Block)
+        course_student = baker.make(CourseStudent, user=student, block=block)
+        self.assertTrue(CourseStudent.objects.count(), 1)
+        self.assertEqual(block, course_student.block)
+
+        # see if models.PROTECT is in place
+        self.assertRaises(Exception, block.delete)
```
tylerecouture, line 364: Be more specific in which exact exception you are testing for.  Otherwise there could be a different exception that occurs  do to some other error and your test will pass when it shouldn't.

src/courses/views.py
```diff
@@ -261,6 +261,22 @@ class BlockDelete(NonPublicOnlyViewMixin, DeleteView):
     model = Block
     success_url = reverse_lazy('courses:block_list')
 
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        course_student_qs = self.get_object().coursestudent_set
+        context["population"] = course_student_qs.count()
+        context['populated'] = course_student_qs.exists()
```
tylerecouture, line 269: The mixture of single and double quotes is triggering me.  We don't actually have a standard in this project which to use when, but side by side like this means I can't even pretend that we do, lol.

src/courses/views.py
```diff
@@ -261,6 +261,22 @@ class BlockDelete(NonPublicOnlyViewMixin, DeleteView):
...
+        return context
+
+    def post(self, request, *args, **kwargs):
+        
+        populated = self.get_object().coursestudent_set.exists()
+        if populated:
+            return self.get(request, *args, **kwargs)
+
+        return super().post(request, *args, **kwargs)
+
```
tylerecouture, line 279: It's not obvious to me what you're doing here, can you add a code comment explaining the purpose of this?

tylerecouture (changes requested):
Excellent! A few tweaks requested.  I appreciate that you tool the opportunity to change the foreign key on_delete.  Thank you!

src/courses/forms.py
9921622: done

src/courses/views.py
9921622: fixed

src/courses/views.py
9921622: done

9921622:
added requested changes

also added the ProtectedError to ```CourseModelTest.test_model_protection``` since its basically the same code as ```BlockModelTest.test_model_protection```

9921622:
Removed the check for forced post request

Also removed the test for forced post request since I cant figure out how to test for 500 errors

Issue #1032, opened 2022-06-15
Create a `deck_owner` field in the siteconfig model


tylerecouture:
Create a `deck_owner` field in the siteconfig model:
- [ ] queryset for this foreignkey field should include active staff users only. help_text should warn indicate that only the current deck owner can change this setting.
- [ ] on new decks, set this to the deck owner created during initialization
- [ ] in the site config form, only allow this field to be edited by: deck admin user, or current deck_owner.
- [ ] Will probably need a migration to handle existing production decks to populate this field.  Easiest way is to set a default on the field, and point it to the "get_deck_owner" method discussed here: https://github.com/bytedeck/bytedeck/pull/1031#pullrequestreview-1007813161

9921622:
done will mention this in #1031

Issue #637, opened 2020-09-10
New decks should create bytedeck.admin account and also user account


tylerecouture:
bytedeck.admin with specific password set in .env, and different password for the user/owner's admin account.

tylerecouture:
i.e. new decks should also create a second staff/superuser account.  Username: `owner`, password from env file.  This will be given to new deck owners who can change their name and username themselves to whatever they want.

PR #1034, opened 2022-06-16
Feat: Added deck_owner to SiteConfig; Closes: #1032

9921622:
# **This PR is dependent on #1031**

- [x] queryset for this foreignkey field should include active staff users only. help_text should warn indicate that only the current deck owner can change this setting.
- [x] on new decks, set this to the deck owner created during initialization
- [x] in the site config form, only allow this field to be edited by: deck admin user, or current deck_owner.
- [x] Will probably need a migration to handle existing production decks to populate this field. Easiest way is to set a default on the field, and point it to the "get_deck_owner" method discussed here: Feat: Added new default user; Closes: #637  #1031 (review)

I made it so the deck owner field can only be changed by the current deck owner. Since there would need to be some sort of way to tell which is the admin user


src/siteconfig/models.py
```diff
@@ -1,18 +1,42 @@
...
 
 
+def get_default_deck_owner():
+    """ 
+        This is run once during site initialization
+
+        Also need to initialize owner if none exists yet.
+        (Since initialization.py loads after everything else)
+    """ 
+    qs = User.objects.filter(username=settings.TENANT_DEFAULT_OWNER_USERNAME)
```
tylerecouture, line 24: Don't use `qs`.  give variables meaningful names. `owner_user_qs` or something like that

src/siteconfig/models.py
```diff
@@ -1,18 +1,42 @@
...
+    if qs.exists():
+        return qs.first().pk
+
+    # from testing this should only run the first 3 times?
+    # Duplicates should be prevented
+    # https://stackoverflow.com/questions/21458387/transactionmanagementerror-you-cant-execute-queries-until-the-end-of-the-atom
```
tylerecouture, line 30: Love link to references/docs/etc in the code, thank you.

src/siteconfig/models.py
```diff
@@ -1,18 +1,42 @@
...
+    with transaction.atomic():
+        owner = User.objects.create_superuser(
```
tylerecouture, line 32: This doesn't do what we want it to do for existing decks.  See my slack code.  Modify the logic here to make it work:
```def get_default_deck_owner():
    # handle existing decks by getting an existing staff user that is NOT bytedeck_admin:
    admin_username = settings.TENANT_DEFAULT_SUPERUSER_USERNAME
    non_admin_staff = User.objects.filter(is_staff=True).exclude(username=admin_username)

    if non_admin_staff.exists():  # i.e. there are more staff than just the admin user
        return non_admin_staff[0].id  # return the first one (assumes sorted by id, so should be the oldest one?)
    else:  # no other staff users available yet, so we'll have to create them
        deck_owner = User.objects.create(username=settings.TENANT_DEFAULT_DECK_OWNER_USERNAME, is_staff=True)
        return deck_owner.id
```

src/tenant/initialization.py
```diff
@@ -41,11 +41,16 @@ def create_superusers():
...
-        username=settings.TENANT_DEFAULT_OWNER_USERNAME,
-        email='owner@example.com',
-        password=settings.TENANT_DEFAULT_OWNER_PASSWORD,
-    )
+    # THIS IS HERE BECAUSE siteconfig.models.get_default_deck_owner() will run before this file (initialization.py)
+    # although siteconfig.models.get_default_deck_owner() will not run before this every time based on testing
+    # so this catches an already made owner and skips it
+    owner = User.objects.filter(username=settings.TENANT_DEFAULT_OWNER_USERNAME).first()
+    if not owner:
+        User.objects.create_superuser(
```
tylerecouture, line 49: Is there a reason we can't use get_or_create() here? https://docs.djangoproject.com/en/3.2/ref/models/querysets/#get-or-create
That's much cleaner!

tylerecouture (changes requested):
mostly minor, one major change needed to the get_default_owner method to consider existing decks.  Also needs a test for the config form that non-owner can't change the owner, and that owner can change the owner.  (I find form tests really frustrating and hard, so don't waste too much time on it if it gets too frustrating, just post a new issue that the test is still  needed)

src/siteconfig/models.py
9921622: done

src/siteconfig/models.py
9921622: done

src/tenant/initialization.py
9921622: replaced

src/siteconfig/models.py
9921622: done

9921622:
Added requested changes

src/siteconfig/models.py
```diff
@@ -6,13 +6,41 @@
...
+    """ 
+        This is run once during site initialization
+
+        Also need to initialize owner if none exists yet.
+        (Since initialization.py loads after everything else)
+    """ 
+    # return default owner if it exists. This will run after initialization.py runs
+    owner_user_qs = User.objects.filter(username=settings.TENANT_DEFAULT_OWNER_USERNAME, is_superuser=True, is_staff=True)
+    if owner_user_qs.exists():
+        return owner_user_qs.first().pk
```
tylerecouture, line 27: Lack of consistency: returning pk here but return id down below in same method.  They are the same think in this case (by default pk = id, but pk can be set to something else).  There's probably a proper one (pk or id) that foreign keys are supposed to use, can you look it up and change them to the correct one?

src/siteconfig/tests/test_models.py
```diff
@@ -103,3 +107,15 @@ def test_SiteConfig_cache_expires(self):
         cache_time_expiration = localtime() + timedelta(days=1)
         with freeze_time(cache_time_expiration, tz_offset=0):
             self.assertIsNone(cache.get(SiteConfig.cache_key()))
+
+    def test_get_deck_owner(self):
```
tylerecouture, line 111: get_deck_owner() method can return three different values from 3 different logic branches.  Make sure each of those logic branches are tested and return the expected value.  Can do all in same test if simple, or three separate tests if more complicated.

Also, by the name of this test method I'm expecting it to be tests for that method (call that method with certain parameters and ensure the expected return value) but the method isn't directly called in this test.

This is testing the SiteConfig model itself and the user_owner field.  Should be called something like `test_siteconfig_deck_owner__default_value`

src/siteconfig/tests/test_models.py
```diff
@@ -103,3 +107,15 @@ def test_SiteConfig_cache_expires(self):
...
+        """
+            Test to make sure the default deck owner is the correct one
+        """
```
tylerecouture, line 114: Be more descriptive in the docstring about what specifically is being tested.  I think should be "Test to make sure new decks have the expected default deck_owner, as set in settings.py via .env`

src/siteconfig/tests/test_views.py
```diff
@@ -74,3 +78,66 @@ def testSiteConfigUpdate_uses_newly_saved_cache_data(self):
...
+        owner_user = self.config.deck_owner
+        admin_user = User.objects.get(username=settings.TENANT_DEFAULT_ADMIN_USERNAME)
+        staff_user = baker.make(User, is_staff=True,)
+
+        URL = reverse("config:site_config_update_own")
+
+        def generate_form_data(**kwargs):
+            """ 
+                Helper func to generate valid form data using self.config values in order to use self.client.post()
+            """
```
tylerecouture, line 95: Clever!

src/tenant/initialization.py
```diff
@@ -41,12 +41,17 @@ def create_superusers():
...
+    # although siteconfig.models.get_default_deck_owner() will not run before this every time based on manual testing
+    owner, _ = User.objects.get_or_create(username=settings.TENANT_DEFAULT_OWNER_USERNAME)
+
+    # set owner vars since siteconfig.models.get_deck_owner() wont set all vars and/or has not created model yet
+    owner.email = 'owner@example.com'
+    owner.set_password(settings.TENANT_DEFAULT_OWNER_PASSWORD)
+    owner.is_superuser = True
+    owner.is_staff = True
+    owner.save()
+
```
tylerecouture, line 54: ^^^ Excellent stuff.

tylerecouture (changes requested):
A few more tweaks, mostly to improve test coverage.  Excellent work here.

src/siteconfig/models.py
9921622: seems like pk is preferable https://stackoverflow.com/a/2165880

9921622: done

src/siteconfig/tests/test_views.py
9921622: This func could definitely be genericized and I would love to work on it if its posted as an issue

src/siteconfig/tests/test_models.py
9921622: done

9921622: done

src/siteconfig/tests/test_models.py
```diff
@@ -103,3 +107,35 @@ def test_SiteConfig_cache_expires(self):
         cache_time_expiration = localtime() + timedelta(days=1)
         with freeze_time(cache_time_expiration, tz_offset=0):
             self.assertIsNone(cache.get(SiteConfig.cache_key()))
+
+    def test_get_deck_owner__correct_initialization_default_value(self):
+        """
+            Test to make sure new decks have the expected default deck_owner, as set in settings.py via .env
+        """
```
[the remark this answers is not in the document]
tylerecouture: I think I'm confused by the test name "get_deck_owner" because it doesn't match the field name "deck_owner" or the method name "get_default_deck_owner" so I don't even know what you think you are testing here.

I'd like to see the method explicitly tested.  Call the method, compare return value, one test for each of the two possible return values (or in the same method but with a docstring that makes clear what's going on"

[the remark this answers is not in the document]
tylerecouture: Leave these tests if they are passing (except change their names to match what they are testing: the deck owner field?)

And create two new test for the method itself.

tylerecouture (changes requested):
Still a bit confused by the tests, and I think missing some test coverage but a little unclear.  Thank you!

src/siteconfig/tests/test_models.py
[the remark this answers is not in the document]
9921622: changed name to test_deck_owner__correct_default_value

Also im not exactly sure what to put in the two new test since im only testing one thing and its to check if deck_owner is correctly set to owner user

9921622:
Just added some stuff to add clarity to the test_deck_owner__correct_default_value test

tylerecouture (changes requested):
Not sure why I closed this.  Just re-factor this to use your new form data generator and it's good!

9921622:
Changed the test names in test_models.py and added the new form data generator

Issue #246, opened 2019-10-21
Negative values in the Mark Distribution graph 


aidenmitchell:
![Capture](https://user-images.githubusercontent.com/30846409/67232013-4adc4f00-f3f5-11e9-8f41-32437c17aca6.PNG)

fjbagley:
This might be a good one to revisit with the new developers.

PR #1037, opened 2022-06-17
Fix: No more neg values in mark histogram; Closes: #246

9921622:
Couldn't really figure out how jchart works so i ended up adding a kind of hacky solution to prevent negative values in the mark distribution histogram.

Also i didnt really have the same issue where test_accounts are being loaded in. Its always the same column and value as the user being negative

### DIFF
--- a/src/courses/models.py
+++ b/src/courses/models.py
@@ -610,11 +610,13 @@ class MarkDistributionHistogram(Chart):
                                  )
         # course_dataset['stack'] = 'marks'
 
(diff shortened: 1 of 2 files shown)


tylerecouture (changes requested):
You can leave this one for now if it's too confusing.  I want to solve the underlying issue why negative values are being supplied to the chart, not just chop them off.  (Also, in the future please squash little fixit/typo commits, and don't leave debug print statements in your final pushed code.)

Issue #638, opened 2020-09-10
New blocks should default to specific teacher


tylerecouture:
New decks will have 2 accounts, bytedeck.admin and an owner/user admin.  New Blocks should default to the owner/user admin.

PR #1041, opened 2022-06-18
Style: Added default teacher as deck_owner for new Blocks; Closes: #638

9921622:
# This PR is dependant on #1034
Super simple issue im not even sure there needs to be a test for this since its one line


tylerecouture:
Let's re PR thins once the dependant issue is into the codebase and you can rebase onto it.

PR #1046, opened 2022-06-20
add crud views to menu items; closes #1028

### LINKED ISSUE #1028: CRUD views for Menu Items
/utilities/menuitem/

Currently the "Edit this List" item goes to django admin:
image

Need CRUD views for the front end for this model.

Dylan-RG-Patterson:
Made CRUD views for menu items and moved utilities tests into a folder for future-proofing.
image
image
image
image


src/utilities/forms.py
```diff
@@ -102,3 +104,42 @@ class Meta:
             # https://code.djangoproject.com/ticket/24453
             'sites': forms.MultipleHiddenInput(),
         }
+
+
+class MenuItemForm(forms.ModelForm):
+    url = forms.RegexField(
+        label=_("URL"),
```
tylerecouture, line 111: Why is this required?  the url field uses URLOrRelativeURLField which should already take care of this? If it doesn't then there's no point using URLOrRelativeURLField and we can just make it a standard Django URLField and remove the useless dependency.

src/utilities/forms.py
```diff
@@ -102,3 +104,42 @@ class Meta:
...
+        model = MenuItem
+        fields = '__all__'
+
+    def _trailing_slash_required(self):
+        return (
+            settings.APPEND_SLASH and
+            'django.middleware.common.CommonMiddleware' in settings.MIDDLEWARE
+        )
+
+    def clean_url(self):
```
tylerecouture, line 133: What happens without this?  Aren't these kinds of errors given by default?  Needs docstring explaining why this is needed and what it's doing

tylerecouture (changes requested):
Awesome!  A couple questions about why things are needed.

src/utilities/forms.py
Dylan-RG-Patterson: the regexfield is needed to properly validate relative URLs on the frontend, for a reason I'm currently unsure of, but the code used is sourced from django.contrib.flatpages.forms.FlatPageForm (another frontend field that accepts relative urls properly)
the decision to keep the URLOrRelativeURLField in model creation was made to maintain the same effect when creating menu items from admin.

src/utilities/forms.py
Dylan-RG-Patterson: the functions are necessary to display the correct error messages when a relative url with faulty formatting is submitted
image
tests and documentation added.

Issue #1044, opened 2022-06-20
Pages not found and PermissionDenied exception in tests


tylerecouture:
This has been going on as far back as I can check old test runs.    Note that I was not getting these missing pages or permission errors locally until I deleted and recreated my venv, so it's possible there is some problem with a version of a module we are using in our requirements.txt ?

This is the oldest run that still has data, and it's showing the same thing (Click run test suite to view it)
https://github.com/bytedeck/bytedeck/runs/6438655994?check_suite_focus=true

![image](https://user-images.githubusercontent.com/10604391/174507108-aa89df49-2945-4834-9888-9f9197003cba.png)

tylerecouture:
This might be related: https://code.djangoproject.com/ticket/10827

tylerecouture:
Also this: https://stackoverflow.com/questions/26464838/getting-model-contenttype-in-migration-django-1-7

tylerecouture:
https://stackoverflow.com/a/42791235/2700631

9921622:
# Permission denied error
issue here is that this shows up during tests:
```
Forbidden (Permission denied): <URL>
<CONSOLE ERROR HERE>
```
culprit of this error was: ``django.contrib.auth.mixins.UserPassesTestMixin.test_func`` overriden in the url's CBV.
If ``test_func`` is overriden and its output is ``False`` it will raise ``Forbidden (Permission denied)`` + ``<CONSOLE ERROR>``

Taking ``ProfileList`` as an example, the CBV assocciated with ``/quests/list/all`` overrides the ``django.contrib.auth.mixins.UserPassesTestMixin`` mixin.
ie:
```
class ProfileList(NonPublicOnlyViewMixin, UserPassesTestMixin, ListView):

   def test_func(self):
          return self.request.user.is_staff
```
In this case, if ``self.request.user.is_staff = False`` it will raise ``Forbidden (Permission denied)`` + ``<CONSOLE ERROR>``

A possible semi-solution is to add a middleware to silence to python error message (``<CONSOLE ERROR>``). Although, this doesnt remove the Forbidden (Permission denied) print.

However, this isnt really an issue since this behaviour is normal

# Not found message
issue here is that this shows up during tests:
```
Not Found: <URL>
```
culprit of this error was calling ``self.client.get`` or ``self.client.post`` on a url that does not exist
Most of these messages stem from test functions that are targeted on deleting an object then checking if it exist in the detail view.

ie:
```
@tag('do')
[...]

PR #1047, opened 2022-06-20
Feat: Added Middleware to silence PermissionDenied errors; Closes: #1044

9921622:
## Permission denied error
culprit of this error was just: ```django.contrib.auth.mixins.UserPassesTestMixin.test_func``` overriden in the url's cbv
ie: 
```
Forbidden (Permission denied): /quests/list/all
<CONSOLE ERROR HERE>
```
Then the cbv assocciated with ```/quests/list/all``` overrides the  ```django.contrib.auth.mixins.UserPassesTestMixin.test_func``` 
```
class ProfileList(NonPublicOnlyViewMixin, UserPassesTestMixin, ListView):

   def test_func(self):
          return self.request.user.is_staff
```

The fix for this was just adding a middleware to silence to error message. This doesnt remove the Forbidden (Permission denied) print though

## Not found message
Not found messages just appear when you try to get the reverse of a page that doesnt exist. That error shows up in the flatpages CRUD delete test
```
flatpage = random flatpage
self.client.post(reverse('utilities:flatpage_delete', kwargs={'pk': flatpage.pk}))

# WILL MAKE THE NOT FOUND MESSAGE APPEAR
self.assert404URL(flatpage.get_absolute_url())
```

### DIFF
--- a/src/hackerspace_online/middleware.py
+++ b/src/hackerspace_online/middleware.py
@@ -1,4 +1,6 @@
 from django.db import connections
+from django.core.exceptions import PermissionDenied
+from django.shortcuts import render
 
 
 class ForceDebugCursorMiddleware:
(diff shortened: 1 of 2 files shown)


9921622:
ya so the tests arent actually failing

tylerecouture:
@9921622 As discussed, we're gonna drop this PR and leave as is, and you're gonna close the issue with a detail comment in the issue. Thanks for investigating and figuring them out!

PR #1050, opened 2022-06-24
Feat: Added valid generic form data generation for forms and post requests: Closes: 1034

9921622:
This function should work with every form and model that baker is compatible with. Therefore, fields with validators should be set manually and manytomany fields aren't compatible.


9921622:
I think i added all the changes we discussed

Issue #1033, opened 2022-06-15
User management moved to front (out of django admin)


tylerecouture:
Need to think about the parameters of this more, but need to move user management out of django admin.  Most (all?) of this can probably be done through the profile form.

- [ ] password resetting
- [ ] set various fields on a user:
  - is_staff
  - active

PR #1079, opened 2022-07-05
FEAT: Moved user variables to profile form; Closes: #1033

9921622:
Added password+username resetting, first name, last name, acitve, staff, superuser to profile form

Also added bugfixes to generate_form_data()

### DIFF
--- a/src/hackerspace_online/tests/utils.py
+++ b/src/hackerspace_online/tests/utils.py
@@ -51,7 +51,6 @@ def generate_form_data(model=None, model_form=None, **kwargs):
     if model is None and model_form is None:
         raise ValueError('one of these arguments is required: model, model_form')
     
-    data = None
     fields = []
     exclude = []
     if model_form is not None:  # should default to model_form since it has more data specifically fields + exclude
@@ -61,13 +60,13 @@ def generate_form_data(model=None, model_form=None, **kwargs):
         exclude = model_form._meta.exclude or []
 
     data = baker.prepare(model, **kwargs,)
-    
+
     json_data = serializers.serialize('json', [data])
     json_data = json.loads(json_data)[0]["fields"]
-    json_data = {key: item or "" for key, item in json_data.items()}  # replaces None with empty string
+    json_data = {key: item if item is not None else "" for key, item in json_data.items()}  # replaces None with empty string
     
     # keep only the fields and exclude exclude
-    [json_data.pop(field_name) for field_name in fields if field_name not in fields]
+    [json_data.pop(field_name) for field_name in json_data.copy() if fields and field_name not in fields]
     [json_data.pop(field_name) for field_name in exclude]
 
     return json_data
--- a/src/profile_manager/forms.py
+++ b/src/profile_manager/forms.py
@@ -35,7 +35,7 @@ class ProfileForm(forms.ModelForm):
 
         user = self.instance.user
         user.email = self.cleaned_data['email']
-        user.save(update_fields=['email'])
+        user.save()
 
         return self.instance
 
@@ -50,13 +50,17 @@ class UserForm(forms.ModelForm):
     class Meta:
         model = get_user_model()
         fields = [
-            'username', 'first_name', 'last_name', 'is_staff', 'is_active', 'is_TA',
+            'username', 'first_name', 'last_name', 'is_TA', 'is_staff', 'is_active',
         ]
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        self.__cached_username = self.instance.username
 
-        self.fields['username'].help_text = 'Unique usernames only!'
+        self.fields['username'].help_text = (
+            'WARNING: If you change this user\'s username they will no longer be able to log in with their old username.'
+            'Make sure to inform the user of their new username!'
+        )
 
         self.fields['is_TA'].label = 'TA'
         self.fields['is_TA'].initial = self.instance.profile.is_TA
@@ -65,6 +69,9 @@ class UserForm(forms.ModelForm):
     def save(self):
         # update self.instance since ProfileForm changes User model vars
         # super().save() not saving email correctly so we only update the model fields in UserForm.fields
+        # This code only updates the UserModel fields that is in self._meta.fields
+        # required since Userform saves over user.email field which is also a field saved in ProfileForm. 
+        # This prevents email field from being overwritten in Userform
         user = self.instance
         user_fields = list(set(self._meta.fields) & set([field.name for field in user._meta.fields]))
         for name in user_fields:
--- a/src/profile_manager/forms.py
+++ b/src/profile_manager/forms.py
@@ -1,11 +1,13 @@
 from django import forms
 
+from django.contrib.auth import get_user_model
+
 from .models import Profile
 
 
 class ProfileForm(forms.ModelForm):
 
-    # this will be saved in User.email
+    # this will be saved in User model
     email = forms.EmailField(required=False)
 
     class Meta:
@@ -16,10 +18,12 @@ class ProfileForm(forms.ModelForm):
                   'visible_to_other_students', 'dark_theme', 'silent_mode', 'custom_stylesheet']
 
     def __init__(self, *args, **kwargs):
-        super(ProfileForm, self).__init__(*args, **kwargs)
+        super().__init__(*args, **kwargs)
+
         self.fields['grad_year'] = forms.ChoiceField(
             choices=Profile.get_grad_year_choices()
         )
+        
         self.fields['email'].initial = self.instance.user.email
 
     # UNIQUE if NOT NULL
@@ -27,9 +31,49 @@ class ProfileForm(forms.ModelForm):
         return self.cleaned_data['alias'] or None
 
     def save(self, *args, **kwargs):
-        super(ProfileForm, self).save(*args, **kwargs)
+        super().save(*args, **kwargs)
+
         user = self.instance.user
         user.email = self.cleaned_data['email']
-        user.save()
+        user.save(update_fields=['email'])
+
+        return self.instance
+
+
+class UserForm(forms.ModelForm):
+    """ 
+        Staff only form for profile update view
+    """ 
+
+    is_TA = forms.BooleanField(required=False)
+
+    class Meta:
+        model = get_user_model()
+        fields = [
+            'username', 'first_name', 'last_name', 'is_staff', 'is_active', 'is_TA',
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields['username'].help_text = 'Unique usernames only!'
+
+        self.fields['is_TA'].label = 'TA'
+        self.fields['is_TA'].initial = self.instance.profile.is_TA
+        self.fields['is_TA'].help_text = Profile._meta.get_field('is_TA').help_text
+
+    def save(self):
+        # update self.instance since ProfileForm changes User model vars
+        # super().save() not saving email correctly so we only update the model fields in UserForm.fields
+        user = self.instance
+        user_fields = list(set(self._meta.fields) & set([field.name for field in user._meta.fields]))
+        for name in user_fields:
+            user.__dict__[name] = self.cleaned_data[name]
+        user.save(update_fields=user_fields)
+
+        # Use updated profile instance to save profile model
+        profile = self.instance.profile
+        profile.is_TA = self.cleaned_data['is_TA']
+        profile.save(update_fields=['is_TA'])
 
         return self.instance
--- a/src/profile_manager/templates/profile_manager/form.html
+++ b/src/profile_manager/templates/profile_manager/form.html
@@ -4,8 +4,9 @@
 {% block heading_inner %}{{ heading }}{% endblock %}
 
 {% block content %}
-    <p><a href="{% url 'portfolios:edit' profile.user.pk %}" class="btn btn-info">
-        <i class="fa fa-picture-o"></i> &nbsp;Edit your Portfolio</a></p>
+    <p>
+      <a href="{% url 'portfolios:edit' profile.user.pk %}" class="btn btn-info"> <i class="fa fa-picture-o"></i> &nbsp;Edit your Portfolio</a>
+    </p>
     <div class="panel panel-info" >
         <div class="panel-heading">
             How your names will be used:
@@ -22,10 +23,21 @@
         </div>
     </div>
 
+
 <form action="" enctype="multipart/form-data" method="post">{% csrf_token %}
   <a href="{% url 'profiles:profile_detail' profile.id %}" role="button" class="btn btn-danger">Cancel</a>
   <input type="submit" value="{{ submit_btn_value }}" class="btn btn-success" /><br><br>
-  {{ form|crispy }}
+
+  {% if request.user.is_staff %}
+    <p>Want to change <a href="{% url 'profiles:change_password' profile.user.pk %}">password</a>?</p>
+  {% endif %}
+
+
+  {% for form in forms %}
+    {{ form|crispy }}
+  {% endfor %}
+
+
   <a href="{% url 'profiles:profile_detail' profile.id %}" role="button" class="btn btn-danger">Cancel</a>
   <input type="submit" value="{{ submit_btn_value }}" class="btn btn-success" />
 </form>
--- a/src/profile_manager/templates/profile_manager/form.html
+++ b/src/profile_manager/templates/profile_manager/form.html
@@ -28,12 +28,14 @@
   <a href="{% url 'profiles:profile_detail' profile.id %}" role="button" class="btn btn-danger">Cancel</a>
   <input type="submit" value="{{ submit_btn_value }}" class="btn btn-success" /><br><br>
 
-  {% if request.user.is_staff %}
-    <p>Want to change <a href="{% url 'profiles:change_password' profile.user.pk %}">password</a>?</p>
-  {% endif %}
-
-
   {% for form in forms %}
+    {% if forloop.counter == 2 %}
+      <p>
+        <h3>Authentication and Permissions</h3>
+        Only staff members may edit these fields.
+      </p>
+    {% endif %}
+
     {{ form|crispy }}
   {% endfor %}
 
--- a/src/profile_manager/templates/profile_manager/password_change_form.html
+++ b/src/profile_manager/templates/profile_manager/password_change_form.html
@@ -4,9 +4,11 @@
 {% block heading_inner %}{{ heading }}{% endblock %}
 
 {% block content %}
-<form action="" enctype="multipart/form-data" method="post">{% csrf_token %}
-  {{ form|crispy }}
-  <a href="{% url 'profiles:profile_list' %}" role="button" class="btn btn-danger">Cancel</a>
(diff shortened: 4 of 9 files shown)


src/profile_manager/urls.py
```diff
@@ -21,6 +22,9 @@
     url(r'^(?P<pk>[0-9]+)/edit/$', views.ProfileUpdate.as_view(), name='profile_update'),
     url(r'^edit/own/$', views.ProfileUpdateOwn.as_view(), name='profile_edit_own'),
 
+    path('change_username/<int:pk>/', views.UsernameUpdate.as_view(), name='change_username'),
+    path('reset_password/<int:pk>/', views.PasswordReset.as_view(), name='change_password'),
```
tylerecouture, line 26: Maybe give this a more specific name to the password change since this is for staff to reset student passwords.  name="password_change_bystaff" or something verbose and crystal clear like that.

Also, don't use underscores in urls if you can avoid it, use dashes, or even better, match the existing authentication form: [account/]password/change

So maybe:
[profile/]username/change/
[profile/]password/change/bystaff

tylerecouture (changes requested):
Awesome work, a few considerations and tweaks. Thank you!

src/profile_manager/urls.py
9921622: changed password url to ```password/change/<int:pk>/```

src/hackerspace_online/tests/utils.py
```diff
@@ -61,13 +60,13 @@ def generate_form_data(model=None, model_form=None, **kwargs):
...
+
     json_data = serializers.serialize('json', [data])
     json_data = json.loads(json_data)[0]["fields"]
-    json_data = {key: item or "" for key, item in json_data.items()}  # replaces None with empty string
+    json_data = {key: item if item is not None else "" for key, item in json_data.items()}  # replaces None with empty string
     
     # keep only the fields and exclude exclude
-    [json_data.pop(field_name) for field_name in fields if field_name not in fields]
+    [json_data.pop(field_name) for field_name in json_data.copy() if fields and field_name not in fields]
     [json_data.pop(field_name) for field_name in exclude]
```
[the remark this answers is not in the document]
9921622: i could definitely try. Are there any areas where i could start?

9921622:
added requested changes

src/profile_manager/views.py
```diff
@@ -157,6 +176,35 @@ def get_object(self):
         return self.request.user.profile
 
 
+@method_decorator(staff_member_required, name='dispatch')
+class PasswordReset(FormView):
+    form_class = SetPasswordForm
+    template_name = 'profile_manager/profile_form.html'
```
tylerecouture, line 182: Why is this using the profile form template? Seems like this should have its own template, though maybe it would be identical?  Actually is "profile_form" tempalte even used anywhere?  if it's not, rename to something more explicit and obvious, like password_change_form.html

src/profile_manager/forms.py
```diff
@@ -28,8 +34,60 @@ def clean_alias(self):
...
+    class Meta(ProfileForm.Meta):
+        fields = ProfileForm._meta.fields + [
+            'username', 'first_name', 'last_name', 'staff', 'superuser', 'active', 'is_TA',
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields['username'].initial = self.instance.user.username
+        self.fields['username'].help_text = "Unique usernames only!"
```
tylerecouture, line 63: This is what the default help text is, preferably you can just get this from the User model (User.
image

I think this is a bit too sloppy for a key field such as username.  See if you can tap into the actual field used by username and use that instead of a charfield with none of that default validation.

For example, this could break if someone tries to enter more than 150 chars, because the default CHarfield might have a different limit (the form might accept >150, but the database will reject it)

If you can just get the username's default formfield, it should all be handled:  Maybe this?
https://stackoverflow.com/questions/2584141/how-to-get-django-form-field-from-model-field

src/profile_manager/tests/test_views.py
```diff
@@ -9,6 +9,10 @@
 from hackerspace_online.tests.utils import ViewTestUtilsMixin
 from siteconfig.models import SiteConfig
 
+from profile_manager.forms import ProfileForm
+
+from hackerspace_online.tests.utils import generate_form_data
+
 
 class ProfileViewTests(ViewTestUtilsMixin, TenantTestCase):
```
tylerecouture, line 17: Please add a test that a non-staff member that accesses this view will be given a ProfileForm and another that a staff member will be given the other form.

tylerecouture (changes requested):
Looking good.  I just want to tighten up the username stuff, since that can easily break an account if we don't do it carefully!

src/profile_manager/forms.py
```diff
@@ -28,8 +34,60 @@ def clean_alias(self):
...
+        self.fields['last_name'].initial = self.instance.user.last_name
+        self.fields['staff'].initial = self.instance.user.is_staff
+        self.fields['superuser'].initial = self.instance.user.is_superuser
+        self.fields['superuser'].help_text = "Superusers can access the admin section of the deck"
+        self.fields['active'].initial = self.instance.user.is_active
+
+    def clean_username(self):
+        username = self.cleaned_data['username']
+
+        if not username:
```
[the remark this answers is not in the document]
9921622: removed

src/profile_manager/forms.py
```diff
@@ -28,8 +34,60 @@ def clean_alias(self):
...
+            raise ValidationError("Username CANNOT be empty!")
+        elif get_user_model().objects.exclude(username=self.instance.user.username).filter(username=username).exists():
```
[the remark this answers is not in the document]
9921622: since im making my own form to change the username i dont have access to the username validators that other user forms have. So its just there to check if the username is unique, and it excludes the user's name from the check to allow the username to remain the same if they dont want to change it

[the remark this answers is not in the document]
tylerecouture: So I would refactor and generate a ModelForm with model=user, and add any custom code to your ModelForm (same as already doing with the Profile's ModelForm

src/profile_manager/views.py
9921622: im using the profile_form template since it looks super generic and could work with almost any form
image
 will change name to password_change_form.html

src/profile_manager/tests/test_views.py
9921622: im not sure how to directly test which form a user is given. But i already have something like that where a user tries to force a post request with variables from the staff-only form

tylerecouture: Check this out:
https://www.obeythetestinggoat.com/book/chapter_simple_form.html#_using_the_form_in_our_views

Basically you do a GET request, then check the context data contains the expected form.  This will be more interesting now, cus you'll need one test for staf that checks that BOTH forms are present.  And another test for students.

9921622: nice! will do that

9921622:
added requested changes

src/profile_manager/templates/profile_manager/form.html
```diff
@@ -22,10 +23,21 @@
...
   <input type="submit" value="{{ submit_btn_value }}" class="btn btn-success" /><br><br>
-  {{ form|crispy }}
+
+  {% if request.user.is_staff %}
+    <p>Want to change <a href="{% url 'profiles:change_password' profile.user.pk %}">password</a>?</p>
+  {% endif %}
+
+
+  {% for form in forms %}
+    {{ form|crispy }}
```
tylerecouture, line 37: Nice.  However, I would like an H3 before the user form at the bottom. Something like this (though form1 and form2 are not what they'll be called... I don't know actually!  You can probably just use [django template filter](https://docs.djangoproject.com/en/4.0/ref/templates/builtins/#for) `forloop.counter` to see if you are in the second part of the loop to add the h3 and second form

{{ form1|crispy }}
<h3>Authentication and Permissions</h3>
<p>Only staff members may edit these fields.</p>
{{ form2|crispy}}

src/profile_manager/templates/profile_manager/password_change_form.html
```diff
@@ -0,0 +1,12 @@
+{% extends "profile_manager/base.html" %}
+{% load crispy_forms_tags %}
+
+{% block heading_inner %}{{ heading }}{% endblock %}
+
+{% block content %}
+<form action="" enctype="multipart/form-data" method="post">{% csrf_token %}
+  {{ form|crispy }}
```
tylerecouture, line 8: Can you checkout the signup and signin templates?  Having the fields span the entire width of a fullscreen browser is a bit much.  I think I put the forms in a bootstrap div or something.  Just match those two other views.

src/profile_manager/views.py
```diff
@@ -129,17 +133,20 @@ def get_context_data(self, **kwargs):
 
 class ProfileUpdate(NonPublicOnlyViewMixin, UpdateView):
     model = Profile
-    form_class = ProfileForm
+    public_form_class = ProfileForm
```
tylerecouture, line 136: call it `profile_form_class`

src/profile_manager/views.py
```diff
@@ -129,17 +133,20 @@ def get_context_data(self, **kwargs):
...
+    staff_form_class = UserForm
```
tylerecouture, line 137: `user_form_class`

src/profile_manager/views.py
```diff
@@ -149,6 +156,71 @@ def dispatch(self, *args, **kwargs):
...
+        if self.request.method == "POST":
+            kwargs.update({
+                "data": self.request.POST,
+                "files": self.request.FILES,
+            })
+
+        return kwargs
+
+    # returns instance of ProfileForm
+    def get_form_public(self):
```
tylerecouture, line 194: get_profile_form()

src/profile_manager/views.py
```diff
@@ -149,6 +156,71 @@ def dispatch(self, *args, **kwargs):
...
+        return self.public_form_class(**self.get_form_kwargs(
+            initial={'instance': self.get_object()}
+        ))
+
+    # returns instance of UserForm
+    def get_form_staff(self):
```
tylerecouture, line 200: get_user_form()

src/profile_manager/views.py
```diff
@@ -149,6 +156,71 @@ def dispatch(self, *args, **kwargs):
...
+        forms = self.get_forms()
+
+        # check if all form instances are valid else ...
+        if all([form.is_valid() for form in forms]):
+            return self.form_valid(forms)
+        return self.form_invalid(forms)
+
+    def get_context_data(self, **kwargs):
+        profile = self.get_object()
+        context = {}
```
tylerecouture, line 169: why no call to super? `context = super().get_context_data(**kwargs)` is there a reason you are resetting the context to empty = {} ?

src/profile_manager/views.py
```diff
@@ -149,6 +156,71 @@ def dispatch(self, *args, **kwargs):
...
+
+        context['heading'] = "Editing " + profile.user.get_username() + "'s Profile"
+        context['submit_btn_value'] = "Update"
+        context['profile'] = profile
+
+        return context
+
+    # handles form_kwargs for all forms  # redundant?
+    def get_form_kwargs(self, initial={}):
+        kwargs = initial
```
tylerecouture, line 182: Is there a reason you aren't calling super in this override method? `kwargs = super().get_form_kwargs()`  That initial stuff should be unnecessary.

src/profile_manager/views.py
```diff
@@ -149,6 +156,71 @@ def dispatch(self, *args, **kwargs):
...
+    # returns instance of UserForm
+    def get_form_staff(self):
+        return self.staff_form_class(**self.get_form_kwargs(
+            initial={'instance': self.get_object().user}
+        ))
+
+    # runs if all forms are valid
+    def form_valid(self, forms, *args, **kwargs):
+        # log back in since you get kicked out if you change your own username
+        if self.request.user == self.get_object().user:
```
tylerecouture, line 208: Instead of logging the user back in, can we redirect to the log in page (does it do that by default?)  Although,  I tried commenting this code out and then changing my username and it didn't kick me out anyway?

src/profile_manager/views.py
```diff
@@ -149,6 +156,71 @@ def dispatch(self, *args, **kwargs):
...
+            self.request.user.backend = 'django.contrib.auth.backends.ModelBackend'
+            login(self.request, self.request.user)
+    
+        [form.save() for form in forms]
```
tylerecouture, line 212: No need for a list and fancy comprehension here.  just keep it simple:
```
        for form in forms:
            form.save()
```

src/profile_manager/views.py
```diff
@@ -149,6 +156,71 @@ def dispatch(self, *args, **kwargs):
...
+
+        return HttpResponseRedirect(self.get_success_url())
+
+    # runs if all forms are invalid 
+    # OVERRIDDEN SO WE CAN RENDER ALL FORM VALUES WITHOUT RESETTING FORM (if form is invalid)
```
tylerecouture, line 217: Can you elaborate?  And all caps uneccessary. Maybe explain what happens if you DONT do this?  That might make it clearer why your doing this.

src/profile_manager/views.py
```diff
@@ -157,6 +229,35 @@ def get_object(self):
...
+        context['heading'] = "Changing " + profile.user.get_username() + "'s Password"
+        context['submit_btn_value'] = "Update"
+        
+        return context
+
+    def get_form(self):
+        return PasswordReset.form_class(user=self.get_instance(), **self.get_form_kwargs())
+
+    def form_valid(self, form):
+        form.save()
```
tylerecouture, line 254: Is this necessary?  Doesn't the form save already with default behavior?  Not sure why you're overriding this method

tylerecouture (changes requested):
Great work! Almost there, now just in the cleanup/pretty up phase!

src/profile_manager/forms.py
```diff
@@ -16,20 +18,62 @@ class Meta:
...
         return self.cleaned_data['alias'] or None
 
     def save(self, *args, **kwargs):
-        super(ProfileForm, self).save(*args, **kwargs)
+        super().save(*args, **kwargs)
+
         user = self.instance.user
         user.email = self.cleaned_data['email']
-        user.save()
+        user.save(update_fields=['email'])
```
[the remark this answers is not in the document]
9921622: ```update_fields```  there was left over from manual testing

removed

src/profile_manager/forms.py
```diff
@@ -16,20 +18,62 @@ class Meta:
...
+    """ 
+        Staff only form for profile update view
+    """ 
+
+    is_TA = forms.BooleanField(required=False)
+
+    class Meta:
+        model = get_user_model()
+        fields = [
+            'username', 'first_name', 'last_name', 'is_staff', 'is_active', 'is_TA',
```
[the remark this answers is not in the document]
9921622: done

src/profile_manager/forms.py
```diff
@@ -16,20 +18,62 @@ class Meta:
...
+        ]
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields['username'].help_text = 'Unique usernames only!'
```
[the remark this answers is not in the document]
9921622: changed

src/profile_manager/forms.py
```diff
@@ -16,20 +18,62 @@ class Meta:
...
+
+        self.fields['is_TA'].label = 'TA'
+        self.fields['is_TA'].initial = self.instance.profile.is_TA
+        self.fields['is_TA'].help_text = Profile._meta.get_field('is_TA').help_text
+
+    def save(self):
+        # update self.instance since ProfileForm changes User model vars
+        # super().save() not saving email correctly so we only update the model fields in UserForm.fields
```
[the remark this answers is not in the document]
9921622: added extra comment: ```# This code only updates the UserModel fields that is in self._meta.fields```

not really sure how else to explain the code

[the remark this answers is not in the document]
tylerecouture: Why is this required?  It's not normally required in a model form.  Something to do with when this form is combined with the UserForm in the same view and processed together right?  Can you explain that and the "Something"?

src/profile_manager/templates/profile_manager/form.html
```diff
@@ -22,10 +23,21 @@
...
     </div>
 
+
 <form action="" enctype="multipart/form-data" method="post">{% csrf_token %}
   <a href="{% url 'profiles:profile_detail' profile.id %}" role="button" class="btn btn-danger">Cancel</a>
   <input type="submit" value="{{ submit_btn_value }}" class="btn btn-success" /><br><br>
-  {{ form|crispy }}
+
+  {% if request.user.is_staff %}
+    <p>Want to change <a href="{% url 'profiles:change_password' profile.user.pk %}">password</a>?</p>
```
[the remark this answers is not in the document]
9921622: done

src/profile_manager/forms.py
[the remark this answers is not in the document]
9921622: ok ill add an extra comment: 
```# required since Userform saves over user.email field which is also a field saved in ProfileForm. This prevents email field from being overwritten in Userform```

src/profile_manager/templates/profile_manager/form.html
9921622: done

src/profile_manager/views.py
9921622: ```super().get_context_data(**kwargs)``` starts throwing throwing errors if i dont include a bunch of unnecessary variables. Calling super is not unnecessary anyway since all of the extra kwargs added is not used

src/profile_manager/views.py
9921622: initially overrode ```get_form_kwargs``` since i didnt know what the initial args would do to the form. Initial is necessary because ```instance``` kwarg gave the wrong instance.

Changed

src/profile_manager/views.py
9921622: changed

src/profile_manager/views.py
9921622: initially i had more complicated code when i overrode the function. now all that will happen if you remove form_invalid is that form instances will be called form instead of forms

removed

src/profile_manager/views.py
9921622: im now just calling super().get_form_kwargs() and updating the instance in both get_form functions

src/profile_manager/views.py
9921622: necessary since it doesnt save without form.save()

also all of the inherited classes dont even save the form. But there are definitely classes i missed since its hard to find all the inherited classes

also manually tested this by removing form.save(), form doesnt save

src/profile_manager/views.py
9921622: hmm before when userform was on its own page it would log me out

src/profile_manager/templates/profile_manager/password_change_form.html
9921622: resized it to col-sm-7

9921622:
added requested changes

tylerecouture (approved):
Kick ass.  That was a big one.

Issue #1059, opened 2022-06-28
deck_owner should be staff only


tylerecouture:
in initialization.py, they are created as a superuser, but in get_default_owner they are created differently.  These two should be identical, and should probably refer to the same method to ensure they are identical.

![image](https://user-images.githubusercontent.com/10604391/176095756-9bfcfc36-3f09-4bbe-97ee-fe25de00e4f1.png)

Also, these parameters can be sent as "defaults" to get_or_create() method.

Actually probably want the update_or_create() method here anyway:
https://docs.djangoproject.com/en/4.0/ref/models/querysets/#update-or-create

Ensure test to ensure deck owner is_staff=True, is_superuser=False

PR #1080, opened 2022-07-05
Fix: deck_owner no longer superuser; Closes: #1059

9921622:
cant put password in ```defaults``` so
```
owner.set_password(settings.TENANT_DEFAULT_OWNER_PASSWORD,)
owner.save()
```
is neccessary

### DIFF
--- a/src/tenant/tests/test_initialization.py
+++ b/src/tenant/tests/test_initialization.py
@@ -7,6 +7,9 @@ from courses.models import MarkRange
 from utilities.models import MenuItem
 
 
+User = get_user_model()
+
+
 class TenantInitializationTest(TenantTestCase):
 
     def test_default_badge_rarities_created(self):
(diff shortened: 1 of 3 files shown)


src/tenant/tests/test_initialization.py
```diff
@@ -42,15 +42,15 @@ def test_superusers_created(self):
         accounts = {usernames[i]: passwords[i] for i in range(len(usernames))}
 
         for username, password in accounts.items():
```
tylerecouture, line 44: I think it would be better to explicitly test each individual user that is created.  There are only two, and it makes it much clearer what is being test.  In fact, they should probably be in their own tests.  test_owener_create and test_admin_created

tylerecouture (changes requested):
Test tweak suggestion for clarity. Also failing tests.

9921622:
fixed failing test and added requested change

tylerecouture (approved):

PR #1090, opened 2022-07-16
add quest count to campaign list and detail view; closes #1086

### LINKED ISSUE #1086: Add # of quests in campaign to Campaign List view
Add a column to he table with number of quests in the campaign

image

Dylan-RG-Patterson:
campaign list view now displays number of quests.
image
campaign detail view has also been updated accordingly:
image

### DIFF
--- a/src/quest_manager/models.py
+++ b/src/quest_manager/models.py
@@ -46,11 +46,19 @@ class Category(IsAPrereqMixin, models.Model):
             return self.icon.url
         else:
             return SiteConfig.get().get_default_icon_url()
+    
+    def current_quests(self):
+        """ Returns a queryset containing every currently available quest in this campaign."""
+        return self.quest_set.all().visible().not_archived()
+
+    def quest_count(self):
+        """ Returns the total number of quests available in this campaign."""
+        return self.current_quests().count()
 
     def xp_sum(self):
         """ Returns the total XP available from completing all visible quests in this campaign.
         Repeating quests are only counted once."""
-        return self.quest_set.all().visible().not_archived().aggregate(Sum('xp'))['xp__sum']
+        return self.current_quests().aggregate(Sum('xp'))['xp__sum']
 
     def condition_met_as_prerequisite(self, user, num_required=1):
         """
--- a/src/quest_manager/templates/quest_manager/category_list.html
+++ b/src/quest_manager/templates/quest_manager/category_list.html
@@ -15,6 +15,8 @@
   <tr>
     <th>Title</th>
     <th>Icon</th>
+    <th># of Quests</th>
+    <th>XP Available</th>
     <th>Active</th>
     <th>Action</th>
   </tr>
@@ -22,6 +24,8 @@
   <tr>
     <td>{{ object.title }}</td>
     <td>{% if object.icon %}<img class="img-responsive panel-title-img img-rounded" src="{{ object.icon.url }}">{% endif %}</td>
+    <td>{{ object.quest_count }}</td>
+    <td>{{ object.xp_sum }}</td>
     <td>{{ object.active }}</td>
     <td>
       <a class="btn btn-info" href="{% url 'quests:category_detail' object.id %}" role="button" title="View Details: view the content of this campaign.">
--- a/src/quest_manager/tests/test_models.py
+++ b/src/quest_manager/tests/test_models.py
@@ -59,19 +59,45 @@ class CategoryTestModel(TenantTestCase):  # aka Campaigns
 
     def test_category_url(self):
         self.assertEqual(self.client.get(self.category.get_absolute_url(), follow=True).status_code, 200)
+    
+    def test_current_quests(self):
+        """ Test that the queryset of all quests in a campaign is returned correctly """
+
+        # assert that the current campaign has no quests
+        self.assertEqual(self.category.current_quests().count(), 0)
+
+        # create some quests as part of the test campaign, some are invalid and won't be included
+        baker.make(Quest, campaign=self.category)  # included in queryset
+        baker.make(Quest, campaign=self.category)  # included in queryset
+        baker.make(Quest, campaign=self.category, visible_to_students=False)  # NOT included in queryset because not visible
+        baker.make(Quest, campaign=self.category, archived=True)  # NOT included in queryset because archived
+        baker.make(Quest)  # NOT included in queryset because in a different campaign
+
+        # assert that the current campaign has 2 valid quests after additions
+        self.assertEqual(self.category.current_quests().count(), 2)
 
     def test_xp_sum(self):
         """ Test that the XP sum of all quests in a campaign is returned correctly """
 
         # create some quests as part of the test campaign
-        baker.make(Quest, campaign=self.category, xp=1)  # included in sum
-        baker.make(Quest, campaign=self.category, xp=2)  # included in sum
-        baker.make(Quest, xp=4)  # NOT included in sum because n a different campaign
-        baker.make(Quest, visible_to_students=False, xp=8)  # NOT included in sum becuase not visible
-        baker.make(Quest, archived=True, xp=16)  # NOT included in sum becase archived
+        baker.make(Quest, campaign=self.category, xp=1)
+        baker.make(Quest, campaign=self.category, xp=2)
 
         # check that the XP sum is correct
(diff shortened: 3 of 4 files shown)


src/quest_manager/models.py
```diff
@@ -47,6 +47,10 @@ def get_icon_url(self):
         else:
             return SiteConfig.get().get_default_icon_url()
 
+    def quest_count(self):
+        """ Returns the total number of quests available in this campaign."""
+        return self.quest_set.all().visible().not_archived().count()
```
tylerecouture, line 52: The queryset you are using is identical to the one below.  To keep these to the same and make the code more DRY, could you add a new method `current_quests()` and have it return this  queryset: `self.quest_set.all().visible().not_archived()`

Then, update both of these methods (quest_count and xp_sum) so that they use it: `self.current_quests().count()` and `self.current_quests().aggregate(Sum('xp'))['xp__sum']`

That way in the future if the way we figure out the quests we want, it will affect both equally.

Also add test for the new method.

src/quest_manager/tests/test_models.py
```diff
@@ -66,13 +66,25 @@ def test_xp_sum(self):
...
         self.assertEqual(self.category.xp_sum(), 3)
 
+    def test_quest_count(self):
+        """ Test that the number of all quests in a campaign is returned correctly """
+
+        # create some quests as a part of the test campaign
+        baker.make(Quest, campaign=self.category, xp=0)  # included in count
+        baker.make(Quest, xp=0)  # NOT included in count because in a different campaign
+        baker.make(Quest, campaign=self.category, visible_to_students=False, xp=0)  # NOT included in count because not visible
+        baker.make(Quest, campaign=self.category, archived=True, xp=0)  # NOT included in count because archived
```
tylerecouture, line 83:
```
baker.make(Quest, campaign=self.category, visible_to_students=False, xp=0)  # NOT included in count because not visible
baker.make(Quest, campaign=self.category, archived=True, xp=0)  # NOT included in count because archived
```
 These lines can be removed from this and the above test and moved into a test of the new `current_quests()` method

src/quest_manager/tests/test_models.py
```diff
@@ -66,13 +66,25 @@ def test_xp_sum(self):
...
+        baker.make(Quest, xp=4)  # NOT included in sum because in a different campaign
+        baker.make(Quest, visible_to_students=False, xp=8)  # NOT included in sum because not visible
+        baker.make(Quest, archived=True, xp=16)  # NOT included in sum because archived
 
         # check that the XP sum is correct
         self.assertEqual(self.category.xp_sum(), 3)
 
+    def test_quest_count(self):
+        """ Test that the number of all quests in a campaign is returned correctly """
+
```
tylerecouture, line 78: Can you add another assert at the top to verify the quest count is zero before you add any? (if that's the state at the start, I'm not sure)

src/quest_manager/templates/quest_manager/category_list.html
```diff
@@ -15,13 +15,15 @@
   <tr>
     <th>Title</th>
     <th>Icon</th>
+    <th># of Quests</th>
```
tylerecouture, line 18: While you're working in here.... can you add an `XP Available` column too?  Can just use the existing `xp_sum()` method

tylerecouture (changes requested):
Pretty much perfect, just taking the opportunity to improve some older code and tidy things up while you're in here =)

src/quest_manager/templates/quest_manager/category_list.html
Dylan-RG-Patterson: image
XP Available column added

src/quest_manager/tests/test_models.py
Dylan-RG-Patterson:
```suggestion
    def test_quest_count(self):
        """ Test that the number of all quests in a campaign is returned correctly """

        # assert that the current campaign has no quests
        self.assertEqual(self.category.quest_count(), 0)
```
assertion added

src/quest_manager/models.py
Dylan-RG-Patterson:
```
    def test_current_quests(self):
        """ Test that the queryset of all quests in a campaign is returned correctly """

        # assert that the current campaign has no quests
        self.assertEqual(self.category.current_quests().count(), 0)

        # create some quests as part of the test campaign, some are invalid and won't be included
        baker.make(Quest, campaign=self.category, xp=0)  # included in queryset
        baker.make(Quest, campaign=self.category, xp=0)  # included in queryset
        baker.make(Quest, visible_to_students=False, xp=8)  # NOT included in queryset because not visible
        baker.make(Quest, archived=True, xp=16)  # NOT included in queryset because archived
        baker.make(Quest, xp=4)  # NOT included in queryset because in a different campaign

        # assert that the current campaign has 2 valid quests after additions
        self.assertEqual(self.category.current_quests().count(), 2)
```
method + test added

src/quest_manager/tests/test_models.py
Dylan-RG-Patterson: other tests edited accordingly, see above conversation for newly implemented test for `current_quests()` method

src/quest_manager/models.py
```diff
@@ -46,11 +46,19 @@ def get_icon_url(self):
             return self.icon.url
         else:
             return SiteConfig.get().get_default_icon_url()
+    
+    def current_quests(self):
+        """ Returns a queryset containing every visible quest in this campaign."""
```
tylerecouture, line 51: Not true, it won't return `visible` quests if they are `archived`.  Just say currently available.

src/quest_manager/tests/test_models.py
```diff
@@ -59,20 +59,46 @@ def test_category_icon(self):
...
+        # assert that the current campaign has no quests
+        self.assertEqual(self.category.quest_count(), 0)
+
+        # create some quests as a part of the test campaign
+        baker.make(Quest, campaign=self.category, xp=0)
```
tylerecouture, line 96: Is there a reason you are setting the XP here?  I think you can just remove that field, including it is a bit confusing (it shouldn't exist in the other tests, since the DB is reset for each test, and only setUp is run.

src/quest_manager/tests/test_models.py
```diff
@@ -59,20 +59,46 @@ def test_category_icon(self):
...
         self.assertEqual(self.client.get(self.category.get_absolute_url(), follow=True).status_code, 200)
+    
+    def test_current_quests(self):
+        """ Test that the queryset of all quests in a campaign is returned correctly """
+
+        # assert that the current campaign has no quests
+        self.assertEqual(self.category.current_quests().count(), 0)
+
+        # create some quests as part of the test campaign, some are invalid and won't be included
+        baker.make(Quest, campaign=self.category, xp=0)  # included in queryset
```
tylerecouture, line 70: You can remove all these XP fields, they aren't relevant to this test anymore.

src/quest_manager/tests/test_models.py
```diff
@@ -59,20 +59,46 @@ def test_category_icon(self):
...
+        baker.make(Quest, campaign=self.category, xp=0)  # included in queryset
+        baker.make(Quest, visible_to_students=False, xp=8)  # NOT included in queryset because not visible
+        baker.make(Quest, archived=True, xp=16)  # NOT included in queryset because archived
```
tylerecouture, line 73: based on the code comment, the above two tests should be `campaign=self.category`, otherwise the archived and visible are irrelevant, because they aren't even in the campaign

tylerecouture (changes requested):
Thanks for the changes. a few TINY tweaks requested

src/quest_manager/tests/test_models.py
Dylan-RG-Patterson: unnecessary xp fields removed.

src/quest_manager/tests/test_models.py
Dylan-RG-Patterson: set campaign for the two test quests that needed it.

src/quest_manager/tests/test_models.py
Dylan-RG-Patterson: xp no longer set where unnecessary

tylerecouture (approved):

PR #1091, opened 2022-07-16
iconless campaigns display 'no icon' text instead of blank; closes #1082

### LINKED ISSUE #1082: Campaign List and Detail View no icon
If there is no icon, then put "None" instead of blank

image

Dylan-RG-Patterson:
campaigns with no icons appearing in list or detail views will now display "none" or "no icon" text respectively.
image

image

### DIFF
--- a/src/quest_manager/templates/quest_manager/category_detail_content.html
+++ b/src/quest_manager/templates/quest_manager/category_detail_content.html
@@ -1,7 +1,7 @@
 
 <div class="media">
   <div class="media-left">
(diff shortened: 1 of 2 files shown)


tylerecouture (changes requested):
Hmmm... Doesn't get_icon_url() return the default icon?  I think the problem was that the list view was using some extra uneccessary logic instead of just using get_icon_url which will always return an icon.  I didn't notice that.

So.... return the detail view how it was, which should have been fine, and just use get_icon_url method in the list view.

Dylan-RG-Patterson:
image
changed list view to now display default icon, and reverted change to detail view.

tylerecouture (approved):

Issue #1057, opened 2022-06-27
Add "Excluded dates" formset to Semester form


tylerecouture:
Currently need to use Django admin to enter excluded dates for a semester (holidays, Pro D days, etc):

![image](https://user-images.githubusercontent.com/10604391/176010312-2c0af242-3578-4a03-afea-647584242b59.png)

This uses a "model formset" to create new objects in a related model, but from the Semester form.

We need to implement this in our front end's Semester form, and remove the "via Admin" button

![image](https://user-images.githubusercontent.com/10604391/176010479-3782699f-df8d-400c-9e3a-8c411ca66cd9.png)

Read formset docs for familiarity: https://docs.djangoproject.com/en/4.0/topics/forms/formsets/
Then, specifically we need a modelformset: https://docs.djangoproject.com/en/4.0/topics/forms/modelforms/#model-formsets

Adding the buttons to include additional ExcludedDate forms is a bit tricky and I think needs to be done with Javascript, but docs are good and there should be lots of examples on the Google.

Formset should include the label and a date widget, but not include the "Date Type" field which we are deprecating.

tylerecouture:
Closed by #1095

PR #1095, opened 2022-07-19
Feat: Add Excluded Dates formset to Semester form: Closes: #1057

9921622:


tylerecouture (changes requested):
Wow, nice work!!  This is a huge one and hopefully you learned a ton figuring this out! A few things I noticed.  

1. The mobile formatting is bad:

image
So I'd suggest not putting them in seperate bootstrap columns, so they don't break.  Same with the label field, it will all fit on the smallest screen so no need to bootstrap wrapping.

2. You didn't pull before this PR?  You are a few commits behind
3. The optional `label` field on the ExcludedDate model is missing on the formset, (which you currently don't have because your code is out of date; you added that field on 11 July)

9921622:
added request changes

9921622:
added discussed changes

full screen:
image

small screen:
image

tylerecouture (changes requested):
If I enter a label without a date:

image

This needs to be caught in a test, and corrected.

9921622:
added requested changes + 

small screen now doesnt have buttons clip the scrollbar
image

tylerecouture (approved):

Issue #1075, opened 2022-07-03
add pre-commit and flake8 pre-commit hook


tylerecouture:
https://pre-commit.com/

Add to development requirements, then create a `.pre-commit-config.yaml` that runs flake 8 to the repo

https://flake8.pycqa.org/en/latest/user/using-hooks.html

use existing `.flake8` file for config options

tylerecouture:
Still needs an update somewhere.  Either fix the file or update the readme:

![image](https://user-images.githubusercontent.com/10604391/180315416-c5823005-35c9-4628-b246-fcc2521e22ed.png)

PR #1098, opened 2022-07-20
Feat: Added a pre-commit hook; Closes: #1075

9921622:


.pre-commit-config.yaml
```diff
@@ -0,0 +1,10 @@
+repos:
+    -   repo: https://github.com/PyCQA/flake8
+        rev: 4.0.1
+        hooks:
+        -   id: flake8
+            args: [
```
tylerecouture, line 6: can you use the existing .flake8 file instead of duplicating args here?  otherwise need to always remember to update in two locations.

tylerecouture (changes requested):
Missing requirement from issue: use .flake8 for config.

Also, can you update the README with a line on how to "install" the pre-commit hook? does it need to be installed somehow or registered with a new development environment or does it kick in automatically? Not sure how git would know about it unless it was registered somehow?

.pre-commit-config.yaml
tylerecouture: Readme on how to install... or auto install it:
https://pre-commit.com/#automatically-enabling-pre-commit-on-repositories

9921622:
Pretty sure it works now.

For old existing + new repositories you just need to re init git with `git init --template=.git-template`

PR #1100, opened 2022-07-21
add visible value to menu item list view; closes #1053

### LINKED ISSUE #1053: Add visible to Menu Item list view
Add column for "visible".  Don't say True or False though, find something better

Maybe use eye and eye-slash?  Or maybe a checkmark if true and x if false (using fa icons again)

Dylan-RG-Patterson:
image
added visible value as a column in the menu item list view, and created a filter to convert true or false values to corresponding fontawesome icons


tylerecouture (approved):
Nice utility tag!!

Issue #1010, opened 2022-06-07
Dropdown menu for Quests and Badges


tylerecouture:
Carve off the right side of the Quests and Badges buttons in the left menu, and place drop down menus there:
![image](https://user-images.githubusercontent.com/10604391/172280264-a9d442d2-158d-405e-a30c-bb5c46b43500.png)

Content for now:

Quests drop down menu items:
- In Progress (students)
- Completed (students)
- Drafts (contextual, if they have a drafts tab - TAs and staff)
- Campaigns (staff)
- Common quest info (staff)

Badges drop down menu items:
- Badge Types (staff)

tylerecouture:
Consider replacing current logic and styling to use bootstrap Split button dropdown:
https://getbootstrap.com/docs/3.3/components/#btn-dropdowns-split

9921622:
Need to also add link for CommonData

from #1072:
```
Maybe rename to "Common Quest Info"?

Place link to List view in drop down menu from https://github.com/bytedeck/bytedeck/issues/1010
```

PR #1103, opened 2022-07-23
Feat: CommonQuestInfo CRUD + Rename; Closes #1072

### LINKED ISSUE #1072: CRUD Views for "Common Data" and possible rename
Maybe rename to "Common Quest Info"?

Place link to List view in drop down menu from #1010

9921622:
Renamed all instances of CommonData to CommonQuestInfo  + its reference in QuestModel
Added CRUD views + tests for CommonQuestInfo 

NOTE: 
I haven't added a proper way to get into the crud views yet since it relies on #1010 which hasn't been PR'ed yet


tylerecouture (changes requested):
I'm very sorry, the Issue was ambiguous.  There is no need to change the name of the model, only the name as it appears in templates (front facing that people see).  Changing model names is stressful because it can bork the production database table for that model if not done correctly.

PR #1115, opened 2022-07-27
Feat: Added total_xp_by_tags + Tests; Closes #1102

### LINKED ISSUE #1102: tags method to calculate XP earned by tag for a user
Create a method in the tags app (in models.py?) that calculates the amount of XP earned in the current semester, by a user, with a specific tag.

Some notes on tags: implemented with [django-taggit](https://django-taggit.readthedocs.io/en/latest/index.html), so no direct model, but we can access [querysets](https://django-taggit.readthedocs.io/en/latest/api.html#filtering), and it has its own app in the project to collect views and other tag code.

The two models that implement tags are Badge and Quest.  I can't imagine any more being added in the future, so you might as well just do them individually and just add them.  However if you wanted to be fancy, you could create a method like `get_models_with_tags()` that looks through all the models in the project and returns a list of the ones that use tags (i.e. have `tags = TaggableManager()`, in fact, the django-taggit API might already have something like that.

Start by using the existing method that gets you a queryset of the quests/badges completed in the current semester, then do work on that.

Once you got that working, and fully tested we can start building out the tags feature!

9921622:


src/tags/tests/test_models.py
```diff
@@ -0,0 +1,204 @@
...
+
+        return badge, ba
+
+    # QUEST ONLY TESTS
+
+    def test_one_tag_quests_only(self):
+        """ 
+            See if correct xp is displayed using only 1 tag added to Quest objects
+        """ 
+        xp_list = [self.random.randint(0, 100) for i in range(10)]
```
tylerecouture, line 51: don't use random numbers in tests.  Tests need to be deterministic, running the identically each time, by introducing an element that can't be repeated, it reduces the value of the test.  Pick a reasonable number and use that.  I know it seems to make sense to pick a random number, to cover more possible cases, however when you run a test you are only using ONE of those possibility, but when it fails,  you don't know which one, which can make the test useless.  It would make more sense to run the test 100 times, once for each value, but then our testing suite would take forever.   So, just pick a reasonable (smallish for speed) number of items to create.

tylerecouture: I should add, I used to do this too (it SEEMS logical to do!) but have learned since then, haha.  There's lots online about this, e.g. https://softwareengineering.stackexchange.com/a/221636

src/tags/models.py
```diff
@@ -1,8 +1,29 @@
...
+    Quest = apps.get_model("quest_manager", "Quest")
+    QuestSubmission = apps.get_model("quest_manager", "QuestSubmission")
+    Badge = apps.get_model("badges", "Badge")
+    BadgeAssertion = apps.get_model("badges", "BadgeAssertion")
+
+    # get all objects with xp that is related to user
+    # convert submission/assertion qs into Quest/Badge qs
+    xpitems = [
+        Quest.objects.filter(id__in=[qs.quest.id for qs in QuestSubmission.objects.all_completed(user=user)], tags__name__in=tags).distinct(), 
+        Badge.objects.filter(id__in=[ba.badge.id for ba in BadgeAssertion.objects.all_for_user(user=user)], tags__name__in=tags).distinct(),
```
tylerecouture, line 21: Does this handle cases where a student earns multiple assertions of the same badge, or submits a repeatable quests mutliple times? Do they get all the XP for those?  I didn't see a test for that.

tylerecouture (changes requested):
Great work.  Sorry to annoy you with some required test changes =)

Great testing suite though!

src/tags/models.py
tylerecouture: I think you might be using the wrong Submission method. "Completed" for submissions means everything that shows up in the completed tab for a student. I.e. quest has been submitted by the student but might be approved by a teacher already, or might be awaiting approval). I think you want all_approved()? Please double check (reviewing from my phone :)

Sorry my method docstrings were very bad back when I wrote that stuff!

tylerecouture: Also, surprised this wasn't caught in testing.

src/tags/tests/test_models.py
9921622: im using a seed for random
image
so it should be the same set of numbers everytime
```[55, 64, 1, 34, 38, 42, 90, 24, 46, 100]``` are the first 10 set of numbers i get when i use ```self.random.randint(0, 100)```

File: src/tags/models.py (hunk 2 above)
Comment (9921622) [reply src/tags/models.py]: function now accounts for ordinal for badge assertions and quest submission + tests,
changed all_completed to all_approved()

**Comment by 9921622** [comment]
Added requested changes

File: src/tags/models.py [hunk 3]
```diff
@@ -1,8 +1,27 @@
...
+        Returns user's total xp over active_semester filtered by tags
+    """ 
+    # get models through here to prevent circular imports
+    QuestSubmission = apps.get_model("quest_manager", "QuestSubmission")
+    BadgeAssertion = apps.get_model("badges", "BadgeAssertion")
+
+    # get all objects with xp that is related to user
+    # convert submission/assertion qs into Quest/Badge qs
+    xp_list = [
+        [qs.quest.xp * qs.ordinal for qs in QuestSubmission.objects.all_approved(user=user).filter(quest__tags__name__in=tags).distinct()],
```
tylerecouture, line 18: I  don't think ordinal will work here.  For 2 reasons:
1. ordinals don't reset each semester, so for example, if a student submits only one Tinkering (10 XP) quest in a new semester, but has submitted nine last course/semester, will this count it as 100 xp? (I think semester - only xp is not considered here actually.  I can't remember if all_approved defaults to current semester only or not, same with Badge all for user).  There may already be a method on the Quest or QuestSubmission model that counts how many of that quest the student has completed (for this same problem), if you want to continue with this approach.
2. ordinals don't indicate how many quests they've completed, only how many they's started.  Say they have Tinkering (1) and Tinkering (2) and Tinkering (3), it's possible Tinkering (2) could be returned, or even dropped.  So now they've actually only got XP for Tinkering (1) and (3). As mentioned above, I think there is a method that gets you this already...I hope there is anyway =)

src/tags/models.py
```diff
@@ -1,8 +1,27 @@
 from django.db import models
 
+from django.apps import apps
 from taggit.managers import TaggableManager
 
 
+def total_xp_by_tags(user, tags: list):
```
tylerecouture, line 7: Haven't thought about it yet, but more likely this will be a queryset of tags. I think your code below will be fine, but forcing a list here might be a problem.  Maybe just indicate in the docstring the expected type for now,

src/tags/tests/test_models.py
tylerecouture: So this seems to make the use of random redundant?  All these tests will be using 55 if that's the case.  So why not just use 55 explicitly?  Also, because we run tests multithreaded/parallel, there's no guarantee even that reseeding will give the same result (at least according to the docs for Random)

I'm going to pull out the Zen of Python:
`Explicit is better than implicit`
`Simple is better than complex`

So, remove the random code, and just provide that list explicitly `xp_list = [55, 64, 1, 34, 38, 42, 90, 24, 46, 100]`

I do like the idea of randomly generating that list in the first place though, but now that we have it, lets make sure these tests are deterministic, explicit, and a little more simple (and easier to read) by removing the random code.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,247 @@
...
+        xp_iter = itertools.cycle(xp_list)
+        
+        for i in range(len(xp_list) // 2):
+            quest, qs = self.create_quest(next(xp_iter), next(ordinal_iter))
+            quest.tags.add("TAG")
+
+            badge, ba = self.create_badge(next(xp_iter), next(ordinal_iter))
+            badge.tags.add("TAG")
+
+        self.assertEqual(total_xp_by_tags(self.user, ["TAG"]), sum(xp_list[i] * ordinal_list[i] for i in range(len(ordinal_list))))
```
tylerecouture, line 247: As I mentioned above, I'm not sure whether this code is all-time XP or current-semester-only xp?  This will depend on the defaults of those methods you are using (it should say in their docstrings, or be obvious from their code).  Both are useful, however most important is the xp value earned from a tag in the current semester only.

In many model manager methods you'll see a boolean parameter `current_sem_only` or something like that, perhaps you could do the same thing here?

Regardless, we will need a test to verify that the Badge and QuestSubmission/Quest queries are doing the same thing, and which one it is: all semesters or current semester.  We definitly don't want, for example, Quests to be all-semesters and Badges to be current semester only, or vice-versa.

tylerecouture (changes requested):
This one will take a bit of back and forth probably. Thanks for your patience!

src/tags/models.py
```diff
@@ -1,8 +1,29 @@
...
+    BadgeAssertion = apps.get_model("badges", "BadgeAssertion")
+
+    # get all objects with xp that is related to user
+    # convert submission/assertion qs into Quest/Badge qs
+    xpitems = [
+        Quest.objects.filter(id__in=[qs.quest.id for qs in QuestSubmission.objects.all_completed(user=user)], tags__name__in=tags).distinct(), 
+        Badge.objects.filter(id__in=[ba.badge.id for ba in BadgeAssertion.objects.all_for_user(user=user)], tags__name__in=tags).distinct(),
+    ]
+
+    return sum(obj.xp for qs in xpitems for obj in qs)
```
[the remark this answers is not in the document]
9921622: added

src/tags/models.py
9921622: So ordinals are just cosmetic, as they are meant to tell the user how many of a quest/badge they have?

If thatst the case then fixed

src/tags/models.py
9921622: added ``expects a list of tags for tags arg (try list(tags))`` to doc string

src/tags/models.py
tylerecouture: Ya, basically cosmetic, it's to be able to differentiate the different submissions.  It's also the number of that quest they started (but not necessarily the number they've submitted or had approved, etc)

src/tags/tests/test_models.py
9921622: Fixed
Added a list of 50 ints that im calling/slicing instead of calling ``self.random.randint()``

Added an extra arg for badge filter() ``semester=SiteConfig.get().active_semester`` + added test for that

src/tags/tests/test_models.py
9921622: Fixed
Added a list of 50 ints that im calling/slicing instead of calling self.random.randint()

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+                    badge.tags.add(tag_set[1])
+
+        unpacked_tag_list = [tag_set[index] for tag_set in tag_list for index in range(len(tag_set))]
+        self.assertEqual(total_xp_by_tags(self.user, unpacked_tag_list), sum(xp_list))
+    
+    def test_one_tag_multiple_quests_badges(self):
+        """ 
+             See if correct xp is displayed using only one tag + quests and badges with multiple questsubmissions and badgeassertions
+        """ 
+        xp_list = self.xp_list[:len(self.set_list)]
```
tylerecouture, line 267: Please just put the list of values right here, explicitly. This extra complication doesn't seem necessary and just adds complication and make it harder to read the test.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+
+        for i in range(len(xp_list) // 2):
```
tylerecouture, line 269: Code comments please.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+            quest, qs = self.create_quest(next(self.xp_iter), next(self.set_iter))
+            quest.tags.add("TAG")
+
+            badge, ba = self.create_badge(next(self.xp_iter), next(self.set_iter))
+            badge.tags.add("TAG")
+
+        self.assertEqual(total_xp_by_tags(self.user, ["TAG"]), sum(xp_list[i] * self.set_list[i] for i in range(len(self.set_list))))
```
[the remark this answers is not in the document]
tylerecouture: Please apply to other tests too.

src/tags/tests/test_models.py
tylerecouture: Please apply this to other tests as needed

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+
+        self.assertEqual(total_xp_by_tags(self.user, ["TAG"]), sum(xp_list))
+
+    def test_multiple_separate_tags_quests_badges(self):
+        """ 
+            See if correct xp is displayed using only multiple tags added to Quest and Badge objects
+            only 1 unique tag per quest and badge
+        """ 
+        tag_list = [f"tag-{i}" for i in range(5 * 2)]  # 10
+        xp_list = self.xp_list[:5 * len(tag_list)]  # 50
```
tylerecouture, line 226: Remove all this and explicitly put the lists

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+            badge.tags.add("TAG")
+
+        self.assertEqual(total_xp_by_tags(self.user, ["TAG"]), sum(xp_list))
+
+    def test_multiple_separate_tags_quests_badges(self):
+        """ 
+            See if correct xp is displayed using only multiple tags added to Quest and Badge objects
+            only 1 unique tag per quest and badge
+        """ 
+        tag_list = [f"tag-{i}" for i in range(5 * 2)]  # 10
```
tylerecouture, line 225: Why 10? Why 5x2? Tests are lacking in code comments throughout.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+            quest.tags.add("TAG")
+
+            badge, _ = self.create_badge(next(self.xp_iter))
+            badge.tags.add("TAG")
+
+        self.assertEqual(total_xp_by_tags(self.user, ["TAG"]), sum(xp_list))
+
+    def test_multiple_separate_tags_quests_badges(self):
+        """ 
+            See if correct xp is displayed using only multiple tags added to Quest and Badge objects
```
tylerecouture, line 222: "displayed" do you mean returned (by the function)?

tylerecouture, line 222: Not sure what "only multiple tags added to...." means here.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+        self.xp_list = [  # 50
+            55, 84, 74, 85, 61, 61, 22, 39, 12, 46, 
+            36, 22, 93, 16, 2, 36, 28, 95, 10, 79, 
+            80, 91, 16, 74, 94, 50, 28, 21, 64, 62, 
+            54, 44, 100, 79, 8, 39, 92, 97, 16, 14, 
+            75, 62, 37, 33, 75, 19, 13, 7, 71, 63,
+        ]
+        self.xp_iter = itertools.cycle(self.xp_list)
+
+    def create_quest(self, xp, qs_quant=1):
```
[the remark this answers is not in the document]
tylerecouture: Use: ‘num_submissions‘

sub or submission is the standard shortened form for a QuestSubmission object(s) in this project.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+        self.assertEqual(total_xp_by_tags(self.user, unpacked_tag_list), sum(xp_list))
+
+    def test_one_tag_multiple_badges_only(self):
+        """ 
+            See if correct xp is displayed using only one tag + badges with multiple badgeassertions
+        """
+        xp_list = self.xp_list[:len(self.set_list)]  # 10
+
+        for i in range(len(xp_list)):
+            badge, ba = self.create_badge(xp_list[i], self.set_list[i])
```
tylerecouture, line 184: What is set_list? I can't follow this code

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+        unpacked_tag_list = [tag_set[index] for tag_set in tag_list for index in range(len(tag_set))]
+        self.assertEqual(total_xp_by_tags(self.user, unpacked_tag_list), sum(xp_list))
+
+    def test_one_tag_multiple_badges_only(self):
+        """ 
+            See if correct xp is displayed using only one tag + badges with multiple badgeassertions
+        """
+        xp_list = self.xp_list[:len(self.set_list)]  # 10
+
+        for i in range(len(xp_list)):
```
tylerecouture, line 183: What's the purpose of this loop? Code comments needed everywhere

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+
+    # QUEST + BADGE TESTS
+
+    def test_one_tag_quests_badges(self):
+        """ 
+            See if correct xp is displayed using only 1 tag added to Quest and Badge objects
+        """ 
+        xp_list = self.xp_list[:20]  # 20
+        
+        for i in range(len(xp_list) // 2):
```
tylerecouture, line 211: Is there a reason not to just use: 

for xp in xp_list:

Way easy to follow, less code, can get rid of the confusing iter stuff?

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+
+    def setUp(self):
+        self.user = baker.make(User)
+
+        # pre determined amount of qs or ba linked to quest/badges
+        self.set_list = [1, 1, 2, 2, 3, 3, 4, 4, 5, 5]  # 10
+        self.set_iter = itertools.cycle(self.set_list)
+
+        # pre determined values for xp
+        self.xp_list = [  # 50
```
tylerecouture, line 29: Please just put the values needed explicitly in each test. I don't think you need more than 5 values for any given test, to avoid slowing down our already slow test suite.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+class Tag_total_xp_by_tags_Tests(TenantTestCase):
+    """ 
+        Specialized TestClass for testing total_xp_by_tags function
+    """ 
+
+    def setUp(self):
+        self.user = baker.make(User)
+
+        # pre determined amount of qs or ba linked to quest/badges
+        self.set_list = [1, 1, 2, 2, 3, 3, 4, 4, 5, 5]  # 10
```
tylerecouture, line 25: What is this

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+        self.set_iter = itertools.cycle(self.set_list)
```
tylerecouture, line 26: What is this for?

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+
+class Tag_total_xp_by_tags_Tests(TenantTestCase):
+    """ 
+        Specialized TestClass for testing total_xp_by_tags function
+    """ 
+
+    def setUp(self):
+        self.user = baker.make(User)
+
+        # pre determined amount of qs or ba linked to quest/badges
```
tylerecouture, line 24: Qs or ba? Don't use shortened forms unless they are commonly used through the project. Qs = queryset.

Full variable names please.

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+
+        # pre determined values for xp
+        self.xp_list = [  # 50
+            55, 84, 74, 85, 61, 61, 22, 39, 12, 46, 
+            36, 22, 93, 16, 2, 36, 28, 95, 10, 79, 
+            80, 91, 16, 74, 94, 50, 28, 21, 64, 62, 
+            54, 44, 100, 79, 8, 39, 92, 97, 16, 14, 
+            75, 62, 37, 33, 75, 19, 13, 7, 71, 63,
+        ]
+        self.xp_iter = itertools.cycle(self.xp_list)
```
tylerecouture, line 36: No idea what this is doing

src/tags/tests/test_models.py
tylerecouture, line 38: Name is inaccurate?  `create_quest_and_submissions` seems more accurate

src/tags/models.py
```diff
@@ -1,7 +1,38 @@
...
+    """ 
+    # get models through here to prevent circular imports
+    QuestSubmission = apps.get_model("quest_manager", "QuestSubmission")
+    BadgeAssertion = apps.get_model("badges", "BadgeAssertion")
+
+    # get all objects with xp that is related to user and tag
+    xp_list = [
+        # QuestSubmission xp sum  
+        QuestSubmission.objects.all_approved(user, active_semester_only=True).filter(
+            is_completed=True,
```
tylerecouture, line 24: I feel like this is a huge improvement from your last version! You also got it into a single operation. Nicely done! Your Django queryset Kung Fu is improving :)

I think `is_completed` filter is uneccessary, since that's already taken care of in `all_approved()`. Can you confirm?

src/tags/models.py
```diff
@@ -1,7 +1,38 @@
...
+            quest__tags__name__in=tags
+        ).exclude(do_not_grant_xp=True).distinct().aggregate(Sum("quest__xp"))['quest__xp__sum'],
+        
+        # BadgeAssertion xp sum 
+        BadgeAssertion.objects.all_for_user(user).filter(
+            semester=SiteConfig.get().active_semester, 
+            badge__tags__name__in=tags
+        ).exclude(do_not_grant_xp=True).distinct().aggregate(Sum("badge__xp"))['badge__xp__sum'],
+    ]
+    return sum(xp for xp in xp_list if xp is not None)
```
tylerecouture, line 34: With your new version, would the sum ever be None? If not, I think it would be simpler to get rid of the list and do

`return quest_xp + badge_xp`

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+
+        return badge, ba
+
+    # QUEST ONLY TESTS
+
+    def test_one_tag_quests_only(self):
+        """ 
+            See if correct xp is displayed using only 1 tag added to Quest objects
+        """ 
+        xp_list = self.xp_list[:10]  # 10
```
tylerecouture, line 65: Be explicit

`x_list =[1, 2, 3, 4, 5]` or whatever values.

src/tags/models.py
```diff
@@ -1,7 +1,38 @@
...
+from siteconfig.models import SiteConfig
+
+
+def total_xp_by_tags(user, tags: list):
+    """ 
+        Returns user's total xp over active_semester filtered by tags
+
+        expects a list of tags for tags arg (try list(tags))
+    """ 
+    # get models through here to prevent circular imports
```
tylerecouture, line 16: Nicely done. Code comment appreciated

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+                    badge.tags.add(tag_el)
+
+        self.assertEqual(total_xp_by_tags(self.user, tag_list), sum(xp_list))
+
+    def test_multiple_crossing_tags_quest_badges(self):
+        """ 
+            See if correct xp is displayed using only multiple tags added to Quest and Badge objects
+            each Quest and Badge object should have multiple tags
+        """ 
+        tag_list = [(f"tag0-{i}", f"tag1-{i}") for i in range(5 * 2)]
```
tylerecouture, line 245: Do this in a python prompt, or print the output, then paste the literal values here explicitly.

Also, maybe call it `tag_tuples` which is more accurate (plural variable names usually indicate lists, so you want the variable to tell you what it's a list of)

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,276 @@
...
+        xp_list = self.xp_list[:5 * len(tag_list)]  # 50
+
+        for tag_set in tag_list:
```
tylerecouture, line 248: What does this mean? Do you mean a set as in a python/mathematical set? Or do you mean it in the normal "grouping" sense? Also, not sure what this loop is doing and why without a comment. From your list it appears these are tuples not sets?

tylerecouture (changes requested):
Great improvement on the method! Thank you.

However, I am all over the place with my review of your tests. Sorry! It's hard to do from a phone, haha.

Please read through all the review comments before starting work on them, because they are scattered, but I think most comments apply to all tests.

Because this is a brand new feature with no live testing, these tests need to be perfect, readable, as simple as possible, and thoroughly documented/commented.

src/tags/models.py
```diff
@@ -1,7 +1,38 @@
...
 from taggit.managers import TaggableManager
 
+from siteconfig.models import SiteConfig
+
+
+def total_xp_by_tags(user, tags: list):
+    """ 
+        Returns user's total xp over active_semester filtered by tags
+
+        expects a list of tags for tags arg (try list(tags))
```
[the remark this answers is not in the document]
9921622: done

src/tags/tests/test_models.py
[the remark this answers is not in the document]
9921622: fixed to ``quest_submission_quantity``

src/tags/models.py
9921622: sum can be none since ``aggregate()`` will always return a dictionary of keys of whatever arg you put in.
if there are no badges or quests linked to the user, aggregate will return a dictionary like: ``{ 'badge__xp__sum': None }``

9921622: i could write 
``
return( quest_xp or 0) + (badge_xp or 0)
``
but it looks ugly IMO

src/tags/models.py
9921622: nope ``is_completed`` is necessary, ``all_approved()`` doesnt seem to filter for completed quests only approved ones.

Although, there were a bunch of commented out lines that did filter for it

src/tags/tests/test_models.py
9921622: mixed the 2 up, fixed

src/tags/tests/test_models.py
9921622: done

src/tags/tests/test_models.py
9921622: fixed
now using ``range(0, len(xp_list), 2)``

src/tags/tests/test_models.py
9921622: renamed to `` quantity_list``

src/tags/tests/test_models.py
9921622: 10 since 5 per quest and 5 per badge.

src/tags/tests/test_models.py
9921622: changed displayed to returned

src/tags/tests/test_models.py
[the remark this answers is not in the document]
9921622: turned the complicated sum functions into ``expected_xp = sum(...code..)``

src/tags/tests/test_models.py
9921622: done

src/tags/tests/test_models.py
9921622: done

src/tags/tests/test_models.py
9921622: renamed to ``See if correct xp is returned using two tags per Quest and Badge object``

9921622:
Added requested changes

src/tags/models.py
tylerecouture: Can you paste this in as a code comment?

```
sum can be none since aggregate() will always return a dictionary of keys of whatever arg you put in.
if there are no badges or quests linked to the user, aggregate will return a dictionary like: { 'badge__xp__sum': None }
```

tylerecouture (approved):
I left some review comments, action if you want, or not, but I'm going to approve.

Note we may need to revisit if we discover a case where the method doesn't work as intended, but the test suite you have now covers all the cases I can think of preemptively!

src/tags/models.py
9921622: done

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,277 @@
...
+    def setUp(self):
+        self.user = baker.make(User)
+
+    def create_quest_and_submissions(self, xp, quest_submission_quantity=1):
+        """ 
+            Creates and returns quest with linked quest submission objects
+            Args
+                xp - amount of xp quest obj will have
+                quest_submission_quantity - how many submissions are crated
+        """
```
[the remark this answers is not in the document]
9921622: fixed

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,277 @@
...
+
+    def test_multiple_separate_tags_quests_only(self):
+        """ 
+            See if correct xp is returned using only multiple tags added to Quest objects
+            only 1 unique tag per quest
+        """ 
+        xp_list = [55, 84, 74, 85, 61]
+        tag_list = ['tag-0', 'tag-1', 'tag-2', 'tag-3', 'tag-4']
+
+        for i in range(len(xp_list)):
```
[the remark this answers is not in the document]
9921622: replaced

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,277 @@
...
+            semester=SiteConfig().get().active_semester,
+
+            _quantity=quest_submission_quantity,
+        )
+
+        return quest, quest_submissions
+
+    def create_badge_and_assertions(self, xp, badge_assertion_quantity=1):
+        """ 
+            Creates and returns badge with linked badge assertion objects
```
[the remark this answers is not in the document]
9921622: added this to first line  ``..that are also linked to self.user``

src/tags/tests/test_models.py
```diff
@@ -0,0 +1,277 @@
...
+            quest, _ = self.create_quest_and_submissions(xp_list[i], quantity_list[i])
+            quest.tags.add("TAG")
+
+        calculated_xp = total_xp_by_tags(self.user, ["TAG"])
+        expected_xp = sum(xp_list[i] * quantity_list[i] for i in range(len(xp_list)))
+        self.assertEqual(calculated_xp, expected_xp)
+
+    def test_one_tag_quests_only__different_semesters(self):
+        """ 
+            See if correct xp is returned using only one tag + quests with multiple questsubmissions in different semesters
```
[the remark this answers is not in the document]
9921622: active semester quest is now ``100 xp``

while the others are ``399 xp``

src/tags/tests/test_models.py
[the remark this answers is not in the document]
9921622: added comments to all forloops

not really sure why there is a need to comment on these loops

for example
image
looking at the code it has ``self.create_quest_and_submissions()`` and ``quest.tags.add()``
with that in mind, couldnt the reader just infer that It creates 5 quests and gives each quest a single, unique tag?

PR #1122, opened 2022-07-30
Tweak: PR #1115 code cleanup

9921622:
code cleanup from pr #1115

### DIFF
--- a/src/tags/models.py
+++ b/src/tags/models.py
@@ -8,27 +8,33 @@ from siteconfig.models import SiteConfig
 
 
 def total_xp_by_tags(user, tags):
-    """ 
-        Returns user's total xp over active_semester filtered by tags
-        Args:
-            tags - a queryset or list of Tag objects
-    """ 
+    """
+    Returns user's total xp over active_semester filtered by tags
+
+    Args:
+        tags (list[str], Queryset<Tag>): a queryset or list of Tag objects
+
+    Returns:
+        int : total xp of user related to tag
+    """
     # get models through here to prevent circular imports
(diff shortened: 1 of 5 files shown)


src/tags/tests/test_models.py
```diff
@@ -20,12 +20,17 @@ def setUp(self):
...
 
     def create_quest_and_submissions(self, xp, quest_submission_quantity=1):
-        """ 
-            Creates and returns quest with linked quest submission objects
-            Args
-                xp - amount of xp quest obj will have
-                quest_submission_quantity - how many submissions are crated
-        """ 
+        """
+        Creates and returns quest with linked quest submission objects
```
tylerecouture, line 24: ...for self.user

src/tags/models.py
```diff
@@ -8,16 +8,22 @@
...
-    """ 
-        Returns user's total xp over active_semester filtered by tags
-        Args:
-            tags - a queryset or list of Tag objects
-    """ 
+    """
+    Returns user's total xp over active_semester filtered by tags
+
+    Args:
+        tags (list[str], Queryset<Tag>): a queryset or list of Tag objects
```
tylerecouture, line 15: Is the ability to accept a Queryset actually tested?  if not, can you add one?

tylerecouture (changes requested):
Can you add a test to ensure that your method will work with a queryset?  If it doesn't, you can remove the info about querysets in your docs and add it as a new issue (we can worry about it later, not important for our current deadline)

src/tags/models.py
9921622: added test for qs and list

tylerecouture (approved):
Nice

PR #1123, opened 2022-08-03
Feat: Tags CRUD View + Tests; Closes #1120

### LINKED ISSUE #1120: Tags CRUD view completion
There are basic CRUD views already I think, but a create view is missing. Currently Tags are created directly by adding them via the quest or badge form.

Also missing is an edit view. There should be the ability to edit the name of a tag from the lost view (just like all our other CRUD lost views)

Finally a delete tag button with confirmation page (confirmation page lists all items that will lose the tag of deleted, similar to tag detail view)

Admin buttons on right of the list view table, just like other list views.

9921622:
Tags are accessed from the admin menu

create/update:
image

detail view:
image

delete view:
image

### DIFF
--- a/src/tags/forms.py
+++ b/src/tags/forms.py
@@ -1,4 +1,9 @@
+from django import forms
+from django.utils.text import slugify 
+
 from dal import autocomplete
+
+from taggit.models import Tag
 
 
 class BootstrapTaggitSelect2Widget(autocomplete.TaggitSelect2):
@@ -14,3 +19,17 @@ class BootstrapTaggitSelect2Widget(autocomplete.TaggitSelect2):
         css = {
             'all': ('https://cdnjs.cloudflare.com/ajax/libs/select2-bootstrap-theme/0.1.0-beta.10/select2-bootstrap.min.css',)
         }
+
+
+class TagForm(forms.ModelForm):
+
+    class Meta:
+        model = Tag
+        fields = ['name']
+
+    def save(self, *args):
+        # will be saved on super func, assuming commit=True
+        # necessary since slug is unique=True, so we use unique name to generate a new slug to prevent possible duplicates
+        self.instance.slug = slugify(self.cleaned_data['name'])  
+
+        return super().save(*args)
--- a/src/tags/templates/tags/detail.html
+++ b/src/tags/templates/tags/detail.html
@@ -9,24 +9,9 @@
 {% endblock %}
 
 {% block content %}
-<h3>Quests ({{ quests|length }})</h3>
-<ul>
-    {% for quest in quests %}
-        <li>
-            <a href="{{ quest.get_absolute_url }}">{{ quest }}</a>
-        </li>
-    {% empty %} <li>None</li>
-    {% endfor %}
-</ul>
-<h3>Badges ({{ badges|length }})</h3>
-<ul>
-    {% for badge in badges %}
-        <li>
-            <a href="{{ badge.get_absolute_url }}">{{ badge }}</a>
-        </li>
-    {% empty %} <li>None</li>
-    {% endfor %}
-</ul>
+
+<br>
+{% include 'tags/snippets/related-xp-item.html' with type='default' %}
 
 {% endblock %}
 
--- a/src/tags/views.py
+++ b/src/tags/views.py
@@ -1,11 +1,16 @@
-
+from django.urls import reverse_lazy
+from django.utils.decorators import method_decorator
 from django.contrib.auth.mixins import LoginRequiredMixin
-# from django.urls import reverse_lazy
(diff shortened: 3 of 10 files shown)


src/tags/templates/tags/detail.html
```diff
@@ -9,24 +9,32 @@
...
+<br>
+
+<div class="panel panel-default">
+    <div class="panel-heading"><h4>Quests ({{ quests|length }}) <i class="fa fa-shield fa-fw"></i></h4></div>
+    <div class="panel-body">
+        <p></p>
+        <ul>
+            {% for quest in quests %}
+                <li><a href="{{ quest.get_absolute_url }}">{{ quest }}</a></li>
+            {% endfor %}
```
tylerecouture, line 21: Missing the empty case (see deleted code)

src/tags/templates/tags/detail.html
```diff
@@ -9,24 +9,32 @@
...
+
+
+<div class="panel panel-default">
+    <div class="panel-heading"><h4>Badges ({{ badges|length }}) <i class="fa fa-certificate fa-fw"></i></h4></div>
+    <div class="panel-body">
+        <p></p>
+        <ul>
+            {% for badge in badges %}
+                <li><a href="{{ badge.get_absolute_url }}">{{ badge }}</a></li>
+            {% endfor %}
```
tylerecouture, line 34: Empty case

src/tags/forms.py
```diff
@@ -14,3 +19,23 @@ class Media:
...
+class TagForm(forms.ModelForm):
+
+    class Meta:
+        model = Tag
+        fields = '__all__'
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.fields['slug'].required = False
```
tylerecouture, line 33: What's the purpose of adding the slug field? Can we remove it? I don't think the user needs to know about it at all, if it's required for the form, can we just hide it?

src/tags/views.py
```diff
@@ -33,48 +38,57 @@
...
 
     def get_context_data(self, **kwargs):
-        kwargs['quests'] = Quest.objects.filter(tags__name=self.object.name)
-        kwargs['badges'] = Badge.objects.filter(tags__name=self.object.name)
+        ctx = super().get_context_data(**kwargs).copy()
+        ctx['quests'] = Quest.objects.filter(tags__id=self.object.id)
+        ctx['badges'] = Badge.objects.filter(tags__id=self.object.id)
+        return ctx
 
-        return super().get_context_data(**kwargs)
```
tylerecouture, line 48: How come you changed the structure of this?

tylerecouture (changes requested):
A few minor comments. Also:

 from the screenshots, the styling doesn't match between the detail vs delete view, can you make them match? Perhaps put that code into a template snippet so you can call common code from both instead of duplicating code. (Keep it DRY)

src/tags/forms.py
9921622: its set as unique=True and non optional (at least in the github version)
image
ill remove it from the form, but set it in the save func
image

src/tags/views.py
9921622: i actually have no idea, reverted

9921622:
added requested changes

detail view
image
delete view
image


update view without slug field
image

9921622:
forgot to link this to issue #1120

tylerecouture (changes requested):
Awesome!  Two small things:

1. For consistancy across the other apps, can you call the directory "snippets" instead of "partials"?

2. Can you add a quick test to check that the admin buttons aren't displayed to non-staff users?

9921622:
Added requested changes

src/tags/views.py
```diff
@@ -33,48 +38,56 @@
...
 
-# @method_decorator(staff_member_required, name='dispatch')
-# class TagDelete(NonPublicOnlyViewMixin, DeleteView):
-#     model = MenuItem
-#     success_url = reverse_lazy('utilities:menu_items')
+    def get_context_data(self, **kwargs):
+        ctx = super().get_context_data(**kwargs).copy()
+        ctx['quests'] = Quest.objects.filter(tags__id=self.object.id)
+        ctx['badges'] = Badge.objects.filter(tags__id=self.object.id)
+        return ctx
```
tylerecouture, line 93: this one still has the alternate implementation.  doesn't really matter, but might as well be consistent?

tylerecouture (approved):
Still one inconsistent context_data_method, but I'll approve.

src/tags/views.py
9921622: doesnt really matter since my next pr changes this

PR #1124, opened 2022-08-03
Feat: XP tag breakdown in user profile; Closes #1121

### LINKED ISSUE #1121: XP by tag breakdown in user profile.
Add a new section in a user profile, below the badges, that lists all the tags (that the user has earned XP in) on the left and how much XP they have earned on the right.

Display it similar to the quest section, that lists quests on the left and date on the right, and links to the submission.

9921622:
Not sure if its necessary to add a test for views since its only 1 line + i added a test for the ``get_tags_from_user`` in ``tag.models`` and there were lot of tests for ``total_xp_by_tags``

### DIFF
--- a/src/profile_manager/views.py
+++ b/src/profile_manager/views.py
@@ -17,6 +17,7 @@ from badges.models import BadgeAssertion
 from courses.models import CourseStudent
 from notifications.signals import notify
 from quest_manager.models import QuestSubmission
+from tags.models import get_user_tags_and_xp
 from tenant.views import NonPublicOnlyViewMixin, non_public_only_view
 
 from django.contrib.auth.forms import SetPasswordForm
@@ -114,6 +115,7 @@ class ProfileDetail(NonPublicOnlyViewMixin, DetailView):
         context['completed_past_submissions'] = QuestSubmission.objects.all_completed_past(profile.user)
         context['xp_per_course'] = profile.xp_per_course()
         context['badge_assertions_dict_items'] = BadgeAssertion.objects.badge_assertions_dict_items(profile.user)
+        context['tags'] = get_user_tags_and_xp(profile.user)
 
         # earned_assertions = BadgeAssertion.objects.all_for_user_distinct(profile.user)
         # assertion_dict = defaultdict(list)
--- a/src/tags/models.py
+++ b/src/tags/models.py
@@ -2,9 +2,36 @@ from django.apps import apps
 from django.db import models
 from django.db.models import Sum
 
+from taggit.models import Tag
 from taggit.managers import TaggableManager
 
 from siteconfig.models import SiteConfig
+
+
+def get_tags_from_user(user):
+    """
+    Returns all tags from quest and badges that are linked to user
+
+    Args:
+        user (User model): user model
+
+    Returns:
+        list[Tag] : list of Tag models
+    """
+    # get models through here to prevent circular imports
+    QuestSubmission = apps.get_model("quest_manager", "QuestSubmission")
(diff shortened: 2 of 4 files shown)


tylerecouture (commented):
Can you provide a screenshot of what this looks like?

9921622:
image

tylerecouture:
Thanks, this is much easier. Can you capitalize XP and remove total and parentheses, e.g.

TAG1        25 XP
TAG2        45 XP
Etc....

And rename the title of the section from "Tags" to "XP earned by Tag"

9921622:
Added requested changes
image

src/profile_manager/views.py
```diff
@@ -115,6 +116,9 @@ def get_context_data(self, **kwargs):
         context['xp_per_course'] = profile.xp_per_course()
         context['badge_assertions_dict_items'] = BadgeAssertion.objects.badge_assertions_dict_items(profile.user)
 
+        # get tags and its xp that are related to user + sort by xp total
+        context['tags'] = sorted([(tag, total_xp_by_tags(profile.user, [tag])) for tag in get_tags_from_user(profile.user)], key=lambda el: el[1])
```
tylerecouture, line 120: Can you move this to a method on the Tags app, and preferably, expand it out so it is commented, readable, and tested? So much code crammed into a single line is unreadable.

I would much prefer loops and variables than everything crammed into a single line.  Space is free, there is usually little gained by trying to cram more into less space, and it usually violated the implicit vs explicit and simple vs complex.

src/tags/models.py
```diff
@@ -7,6 +7,29 @@
...
+        user (User model): user model
+
+    Returns:
+        list[Tag] : list of Tag models
+    """
+    # get models through here to prevent circular imports
+    QuestSubmission = apps.get_model("quest_manager", "QuestSubmission")
+    BadgeAssertion = apps.get_model("badges", "BadgeAssertion")
+
+    quest_tags = [name for submission in QuestSubmission.objects.filter(user=user).order_by('quest__id').distinct('quest__id')
```
tylerecouture, line 24: This code seems similar to the previous method you wrote (same as below lines). Would it make sense to put these into their own method and call from both instead of duplicating the code? I'm not sure, so please just take a look and do it if it makes sense.  Let's keep it DRY where we can.

tylerecouture (changes requested):
Some code readability and DRY suggestions

src/tags/models.py
9921622: django-fied the code
image

9921622: This is also a possibility
image

src/profile_manager/views.py
9921622: done 
image

src/tags/models.py
9921622: im going to go with the second img for now

src/profile_manager/views.py
tylerecouture: Awesome! 

From your code comments and docstring it's not clear HOW the data is sorted. (By XP? By alphabetical tag names?)

src/tags/models.py
tylerecouture: First! Way easier to read!

src/profile_manager/views.py
9921622: added ``Sorted by total xp in descending order``

added ``(descending order)`` to ``# sort by total xp then return``

9921622:
added requested changes + test for ``get_user_tags_and_xp``

tylerecouture (approved):

PR #1130, opened 2022-08-06
add new tenant defaults; closes #1107

### LINKED ISSUE #1107: default new deck values
- max students should be 5
- trial period end date should be now+60 days

Dylan-RG-Patterson:
image
changed default active user limit and added a default trial-end-date of 60 days after tenant creation.


src/tenant/models.py
```diff
@@ -44,15 +45,15 @@ class Tenant(TenantMixin):
...
-        default=40,
+        default=5,
         help_text="The maximum number of users that can be active on this deck; -1 = unlimited."
     )
     max_quests = models.SmallIntegerField(
         default=100,
         help_text="The maximum number of quests that can be active on this deck (archived quests are considered inactive); -1 = unlimited."
     )
     trial_end_date = models.DateField(
-        blank=True, null=True,
```
tylerecouture, line 55: this should still be allowed to be blank. In fact it will be blank if they aren't in a Trial.

src/tenant/models.py
```diff
@@ -44,15 +45,15 @@ class Tenant(TenantMixin):
...
+        default=(timezone.now() + timezone.timedelta(days=60)),
```
tylerecouture, line 56: This won't work because the function will be called when the server starts up, not when the tenant is created.
Fro example compare: [`default=timezone.now` vs `default = timzone.now()`](https://stackoverflow.com/questions/38237777/django-timezone-now-vs-timezone-now)  (I may have come across this before ;)

src/tenant/tests/test_admin.py
```diff
@@ -120,7 +121,7 @@ def setUp(self):
             'max_active_users': 50,
             'max_quests': 100,
             # 'paid_until': None,
-            # 'trial_end_date': None
+            'trial_end_date': timezone.now()
```
tylerecouture, line 124: leave blank, should be able to be None

tylerecouture (changes requested):
Setting a model field default to a function is tricky (when is the function run?) So that will need to be fixed (probably) and properly tested.  To test, you'll need to fake the system time.

i.e. During a tenant creation test (just piggy back on one that already exists) set the time to some date in the future (before the tenant is created) then create the tenant, and make sure it's 60 day from that fake creation date, NOT 60 days from NOW (when the servers started running)

To set a fake time during testing (at anypoint) you can use `from [freezegun](https://github.com/spulec/freezegun#readme) import freeze_time`

Some examples:

decorator to set datetime for an entire test: /src/prerequisites/tests/test_signals:

```
@freeze_time('2018-10-12 00:54:00', tz_offset=0)
class PrerequisitesSignalsTest(TenantTestCase):
```

Context manager to fake time for specific code, here's an example testing that cache expires in the future, from siteconfig/test_models.py

```
    def test_SiteConfig_cache_expires(self):
        """ SiteConfig cache should expire after a certain period """

        # Cache should not be empty as of this moment
        self.assertIsNotNone(cache.get(SiteConfig.cache_key()))

        # Cache should now expire
        cache_time_expiration = localtime() + timedelta(days=1)
        with freeze_time(cache_time_expiration, tz_offset=0):
            self.assertIsNone(cache.get(SiteConfig.cache_key()))
```

src/tenant/models.py
```diff
@@ -26,6 +29,10 @@ def check_tenant_name(name):
         raise ValidationError("Invalid string used for the tenant name.")
 
 
+def default_trial_end_date():
+    return date.today() + timedelta(days=60)
```
tylerecouture, line 33: Need to use Django's timezone.now() so that our TZ info is taken into account.  BUUUUT Considering Semester.last_day doesn't even do that, I'll make it a new issue.

tylerecouture (approved):

PR #1134, opened 2022-08-08
Feat: User specific Tag detail view; Closes #1125

### LINKED ISSUE #1125: User specific Tag detail view
Create a Tag detail view (or extend the current  Tag detail view) specific to a user:

- [x] Instead of displaying all quests and badges for the tag, only shows those earned by the student
- [x] no admin buttons obviously...
- [x] make sure the page h1 title also indicates the username of the specific user (TAG1 Tag Details for Username)
- [ ] link the Tags listed in a user profiles to these user specific Tag detail view
- [x] rest of the detail view should be the same as the existing detail view (consider moving common html into a template snippet, if you plan to use 2 different templates)

9921622:
- [x] Instead of displaying all quests and badges for the tag, only shows those earned by the student

 - [x] no admin buttons obviously...

 - [x] make sure the page h1 title also indicates the username of the specific user (TAG1 Tag Details for Username)
 - im assuming youre talking about the heading_inner template tag
 
 - [x] link the Tags listed in a user profiles to these user specific Tag detail view
 - See pr #1124

 - [x] rest of the detail view should be the same as the existing detail view (consider moving common html into a template snippet, if you plan to use 2 different templates)

### DIFF
--- a/src/tags/templates/tags/detail.html
+++ b/src/tags/templates/tags/detail.html
@@ -5,7 +5,7 @@
 {% endblock %}
 
 {% block heading_inner %}
-    Tag: {{object.name}}
+    {{object.name}} Tag Details {% if user_obj %}for {{user_obj.username}}{% endif %}
 {% endblock %}
 
 {% block content %}
--- a/src/badges/templates/badges/detail.html
+++ b/src/badges/templates/badges/detail.html
@@ -28,8 +28,11 @@
         </div>
         <div class="col-sm-6">
           <ul class="list-unstyled">
-            {% with badge as object %}
-            <li>Tags: {% include 'tags/snippets/tag-list.html' %}{% endwith %}<br></li>
+            {% if not request.user.is_staff %}
+            <li>Tags: {% include 'tags/snippets/tag-list.html' with object=badge user_obj=request.user %}<br></li>
+            {% else %}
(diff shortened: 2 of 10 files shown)


src/tags/templates/tags/detail.html
```diff
@@ -5,7 +5,7 @@
 {% endblock %}
 
 {% block heading_inner %}
-    Tag: {{object.name}}
+    {{object.name}} Tag Details {% if not request.user.is_staff %}for {{request.user.username}}{% endif %}
```
tylerecouture, line 8: Hmmm..,... what if a staff member is viewing a student-specific tag page?  I think this needs more seperation.  url would be something like `/tags/<tag>/<user-id>/` or something like that and the user would be passed in the context (seperately from the viewer / request.user)

9921622:
staffv iew
image

student view
image

tylerecouture (changes requested):
user-specific Tag view will need its own url and view, though maybe the view can just extend the default TagDetail view with different queryset?

src/tags/templates/tags/detail.html
9921622: made a separate view for this
image

9921622:
added requested changes + made it so that all references to tag-list.html have direct to TagDetailStudent or TagDetailStaff depending on page and whos viewing said page

PR #1139, opened 2022-08-09
Feat: Copying a quest or Badge copies tags; Closes #1135

### LINKED ISSUE #1135: Copying a quest or Badge copies tags
By default, a quest or badge that is copied should retain the Tags of the copied object

9921622:
Quest Copy:
image

Badges Copy:
image

### DIFF
--- a/src/badges/tests/test_views.py
+++ b/src/badges/tests/test_views.py
@@ -30,7 +30,8 @@ class BadgeViewTests(ViewTestUtilsMixin, TenantTestCase):
         # needed because BadgeAssertions use a default that might not exist yet
         self.sem = SiteConfig.get().active_semester
 
-        self.test_badge = baker.make(Badge, xp=5)
+        self.test_badge = baker.make(Badge, name="badge", xp=5)
+        self.test_badge.tags.add('tag')
         self.test_badge_type = baker.make(BadgeType)
         self.test_assertion = baker.make(BadgeAssertion)
 
@@ -123,7 +124,19 @@ class BadgeViewTests(ViewTestUtilsMixin, TenantTestCase):
         new_badge = Badge.objects.latest('datetime_created')
         self.assertEqual(new_badge.name, "badge test")
 
-    def test_badge_copy(self):
+    def test_badge_copy__GET(self):
+        """ initial values in form GET is the same as the self.test_badge (badge that is being copied)  """
+        self.client.force_login(self.test_teacher)
+        response = self.client.get(reverse('badges:badge_copy', args=[self.test_badge.id]))
+        
+        form_data = response.context['form'].initial
+
+        # Badge name should have changed
+        self.assertEqual(form_data['name'], "Copy of " + self.test_badge.name)
+        # Tags should be the same as original
+        self.assertEqual(list(form_data['tags'].values_list('name', flat=True)), ['tag'])
+
+    def test_badge_copy__POST(self):
         # log in a teacher
         success = self.client.login(username=self.test_teacher.username, password=self.test_password)
         self.assertTrue(success)
--- a/src/badges/views.py
+++ b/src/badges/views.py
@@ -95,12 +95,13 @@ def badge_create(request):
 @non_public_only_view
 @staff_member_required
 def badge_copy(request, badge_id):
+    original_badge = get_object_or_404(Badge, pk=badge_id)  # prevent: Badge objects need to have a primary key value before you can access their tags
     new_badge = get_object_or_404(Badge, pk=badge_id)
     new_badge.pk = None  # autogen a new primary key (quest_id by default)
     new_badge.import_id = uuid.uuid4()
     new_badge.name = "Copy of " + new_badge.name
 
-    form = BadgeForm(request.POST or None, instance=new_badge)
+    form = BadgeForm(request.POST or None, instance=new_badge, initial={'tags': original_badge.tags.all()})
     if form.is_valid():
         form.save()
         return redirect('badges:list')
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -1234,27 +1234,28 @@ class QuestCopyViewTest(ViewTestUtilsMixin, TenantTestCase):
         self.client = TenantClient(self.tenant)
         self.test_student = User.objects.create_user('test_student', password="password")
         self.test_teacher = User.objects.create_user('test_teacher', password="password", is_staff=True)
+
         self.quest = baker.make(Quest, name="Test Quest")
+        self.quest.tags.add('tag')
 
         # simulate a logged in TA (teaching assistant = a student with extra permissions)
         self.test_ta = User.objects.create_user('test_ta')
         self.test_ta.profile.is_TA = True  # profiles are create automatically via User post_save signal
         self.test_ta.profile.save()
 
-        self.valid_copy_form_data = {
-            'name': "Test Quest - COPY",  # only blank required field
-            # these fields are required but they have defaults
-            'xp': 0,
-            'max_repeats': 0,
-            'max_xp': -1,
-            'hours_between_repeats': 0,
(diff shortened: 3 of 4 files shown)


src/badges/views.py
```diff
@@ -106,12 +105,22 @@ def badge_create(request):
 @non_public_only_view
 @staff_member_required
 def badge_copy(request, badge_id):
-    new_badge = get_object_or_404(Badge, pk=badge_id)
-    new_badge.pk = None  # autogen a new primary key (quest_id by default)
-    new_badge.import_id = uuid.uuid4()
-    new_badge.name = "Copy of " + new_badge.name
+    badge = get_object_or_404(Badge, pk=badge_id)
+    
+    form_data_initial = model_to_dict(badge)
```
tylerecouture, line 110: I think you should leave the old code.  Why change to this?  Also, you are now copying the uuid I think, which needs to be unique (and if your code passes tests, then there is a missing test that needs to ensure a copy has a unique ID!)

Note if the docs say this is the prefered way to doing it, link to docs and we can do it this way, otherwise, for consistancy throughout the project, setting the id to None is the way the rest of the project copies model objects.

src/badges/views.py
```diff
@@ -106,12 +105,22 @@ def badge_create(request):
...
+    form_data_initial['tags'] = badge.tags.all()
+
+    # rename copied badge
+    # strips front if len(name + ' - copy') is too long
+    char_limit = BadgeForm._meta.model._meta.get_field('name').max_length
```
tylerecouture, line 115: I think this is all unnecessary.  The badge form will show a super long name, but the form won't submit and the user can change the name if too long.  Not a problem and leaves the full name for the user to decide what to change

src/quest_manager/views.py
```diff
@@ -184,17 +184,25 @@ class QuestCopy(QuestCreate):
...
+
+        # rename copied quest
+        # strips front if len(name + ' - copy') is too long
+        char_limit = self.model._meta.get_field('name').max_length
+        copy_text = ' - Copy'
+        name = quest.name
 
-        new_quest = get_object_or_404(Quest, pk=self.kwargs['quest_id'])
-        new_quest.pk = None  # autogen a new primary key (quest_id by default)
-        new_quest.import_id = uuid.uuid4()
```
tylerecouture, line 194: Please revert to old copy code unless there's a specific reason for changing.  Also, please add test to check that a copied quest has a unique UUID!!  Does this pass tests?  I thought that field was unique so I'm surprised the form can even be submitted without a new UUID, unless it's auto generated in the background?

src/quest_manager/tests/test_views.py
```diff
@@ -265,6 +266,26 @@ def test_student_no_quests_help_text(self):
         self.assertContains(response, "No quests are available.")
         self.assertContains(response, "There are currently no new quest available to you!")
 
+    def test_QuestCopy__correct_clone(self):
```
tylerecouture, line 269: No docstring... what is this testing?

src/quest_manager/views.py
```diff
@@ -184,17 +184,25 @@ class QuestCopy(QuestCreate):
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
 
+        quest = get_object_or_404(Quest, pk=self.kwargs['quest_id'])
```
tylerecouture, line 187: why change name from copied_quest to quest?  less info and more ambiguous as to what quest this variable holds.

src/quest_manager/views.py
```diff
@@ -184,17 +184,25 @@ class QuestCopy(QuestCreate):
...
+
         # by default, set the quest this was copied from as the new_quest_prerequisite
         # If this is changed in the form it will be overwritten in form_valid() from QuestFormViewMixin
-        copied_quest = get_object_or_404(Quest, pk=self.kwargs['quest_id'])
-        kwargs['initial']['new_quest_prerequisite'] = copied_quest
+        kwargs['initial']['new_quest_prerequisite'] = quest
+
+        # rename copied quest
+        # strips front if len(name + ' - copy') is too long
+        char_limit = self.model._meta.get_field('name').max_length
```
tylerecouture, line 198: Not needed. If it is, let me know why (there are no code comments explaining....)

src/badges/tests/test_views.py
```diff
@@ -223,6 +224,25 @@ def test_bulk_assertion_create(self):
         badge_assertions_after = BadgeAssertion.objects.all().count()
         self.assertEqual(badge_assertions_after, badge_assertions_before + 2)
 
+    def test_BadgeCopy__correct_clone(self):
```
tylerecouture, line 227: Needs docsting explaining explicitly what this is testing.

tylerecouture: Remember when tests fail, you see the docstring in the test error message, so docstrings in tests are important to help future devs troubleshoot failing tests.

tylerecouture (changes requested):
Changes (or better documentation) requested. Thanks!

src/quest_manager/views.py
9921622: because its not the copied_quest its the original quest. Im not cloning the quest, like in the original code because there is no need to. Im passing the same values from quest to the initial form values

src/quest_manager/views.py
9921622: no need to create a new instance of quest (old code is also calling get_object_or_404(Quest, pk=self.kwargs['quest_id']) twice)

i think i just forgot to get new uuid. since i deleted everything in old code because i had a dynamic function that would make a new instance of a model in ``TagModelMixin``

src/quest_manager/views.py
9921622: removed

9921622: removed

src/badges/views.py
9921622: removed

src/badges/views.py
9921622: i dont think theres a need to copy the model if we dont need to do it. Old code also wont work by adding tags to instance. 

using ``initial`` over ``instance`` seems to be the proper way to populate the form fields anyway. As there really isn't a mention of ``instance`` arg in the docs
https://docs.djangoproject.com/en/4.1/ref/forms/api/

9921622: also uuid does not matter since im not using an instance of Quest, just passing args to initial data

tylerecouture: Why won't the old code work?  You just need to make sure you save the model object first (so it gets an ID and is added to the DB), then you can add the tags.

tylerecouture: I don't like this `form_data_initial = model_to_dict(badge)` because it's implicit.  I like explicit.  Zen of Python!

src/quest_manager/views.py
tylerecouture: Why change it though?

tylerecouture: I don't mind refactoring, but there's no documentation why this is better?

src/badges/views.py
9921622: if i save the model obj wont that defeat the purpose of QuestCopy being apart of CreateView?

what if i want to back out of quest copy creation, then the model will still be in the db.

if there is already an instance of ``Quest - Copy`` and i copy ``Quest`` then it will give me a ``404`` because ``Quest - Copy`` already exists.

9921622: also this doesnt work
```
new_quest.tags = copied_quest.tags
kwargs['instance'] = new_quest
```

src/quest_manager/tests/test_views.py
9921622: turns out theres a whole TestClass dedicated to quest copy

9921622:
added requested changes

tylerecouture (approved):

PR #1148, opened 2022-08-10
Feat: Access to list of ALL users, including staff, and inactive; Closes #1137

### LINKED ISSUE #1137: Access to list of ALL users, including staff, and inactive
Currently staff can only view students (in the Student view)

Need two additional views for profiles/users, that can be added to the end of "students" as a submenu:

- [ ] Staff users (just like student view but only shows staff
- [ ] Inactive users (users that have been deactivated/retired/graduated) so that they can be reactivated if needed (this will need some testing, I don't know if you can even view the profile of an inactive user? they may not be visible to the "get" or "filter".  If not, then they just need a single button to reactivate,  in the "inactive users" list view.

9921622:
- [x]   Staff users (just like student view but only shows staff
- [x] Inactive users (users that have been deactivated/retired/graduated) so that they can be reactivated if needed (this will need some testing, I don't know if you can even view the profile of an inactive user? they may not be visible to the "get" or "filter". If not, then they just need a single button to reactivate, in the "inactive users" list view.
- you are able to access the profiles of inactive students, i think youre thinking of logging in as an inactive account

inactive list
image

staff list
image


src/profile_manager/templates/profile_manager/snippets/staff_list.html
```diff
@@ -0,0 +1,121 @@
...
+                <td>
+                    {{ object.user.last_login|date:"y/m/d" }} {{ object.user.last_login|time:"H:i:s"}}
+                </td>
+                <!-- end `last_login_datetime` -->
+                <td><a href="{% url 'profiles:profile_detail' object.id %}">{{ object.user.username|truncatechars:14 }}</a></td>
+
+                {#      Admin buttons #}
+                <td>
+                    <div class="btn-group action-btn-group">
+                        <a href="{% url 'profiles:comment_ban_toggle' object.id %}"
```
tylerecouture, line 97: Comment ban button should be removed for staff

src/profile_manager/templates/profile_manager/snippets/staff_list.html
```diff
@@ -0,0 +1,121 @@
...
+                            title="{% if object.banned_from_comments %}Remove comment ban
+                        {% else %}Ban student from public comments{% endif %}"
+                            class="btn btn-sm {% if object.banned_from_comments %}btn-success{% else %}btn-default{% endif %}">
+                            <i class="fa fa-commenting-o"></i>
+                        </a>
+                        <a href="{% url 'profiles:xp_toggle' object.id %}"
```
tylerecouture, line 103: XP toggle is irrelevant for staff too

src/profile_manager/templates/profile_manager/snippets/staff_list.html
```diff
@@ -0,0 +1,121 @@
...
+                            title="{% if object.not_earning_xp %}Turn quest skipping off
+                        {% else %}Allow user to skip quests. Gives user a button allowing them skip quests without approval. Skipped quests count as completed for prerequisite purposes but don't grant XP.{% endif %}"
+                            class="btn btn-sm
+                    {% if object.not_earning_xp %}btn-success{% else %}btn-default{% endif %}">
+                            <i class="fa fa-magic"></i>
+                        </a>
+                        <a href="{% url 'profiles:change_password' object.user.pk %}" class="btn btn-sm btn-default"
```
tylerecouture, line 110: Hmmm... since we have some logic that prevents staff from changing other staff passwords (I think?) maybe remove this too, can already access from the profile as needed. No need for the quick button here..

src/profile_manager/templates/profile_manager/snippets/staff_list.html
```diff
@@ -0,0 +1,121 @@
...
+                    <small>{{ object.user.last_login|timesince }} ago</small>
+                </td>
+                <!-- `last_login_datetime` -->
+                <td>
+                    {{ object.user.last_login|date:"y/m/d" }} {{ object.user.last_login|time:"H:i:s"}}
+                </td>
+                <!-- end `last_login_datetime` -->
+                <td><a href="{% url 'profiles:profile_detail' object.id %}">{{ object.user.username|truncatechars:14 }}</a></td>
+
+                {#      Admin buttons #}
```
tylerecouture, line 94: See comments below, but I think the admin buttons can be removed when viewing staff list

src/profile_manager/views.py
```diff
@@ -24,10 +24,22 @@
 from django.contrib.auth.forms import SetPasswordForm
 
 
+class ViewTypes:
```
tylerecouture, line 27: I like this.  Could probably be used to good effect in other areas that hack this by looking at the url. Can you add a new issue to "Find and refactor templates that look at the url for setting active buttons" and provide this example for how to do it

src/profile_manager/views.py
```diff
@@ -59,19 +77,36 @@ class ProfileListCurrent(ProfileList):
...
+
+    def get_queryset(self):
+        profiles_qs = Profile.objects.filter(user__is_staff=True)
+        return self.queryset_append(profiles_qs)
+
+
+class ProfileListInactive(ProfileList):
+    view_type = ViewTypes.INACTIVE
+
+    # override the staff requirement for ProfileList
```
tylerecouture, line 103: This should be staff only too.

src/profile_manager/views.py
```diff
@@ -59,19 +77,36 @@ class ProfileListCurrent(ProfileList):
...
+    def test_func(self):
+        return self.request.user.is_authenticated
+
+    def get_queryset(self):
+        profiles_qs = Profile.objects.all_students().filter(user__is_active=False)
+        return self.queryset_append(profiles_qs)
```
tylerecouture, line 109: why append and not just return profile_qs? What does the qs contain at this point that you're appending to? (same question for above get_queryset() methods

src/profile_manager/views.py
```diff
@@ -43,9 +55,15 @@ def queryset_append(self, profiles_qs):
         return profiles_qs
 
     def get_queryset(self):
-        profiles_qs = Profile.objects.all_students()
+        profiles_qs = Profile.objects.all_students().filter(user__is_active=True)
```
tylerecouture, line 58: I think a better way to do this is create a new Profile manager method all_inactive().  Then you can test the manager method directly, and if we use it elsewhere it's already tested and consistent.  Prefer to minimize filtering outside of manager methods, so we can reuse common/obvious filters (DRY!).

Also, this should be ALL inactive users, not just inactive students.

Ditto for other filters below.

src/profile_manager/views.py
```diff
@@ -59,19 +77,36 @@ class ProfileListCurrent(ProfileList):
...
     """
+    view_type = ViewTypes.CURRENT
 
     # override the staff requirement for ProfileList
     def test_func(self):
         return self.request.user.is_authenticated
 
     def get_queryset(self):
-        profiles_qs = Profile.objects.all_for_active_semester()
+        profiles_qs = Profile.objects.all_for_active_semester().filter(user__is_active=True)
```
tylerecouture, line 87: I think this is already the case?  did it show inactive users before you added the filter? I don't think so.

src/profile_manager/views.py
```diff
@@ -59,19 +77,36 @@ class ProfileListCurrent(ProfileList):
...
         return self.queryset_append(profiles_qs)
 
-    def get_context_data(self, **kwargs):
-        context = super(ProfileListCurrent, self).get_context_data(**kwargs)
-        context['current_only'] = True
-        return context
+
+@method_decorator(staff_member_required, name='dispatch')
+class ProfileListStaff(ProfileList):
+    view_type = ViewTypes.STAFF
```
tylerecouture, line 93: Nicely done (Inheritance and the enum)

Doesn't Profile List already implement a staff member required test_func?  Why the duplication?

tylerecouture (changes requested):
Some ntes/questions/changes =)

src/profile_manager/templates/profile_manager/profile_list.html
```diff
@@ -11,186 +11,28 @@
     {% if request.user.is_staff %}
         <div class="btn-group">
             <a href="{% url 'profiles:profile_list_current' %}"
-            class="btn {% if current_only %}btn-primary{% else %}btn-default{% endif %}">Current Semester</a>
+            class="btn {% if view_type == VIEW_TYPES.CURRENT %}btn-primary{% else %}btn-default{% endif %}">Current Semester</a>
 
             <a href="{% url 'profiles:profile_list' %}"
-            class="btn {% if not current_only %}btn-primary{% else %}btn-default{% endif %}">All</a>
```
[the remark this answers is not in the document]
9921622: done

src/profile_manager/templates/profile_manager/snippets/staff_list.html
9921622: removed

src/profile_manager/templates/profile_manager/snippets/staff_list.html
9921622: removed

src/profile_manager/templates/profile_manager/snippets/staff_list.html
9921622: removed

src/profile_manager/views.py
9921622: added

src/profile_manager/views.py
9921622: #1150 added

src/profile_manager/views.py
9921622: removed for both

src/profile_manager/views.py
9921622: added function, but is there really a need to test a single arg filter?
image
image

src/profile_manager/views.py
9921622: despite being set as inactive, test_user still shows up in current semester
image
image
image

tylerecouture: Dang it!  This needs to be fixed in all_for_active_semester(), it should already be doing that filtering. It should NOT include inactive students.

Can you also add a test to the all_for_active_semester() method that verifies inactive students are not included? Thanks!  And nice catch.

tylerecouture (changes requested):
1. Request a fix and test of old bad code of mine =)
2.  There is a LOT of duplication between staff_list and student_list templates.  I don't really fully utilize djangos template system, but I htink this is an opportunity to do that.  If you have to change the same thing in two different places... its just asking for inconsistancies (because assumption is code should only need to be changed in one place!


For example, if I want to change how the alis or last name is displayed, a dev might change one of them without realize it's coded seperately somewhere else for the identical thing.  Keep it DRY!

Feel free to do some research on maximizing the use of django's template system with includes etc.

9921622:
i dont think theres a way to really simplify the list templates without putting if statements everywhere

i tried an approach similar to crispy forms layout where i put everything in json format with custom filters to render it
image
but the list was too inconsistent with how it rendered everything out

src/profile_manager/views.py
9921622: added test_managers.py

9921622:
Added requested changes

9921622:
oops forgot to request review

tylerecouture (approved):

PR #1158, opened 2022-08-12
add campaign active value functionality; closes #1083

### LINKED ISSUE #1083: What happens if a Camapign has Active=False?
Does it do anything?  If not, yes, what does it do? Where is that field used int the app's logic?

There's no help text, so if it does something, maybe add some help text.

Does it prevent that option from appearing in the quest form's campaign field?

Does it prevent the campaign from appearing in maps?  Does it make all the quests within the campaign act as if they were inactive (if it doesn't... which it probably doesn't, maybe it should?!

Dylan-RG-Patterson:
added functionality to the active class variable for campaigns. previously ineffectual, now quests that are a part of inactive campaigns will be treated as inactive, meaning they won't appear in quest maps or a student's available quests tab.
added help text to signify this.
image


src/quest_manager/models.py
```diff
@@ -271,6 +279,13 @@ def not_expired(self):
 
     def visible(self):
         return self.filter(visible_to_students=True)
+    
+    def active_or_no_campaign(self):
+        """With self as an argument, returns a filtered queryset 
+        containing only quests in active campaigns or quests without campaigns.
+        """
+        active_campaigns = Category.objects.filter(active=True)
+        return (self.filter(Q(campaign__in=active_campaigns) | Q(campaign=None)))
```
tylerecouture, line 288: instead of getting all the active campaigns and checking if it's in the list, could you do something like this instead?  You can traverse foreign key fields using double underscores
`campaign__active=True`

Though, you still might need the Q object for the "OR"

ACTUALY!  Maybe you could exclude instead of filter, so it would just be: `self.exclude(campaign__active=False`  one line and much nicer.  However, it assumes that None will NOT be counted as False, which is probably not the case.

src/quest_manager/models.py
```diff
@@ -340,17 +355,18 @@ def get_queryset(self, include_archived=False):
...
-        7. Check for blocking quests (available and in-progress), if present, remove all others
+        3. Quests that are a part of an inactive campaign
+        4. Quests whose prerequisites have not been met
+        5. Quests that are not currently submitted for approval or already in progress
+        6. Quests whose maximum repeats have been completed
+        7. Quests whose repeat time has not passed since last completion
+        8. Check for blocking quests (available and in-progress), if present, remove all others
         """
         qs = self.get_active().select_related('campaign')  # exclusions 1 & 2
         qs = qs.get_conditions_met(user)  # 3
```
tylerecouture, line 372: The code comment references to the docstring seem to be off now.  Comment says # 3, but it's # 4 in your new docstring list.

tylerecouture (changes requested):
Very nicely done.
1. Don't forget to request a review when your PRs are ready, I missed this one.
2. A few very minor edits requested.
3. One significant part missing however: inactive campaigns should not appear in the quest form's campaigns selection widget.  Is that taken care of here?  The easiest way (which will magically fix all the forms etc) might be to set the [limit_choices_to](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.ForeignKey.limit_choices_to) value on the campaign foreign key

src/quest_manager/models.py
Dylan-RG-Patterson: code comments updated to reflect new numbering.
```
        qs = self.get_active().select_related('campaign')  # exclusions 1, 2 & 3
        qs = qs.get_conditions_met(user)  # 4
        available_quests = qs.not_submitted_or_inprogress(user)  # 5, 6 & 7

        available_quests = available_quests.block_if_needed(user=user)  # 8
```

src/quest_manager/models.py
Dylan-RG-Patterson: self.exclude(campaign__active=False) works perfectly with only one line, thank you. The control quest without a campaign returns active in testing all the same.

Dylan-RG-Patterson:
> Very nicely done.
> 
> 1. Don't forget to request a review when your PRs are ready, I missed this one.
> 2. A few very minor edits requested.
> 3. One significant part missing however: inactive campaigns should not appear in the quest form's campaigns selection widget.  Is that taken care of here?  The easiest way (which will magically fix all the forms etc) might be to set the [limit_choices_to](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.ForeignKey.limit_choices_to) value on the campaign foreign key

image

quest campaign form now only displays active campaigns (omitted in this screenshot is an inactive campaign)

image

src/quest_manager/models.py
tylerecouture: Awesome.

tylerecouture (approved):
Solid work here. Clean code, well commented. Thank you.

Issue #1153, opened 2022-08-10
DAL widget: show more than one object per model


tylerecouture:
Currently only shows one object per model (e.g. prerequisites editing).  So if I have a bunch of quests called "Blender 1" and "Blender 2" ... "Blender N" then typing "Blender" into the search only will show one of them instead of all of them (or at least a few... max 10?)

There must be a way to set how many to show per model.

![image](https://user-images.githubusercontent.com/10604391/184025874-36169a5a-0ea3-4c42-9ba9-ce3cc0917d7e.png)

Note, also fix Map DAL widget for same problem, if that issue has been completed at #1141

tylerecouture:
UPDATE NOTE: I suspect you'll have to dig down through the DAL code and find out where this decision is being made (how many to show) and override that method.

tylerecouture:
My ideal solution would just use django-select2 package so we could remove DAL. The reason DAL was used is because it comes with a Generic Foreign Key widget out of the box, but it has caused us nothing but problems (probably cache issues, since the values seem to load no problem in development, but in production (aws/elasticache) they are inconsistent.

Really all we need is something that can mimic what we grapelli already provides in Django Admin for generic foreign keys. I noticed on your github that you've created quite a few widgets. Perhaps you could just create a generic foreign key widget using Select2 and then implement it in our project?

PR #1168, opened 2022-08-16
Feat: Map create form uses DAL widget; Closes #1141

### LINKED ISSUE #1141: Map create form should use DAL widget
Use django autocomplete light widget for the map creation (initial object) Currently you have to know the ID of the object, after selecting the model.

Dal allows you to select the model, then will populate the other field with proper values that can be searched.

DAL is already used for prerequisites in a more complicated manner (as a formset).  This should be easier.  There are some utility methods that already exists, and each model should already have its own dal views you can tap into...maybe?

https://django-autocomplete-light.readthedocs.io/en/master/

9921622:
Ended up using the same DAL widget as prereq, ran into the same problems described in #1153 (except its 3 obj per model)

Also converted generate_map fbv to ScapeGenerateMap cbv

ScapeGenerateMapView:
image
Update View:
image

### DIFF
--- a/src/djcytoscape/templates/djcytoscape/generate_new_form.html
+++ b/src/djcytoscape/templates/djcytoscape/generate_new_form.html
@@ -1,19 +1,25 @@
 {% extends "djcytoscape/base.html" %}
 {% load crispy_forms_tags %}
 {% block heading_inner %}Generate a New Quest Map{% endblock%}
+
+{% block head %}
+  {{ form.media }}
+{% endblock %}
+
 {% block content %}
+
     {% if interlink %}
         <p>This quest map doesn't exist yet, but you can create it here.  Students who attempt to click the link that
         took you here will see a Page not found (404) error page until the map is created.</p>
 
         <p>Additional settings can be edited after the map has been generated.</p>
     {% endif %}
+    
     <form action="" method="post">
-    {% csrf_token %}
-    {{ form|crispy }}
-    <input type="submit" value="Generate a new Map starting from this Quest" class="btn btn-success"/>
-    <a href="{% url 'maps:index' %}" role="button" class="btn btn-danger">Cancel</a>
-</form>
+        {% csrf_token %}
+        {{ form|crispy }}
+        <input type="submit" value="Generate a new Map starting from this object." class="btn btn-success"/>
+        <a href="{% url 'maps:index' %}" role="button" class="btn btn-danger">Cancel</a>
+    </form>
+
 {% endblock %}
-
-
--- a/src/djcytoscape/tests/test_views.py
+++ b/src/djcytoscape/tests/test_views.py
@@ -7,7 +7,9 @@ from django_tenants.test.client import TenantClient
 from model_bakery import baker
 
 from djcytoscape.models import CytoScape
-from hackerspace_online.tests.utils import ViewTestUtilsMixin
+from djcytoscape.forms import GenerateQuestMapForm, QuestMapForm, get_model_options
+
+from hackerspace_online.tests.utils import ViewTestUtilsMixin, generate_form_data
 
 from .test_models import generate_real_primary_map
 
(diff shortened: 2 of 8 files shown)


9921622:
tests arent failing on my end. Looking at the error looks really similar to a migration error
image
image

9921622:
everything should work now

src/djcytoscape/templates/djcytoscape/generate_new_form.html
```diff
@@ -1,19 +1,25 @@
...
+  {{ form.media }}
+{% endblock %}
+
 {% block content %}
+
     {% if interlink %}
         <p>This quest map doesn't exist yet, but you can create it here.  Students who attempt to click the link that
         took you here will see a Page not found (404) error page until the map is created.</p>
 
         <p>Additional settings can be edited after the map has been generated.</p>
```
tylerecouture, line 15: Is this still true?  Create and Edit should pretty much be using the same form now, no?

src/djcytoscape/templates/djcytoscape/generate_new_form.html
```diff
@@ -1,19 +1,25 @@
...
-    {% csrf_token %}
-    {{ form|crispy }}
-    <input type="submit" value="Generate a new Map starting from this Quest" class="btn btn-success"/>
-    <a href="{% url 'maps:index' %}" role="button" class="btn btn-danger">Cancel</a>
-</form>
-{% endblock %}
-
+        {% csrf_token %}
+        {{ form|crispy }}
+        <input type="submit" value="Generate a new Map starting from this Quest" class="btn btn-success"/>
```
tylerecouture, line 21: Could be a badge or rank too, right?  My typo.   Maybe just say "from this object."

src/djcytoscape/tests/test_views.py
```diff
@@ -81,11 +83,62 @@ def test_all_page_status_codes_for_teachers(self):
...
+        # generate form datas
+        content_type = ContentType.objects.filter(CytoScape.ALLOWED_INITIAL_CONTENT_TYPES).first()
+        object_ = content_type.model_class().objects.first()
+
+        form_data = generate_form_data(model_form=GenerateQuestMapForm, name='New Name')
+        form_data.update({'initial_content_object': f'{content_type.id}-{object_.id}'}) 
+
+        # response tests
+        response = self.client.post(reverse('djcytoscape:generate_unseeded'), data=form_data)
+        self.assertEqual(response.status_code, 302)
```
tylerecouture, line 106: Can you verify in the test that it is redirecting where expected to?

src/djcytoscape/tests/test_views.py
```diff
@@ -81,11 +83,62 @@ def test_all_page_status_codes_for_teachers(self):
...
+
+        self.assertTrue(CytoScape.objects.filter(name=form_data['name']).exists())
+
+        map_ = CytoScape.objects.get(name=form_data['name'])
+        self.assertEqual(map_.initial_content_type, content_type)
+        self.assertEqual(map_.initial_object_id, object_.id)
+
+    def test_Form_get_model_options__correct_models(self):
+        """ Quick test to see if the hardcoded model list is equal to CytoScape.ALLOWED_INITIAL_CONTENT_TYPES """
```
tylerecouture, line 136: nice

src/djcytoscape/tests/test_views.py
```diff
@@ -81,11 +83,62 @@ def test_all_page_status_codes_for_teachers(self):
...
+
+        dynamically_loaded_models = [ct.model_class() for ct in ContentType.objects.filter(CytoScape.ALLOWED_INITIAL_CONTENT_TYPES)]
+        hard_coded_models = [option[0] for option in get_model_options()]
+
+        self.assertEqual(set(dynamically_loaded_models), set(hard_coded_models))
```
tylerecouture, line 141: Why set? set will remove doubles, which you probably don't want (i.e. if one of your list has a double, that's a problem and we want that captured by this test!)

Probably better to just make sure they are in the same order (I assume that's why you are using set?)  The order should be defined, so just make sure they match.

alternately, you can use Django's assertQuerysetEqual(qs1, qs_or_list2, ordered=False) if one of them is a queryset

src/djcytoscape/views.py
```diff
@@ -137,54 +132,59 @@ def primary(request):
...
+            self.interlink = True
+
+        return kwargs
+
+    def get_context_data(self, **kwargs):
+        ctx = super().get_context_data(**kwargs).copy()
+        ctx['interlink'] = self.interlink
+        return ctx
+
+    def form_valid(self, form):  # virtually the same as super().form_valid except we're caching form.save() obj for get_success_url()
```
tylerecouture, line 182: Good comment

tylerecouture: Can you add a message in here? Something like:
```
        messages.success(
            self.request,
            f"New map {self.object.name} was successfully generated."
        )
```

tylerecouture (changes requested):

src/djcytoscape/templates/djcytoscape/generate_new_form.html
9921622: fixed

src/djcytoscape/tests/test_views.py
9921622: im using a set since the order is of dynamically loaded models is dependent on its id in ContentTypes. So if someone wants to change the model order in hard_coded_models it doesnt fail the test.

Ill add an extra assert to check for duplicate models

src/djcytoscape/tests/test_views.py
9921622: done

src/djcytoscape/tests/test_views.py
9921622: done

src/djcytoscape/views.py
9921622: done

src/djcytoscape/templates/djcytoscape/generate_new_form.html
9921622: interlinked urls still works

is_primary_scape and autobreak are the additional settings can be edited after the map has been generated. They cant be set in generate_map_form

9921622:
Added requested changes

tylerecouture (approved):

PR #1174, opened 2022-08-18
Feat: Tags to Visual Representation; Closes #1155

### LINKED ISSUE #1155: Tags to Visual Representation
Tags Chart mockup!
In this case, the tags are related to math skills or learning outcomes.

9921622:
I put the tag chart with the other charts -> ``/courses/marks/``

chart wihout any quests:
image
chart with quests and badges
image

### DIFF
--- a/src/courses/templates/courses/chart_tags_progress.html
+++ b/src/courses/templates/courses/chart_tags_progress.html
@@ -1,0 +1,164 @@
+<script>
+
+    function InstantiateChart(labels, questDataset, badgeDataset) {
+        var COLORS = {
+            quest: {
+                {% if request.user.profile.dark_theme %}
+                // dark theme
+                bgColor: 'rgba(65, 65, 65, 1)', 
+                brColor: 'rgba(20, 20, 20, 0.75)'
+                {% else %}
+                // light theme
+                bgColor: 'rgba(140, 140, 140, 1)',
+                brColor: 'rgba(12, 12, 12, 0.5)'
+                {% endif %}
+            },
+            badge: {
+                {% if request.user.profile.dark_theme %}
+                // dark theme
+                bgColor: 'rgba(90, 90, 90, 1)', 
+                brColor: 'rgba(20, 20, 20, 0.75)'
+                {% else %}
+                // light theme
+                bgColor: 'rgba(191, 191, 191, 1)', 
+                brColor: 'rgba(64, 64, 64, 0.5)'
+                {% endif %}
+            },
+        }
+        
+        // generate datasets
+        datasets = []
+
+        // append Quest XP
+        for (var questData of questDataset) {
+            datasets.push({
+                label: questData.name,
+                data: questData.dataset,
+                backgroundColor: COLORS.quest.bgColor,
+                borderColor: COLORS.quest.brColor,
+                borderWidth: 1,
+                stack: 0,
+            });
+        }
+
+        // append Badge XP
+        for (var badgeData of badgeDataset) {
+            datasets.push({
+                label: badgeData.name,
+                data: badgeData.dataset,
+                backgroundColor: COLORS.badge.bgColor,
+                borderColor: COLORS.badge.brColor,
+                borderWidth: 1,
+                stack: 0,
+            });
+        }
+
+        // this is for labels to show
+        // this also needs to be on top of questDataset + badgeDataset stack
+        // to prevent visual bugs
+        datasets.push({
+            label: 'Quest XP',
+            backgroundColor: COLORS.quest.bgColor,
+            borderColor: COLORS.quest.brColor,
+            borderWidth: 1,
+            stack: 0,
+        });
+        datasets.push({
+            label: 'Badge XP',
+            backgroundColor: COLORS.badge.bgColor,
+            borderColor: COLORS.badge.brColor,
+            borderWidth: 1,
+            stack: 0,
+        });
+
+        // instantiate chart
+        var TCtx = document.getElementById('TagChart').getContext('2d');
+        var TagChart = new Chart(TCtx, {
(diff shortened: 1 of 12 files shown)


tylerecouture (changes requested):
This is sweet!  Can you put a slice for each quest completed, rather than for "Quests" as a whole? Same with Badges?  Bonus for putting a popup/tooltip/something that shows the name of quest and XP earned when you hover over it =D

9921622:
- [x] This is sweet! Can you put a slice for each quest completed, rather than for "Quests" as a whole? Same with Badges? Bonus for putting a popup/tooltip/something that shows the name of quest and XP earned when you hover over it =D

image
image

9921622:
i dont think get_quest_by_tag and get_badge_by_tag need tests since they are directly copied from total_xp_by_tags (same queryset, one is calling values_list, the other aggregate)

9921622:
ok i dont know why this specific test is failing. i didnt even touch this part of the codebase

image
image

tylerecouture:
> ok i dont know why this specific test is failing. i didnt even touch this part of the codebase
> 
> image image

Is it failing locally too? or just on github?

src/tags/models.py
```diff
@@ -34,21 +34,66 @@ def get_tags_from_user(user):
...
+
+    id_list = QuestSubmission.objects.all_approved(user, active_semester_only=True).filter(
+        is_completed=True, 
+        quest__tags__name__in=list(tags)
+    ).exclude(do_not_grant_xp=True).distinct().values_list('quest__id', flat=True)
+
+    return Quest.objects.filter(id__in=id_list)
+
+
+def get_badge_by_tag(user, tags):
```
tylerecouture, line 60: Problem with not having a test for this, is that its not the same code.  This code could be change and the other test wont fail.  However there is a solution to prevent duplicate tests.  Put the common code in its own method and then have both these methods call the common one, and the common method used by both can be fully tested.

If you can't put it in a common method, well, then they aren't the same and need their own tests.

Same with the other method.

tylerecouture: Taking a closer look, seems like you should put the common parts into new common methods that are tested (to ensure the proper values are returned)

Then each individual method here can have their own much simpler tests.  They can assume the proper values are returned, but might need to test that the aggregate was added in one, and that the values list was added in the other.

src/courses/templates/courses/chart_tags_progress.html
```diff
@@ -0,0 +1,157 @@
+<script>
```
tylerecouture, line 1: Can you update the styling to match the other chart/histogram?  Use one gray for quests and the other gray for badges, and then maybe no transparency, so the divisions between different quests are visible and not confused with the gridlines.

tylerecouture: We can remove the Total XP bar, that's redundant, right?  Already have that info from the other two stacked together?

tylerecouture (changes requested):
Nicely done sir.  Some changes requested. Thank you.

tylerecouture (changes requested):
One more thing I forgot to add.  Can you create a basic page where this chart can ALSO live on its own? And then add a little chart icon here, and link it to the chart's own page:

image

src/courses/templates/courses/chart_tags_progress.html
9921622: done
done

src/tags/models.py
9921622: added some tests

9921622:
image
image
image
image

src/profile_manager/templates/profile_manager/tag_chart.html
```diff
@@ -0,0 +1,25 @@
...
+
+        <canvas id="TagChart" class="xp-chart">
+            <!-- content will be set by js in chart_tags_progress.html -->
+        </canvas>
+        </div>
+    </div>
+{% endblock %}
+
+{% block js %}
+<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.5.0/Chart.min.js"></script>
```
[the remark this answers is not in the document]
tylerecouture: @9921622  Can you add the integrity hash?  It doesn't look like cdnjs provides them, but you can calculate yourself here: https://www.srihash.org/

src/courses/templates/courses/chart_tags_progress.html
tylerecouture: "then maybe no transparency, so the divisions between different quests are visible and not confused with the gridlines.'  Can still see gridlines through the bars

tylerecouture (changes requested):
I don't understand what I'm looking at here:
You created a tag called Nickel and gave it to the dime tag? or the nickel badge is called dime tag or quest or something... this does not inspire confidence....
image

tylerecouture:
Not sure this title is informative enough, that Tag Chart could be for anyone?
"Tag Chart for some_user" <-- link back to profile
image

src/courses/templates/courses/chart_tags_progress.html
9921622: oops fixed

src/profile_manager/templates/profile_manager/tag_chart.html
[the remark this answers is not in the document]
9921622: done

9921622:
```
You created a tag called Nickel and gave it to the dime tag?
```
Yes, changed Nickel tag name in screenshots for clarity
.
in mark calculations
image
image
in tag charts own view
image

tylerecouture (approved):
No transparency on the bars looks way better. Thanks!

PR #1175, opened 2022-08-18
Feat: Tag details should show XP for quests submission and badges; Closes #1142

### LINKED ISSUE #1142: Tag details should show XP for quests submission and badges
image

Beside each object, sow XP earned in parenthesis

Like they show here:
image

9921622:
staff detail
image
student detail
image

### DIFF
--- a/src/tags/templates/tags/snippets/related-xp-item.html
+++ b/src/tags/templates/tags/snippets/related-xp-item.html
@@ -6,9 +6,22 @@
             <p></p>
             <ul>
                 {% for quest in quests %}
-                    <li><a href="{{ quest.get_absolute_url }}">{{ quest }}</a></li>
+                    {% if view != 'student' %}
+
+                        <li><a href="{{ quest.get_absolute_url }}">{{ quest }} ({{ quest.xp }} XP)</a></li>
+
+                    {% else %}
+                    {% with quest=quest.0 xp=quest.1 %}
+
+                        <li><a href="{{ quest.get_absolute_url }}">{{ quest }} ({{ xp }} XP)</a></li>
+
+                    {% endwith %}
+                    {% endif %}
+
                 {% empty %}
+
                     <li>None.</li>
+
                 {% endfor %}
             </ul>
         </div>
@@ -20,9 +33,22 @@
             <p></p>
             <ul>
                 {% for badge in badges %}
-                    <li><a href="{{ badge.get_absolute_url }}">{{ badge }}</a></li>
+                    {% if view != 'student' %}
+
+                        <li><a href="{{ badge.get_absolute_url }}">{{ badge }} ({{ badge.xp }} XP)</a></li>
+
+                    {% else %}
+                    {% with badge=badge.0 xp=badge.1 %}
+
+                        <li><a href="{{ badge.get_absolute_url }}">{{ badge }} ({{ xp }} XP)</a></li>
+                    
+                    {% endwith %}
+                    {% endif %}
+
                 {% empty %}
+
                     <li>None.</li>
+
                 {% endfor %}
             </ul>
         </div>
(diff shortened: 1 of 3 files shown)


src/tags/templates/tags/snippets/related-xp-item.html
```diff
@@ -20,7 +20,7 @@
             <p></p>
             <ul>
                 {% for badge in badges %}
-                    <li><a href="{{ badge.get_absolute_url }}">{{ badge }}</a></li>
+                    <li><a href="{{ badge.get_absolute_url }}">{{ badge }} ({{ badge.xp }} XP)</a></li>
```
tylerecouture, line 23: Similar to above, are these badges or assertions? Does this take into account multiple awardings of the same badge?

I guess same questions then for the Profile view of XP by tags

src/tags/templates/tags/snippets/related-xp-item.html
```diff
@@ -6,7 +6,7 @@
             <p></p>
             <ul>
                 {% for quest in quests %}
-                    <li><a href="{{ quest.get_absolute_url }}">{{ quest }}</a></li>
+                    <li><a href="{{ quest.get_absolute_url }}">{{ quest }} ({{ quest.xp }} XP)</a></li>
```
tylerecouture, line 9: I don't think this is correct.   Are these quests or submissions?  Seems like they should be submissions. Otherwise this won't take into account custom xp for a submission, will it?  Also, how does this account for multiple completions of a repeatable quest?

src/tags/views.py
```diff
@@ -58,7 +58,7 @@ def get_context_data(self, **kwargs):
         return super().get_context_data(**kwargs)
 
 
-class TagDetailStudent(TagDetailMixin, NonPublicOnlyViewMixin, LoginRequiredMixin, DetailView):
+class TagDetailStudent(TagDetailMixin):
```
tylerecouture, line 61: Mixin is something to "mix in" with a base class to give it extra properties and methods, but here it is being used as a base class it seems?  i.e to add some stuff or set some values for the base class of DetailView.

Might just mean that TagDetail Mixin ISNT a mixin (most of this is just naming conventions, but confusion when not followed, and coders intention is hard to read when conventions are followed=)

src/tags/views.py
```diff
@@ -77,23 +77,27 @@ def get(self, *args, **kwargs):
...
+        return Badge.objects.filter(pk__in=badge_pks)
 
     def get_context_data(self, **kwargs):
         kwargs['user_obj'] = self.user
         return super().get_context_data(**kwargs) 
 
 
 @method_decorator(staff_member_required, name='dispatch')
-class TagDetailStaff(TagDetailMixin, NonPublicOnlyViewMixin, LoginRequiredMixin, DetailView):
+class TagDetailStaff(TagDetailMixin):
```
tylerecouture, line 100: Same comment as above.

src/tags/views.py
```diff
@@ -58,7 +58,7 @@ def get_context_data(self, **kwargs):
         return super().get_context_data(**kwargs)
```
tylerecouture, line 58: `TagDetailMixin`  isn't a Mixin, it's a new base class that extends the Detail View. Should just be called TagDetailView

tylerecouture (changes requested):

src/tags/templates/tags/snippets/related-xp-item.html
9921622: " are these badges or assertions?"
these are badges

"Does this take into account multiple awarding's of the same badge?"
no since it was never a feature to be implemented in (#1125)

src/tags/views.py
9921622: removed mixin from class name

9921622:
added requested changes

PR #1176, opened 2022-08-19
Feat: Refactored Mark Distribution graph, tests, no more negative values; Closes #246

### LINKED ISSUE #246: Negative values in the Mark Distribution graph

9921622:
Removed jchart
Mark distribution graph now uses chart.js

there should no longer be negative values + added a test for that

I made it so the user bar is half the size of the biggest bar for visibility reasons

chart with random student mark values
image

### DIFF
--- a/src/courses/templates/courses/chart_mark_distribution.html
+++ b/src/courses/templates/courses/chart_mark_distribution.html
@@ -1,105 +1,126 @@
-<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.5.0/Chart.min.js"></script>
-<!-- https://github.com/chartjs/Chart.Zoom.js/blob/master/samples/zoom.html -->
-<!-- <script type="text/javascript" src="https://raw.githubusercontent.com/chartjs/Chart.Zoom.js/master/Chart.Zoom.min.js"></script></script> -->
-
 <script type="text/javascript">
+  {% if request.user.pk == user.pk %}
+  var userLabel = "Your mark range";
+  {% else %}
+  var userLabel = "{{user.username}}'s mark range";
+  {% endif %}
 
-  /*jshint esversion: 6 */
-
-  var ctx2 = document.getElementById("chart-mark-distribution");
-
-  data = {
-        labels: ["Red", "Blue", "Yellow", "Green", "Purple", "Orange"],
-        datasets: [{
-            label: '# of Votes',
-            data: [12, 19, 3, 5, 2, 3],
-            backgroundColor: [
-                'rgba(255, 99, 132, 0.2)',
-                'rgba(54, 162, 235, 0.2)',
-                'rgba(255, 206, 86, 0.2)',
-                'rgba(75, 192, 192, 0.2)',
-                'rgba(153, 102, 255, 0.2)',
-                'rgba(255, 159, 64, 0.2)'
-            ],
-            borderColor: [
-                'rgba(255,99,132,1)',
-                'rgba(54, 162, 235, 1)',
-                'rgba(255, 206, 86, 1)',
-                'rgba(75, 192, 192, 1)',
-                'rgba(153, 102, 255, 1)',
-                'rgba(255, 159, 64, 1)'
-            ],
-            borderWidth: 1
-        }]
+  function InstantiateGraph(binLabels, studentData, userIndex) {
+    var COLORS = {
+      students: {bg: 'rgba(128, 128, 128, 0.3)', border: 'rgba(0, 0, 0, 0.2)'},
+      you: {
+        {% if request.user.profile.dark_theme %}
+          // dark theme
+          bg: 'rgba(150, 150, 150, 0.5)', 
+          border: 'rgba(0, 0, 0, 0.1)'
+        {% else %}
+          // light theme
+          bg: 'rgba(25, 25, 25, 0.5)', 
+          border: 'rgba(0, 0, 0, 0.1)'
+        {% endif %}
+      },
     }
 
-  options = {
-        scales: {
-            yAxes: [{
-                ticks: {
-                    beginAtZero:true
-                }
-            }]
+    // set bar colors to be student colors if userIndex is == to barIndex
+    var studentBGs = Array.from({ length: studentData.length }, _ => COLORS.students.bg);
+    var studentBorders = Array.from({ length: studentData.length }, _ => COLORS.students.border);
+    studentBGs[userIndex] = COLORS.you.bg;
+    studentBorders[userIndex] = COLORS.you.border;
+
+    // generate chart
+    var MCtx = document.getElementById("chart-mark-distribution").getContext('2d');
+    var MarkChart = new Chart(MCtx, {
+        type: 'bar',
+        data: {
+          labels: binLabels,
+          datasets: [
+            // Student distribution
+            {
+              label: '# of students at each mark range',
+              data: studentData,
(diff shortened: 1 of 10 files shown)


src/courses/templates/courses/mark_calculations.html
```diff
@@ -176,6 +185,7 @@
           )
       });
   </script>
+  <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.5.0/Chart.min.js"></script>
```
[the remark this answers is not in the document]
9921622: already an instance of this loaded in ``src/courses/templates/courses/chart_xp_progress.html``
i just commented it out and moved it

src/courses/templates/courses/chart_mark_distribution.html
```diff
@@ -1,105 +1,84 @@
-<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.5.0/Chart.min.js"></script>
```
tylerecouture, line 1: This file is pure js, so lets make it a .js static file within the courses app and include that within script tags in the html template whereever it appears.

src/courses/templates/courses/chart_xp_progress.html
```diff
@@ -1,4 +1,4 @@
-<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.5.0/Chart.min.js"></script>
+<!-- <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.5.0/Chart.min.js"></script> -->
```
tylerecouture, line 1: Lets just get rid of all these commented out js files.

tylerecouture (changes requested):
Thanks for the suggestion to remove jchart, that cleans stuff up a lot.

Can you replicate the existing charts in style however?  They should look identical to how they were before, less the bugs =D, unless you have a specific style change you want to improve, but can discuss that on slack.

9921622:
graph where student is the only one at bin:
image
dark mode version + viewing from different account
image
graph with multiple students in the same bin
image

src/courses/templates/courses/chart_mark_distribution.html
```diff
@@ -1,105 +1,106 @@
...
```
tylerecouture, line 1: This is still all js in an html file. Didn't resolve this?

src/courses/templates/courses/chart_xp_progress.html
```diff
@@ -1,7 +1,3 @@
-<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/Chart.js/2.5.0/Chart.min.js"></script>
-<!-- https://github.com/chartjs/Chart.Zoom.js/blob/master/samples/zoom.html -->
-<!-- <script type="text/javascript" src="https://raw.githubusercontent.com/chartjs/Chart.Zoom.js/master/Chart.Zoom.min.js"></script></script> -->
-
 <script type="text/javascript">
```
tylerecouture, line 1: This should be a .js file too

tylerecouture (changes requested):
- [ ] fully js html files should be moved to .js files.
- [ ] preference is to use cdns over local versions of js files.  Would prefer removal of chart.js from our repo and rely on cdn.

tylerecouture:
This shouldn't be "other" students any more.  Please remove "# of students at each mark range" for legend label.  And for the dark band put "Your mark range"
image

src/courses/templates/courses/chart_mark_distribution.html
9921622: static doesnt work with django template tags, so not possible
image

src/courses/templates/courses/chart_mark_distribution.html
9921622: thought you wanted chart,js in static.
Still cant do it since template tags doesnt seem to work with static files

src/courses/templates/courses/chart_xp_progress.html
9921622: cant do it since template tags doesnt seem to work with static files

src/courses/templates/courses/chart_mark_distribution.html
9921622: changed chart.js back to cdn

9921622:
- [ ] fully js html files should be moved to .js files.
 static files dont work with context + template tags
- [x] preference is to use cdns over local versions of js files. Would prefer removal of chart.js from our repo and rely on cdn.

 ``Please remove "# of students at each mark range" `` :  im gonna assume replace to this instead of remove. Since removing will it will end up with dataset 0 having no label
.
viewing from different account + user being the only student at bin
image
viewing from same account + multiple students in same bin
image

9921622:
failing the same tests as #1174. Like #1174 i didnt touch anything that these tests test
image
image

src/courses/templates/courses/chart_mark_distribution.html
tylerecouture: oh ya!

src/courses/templates/courses/chart_mark_distribution.html
tylerecouture: Sorry, miscommunication.

src/courses/templates/courses/chart_mark_distribution.html
```diff
@@ -1,105 +1,105 @@
...
-<!-- https://github.com/chartjs/Chart.Zoom.js/blob/master/samples/zoom.html -->
-<!-- <script type="text/javascript" src="https://raw.githubusercontent.com/chartjs/Chart.Zoom.js/master/Chart.Zoom.min.js"></script></script> -->
-
 <script type="text/javascript">
+  //{% if request.user.pk == user.pk %}
```
tylerecouture, line 2: Why the comments?  It should work with or without, since the template is processed before the JS, just curious.

tylerecouture (changes requested):
Nice work.  A couple final tweaks requested:
1. The dark theme sucks.  I didn't want to bother having two different color schemes for the charts, so tried to create something where we could just change the background and it would "change theme" but not sure it works... can you change the darktheme chart/canvas background to match the light theme and see if it looks ok?  It's so hard to see that dark bar on the dark background.... or do you think it looks fine?

2. Can you offset the x-axis labels so that the bars appear between two numbers?  That makes it more clear that a bar is 0-10%, 10-20%, etc.  See the original screenshot on the issue for what it should look like (though hard to see with the crappy custom background of the person who posted the issue)

3. I'll investigate the test problem

src/courses/templates/courses/chart_mark_distribution.html
9921622: vs code gives me annoying red warning signs when i use django template tags, commenting it out silences it. I can remove it once i commit

9921622: now using django syntax extension

9921622:
1. im not sure how to override the dark theme mode, and changing color of data bar seems easier, so i did that.
2. done

image
image

tylerecouture (changes requested):
Woot! Looking good!  Are you able to change the hover label from "50%" to... "50-60%"?  If it's not obvious and easy, don't worry about it. I don't remember if we had that before.

9921622:
image
image
image

tylerecouture (approved):

PR #1179, opened 2022-08-22
temporary quest sort order fix; closes #1162

### LINKED ISSUE #1162: Quests no longer sorted alphabetically
What happened? I assume a recent PR must have messed with this.

- [ ] Fix so still displayed alpha order
- [ ] Add a test for this to ensure doesn't break again in future =)

Somehow the expected order, indicated here (on XPItem model, which is inherited by Quest model), is getting messed up:

```
    class Meta:
        abstract = True
        ordering = ["-sort_order", "-time_expired", "-date_expired", "name"]
```

image

Dylan-RG-Patterson:
this is a temporary fix for [issue #1162 ](https://github.com/bytedeck/bytedeck/issues/1162), along with a test that should be future-proofed for any other permanent solution to the issue along with a change to the class method used in get_available_without_course().

fixed quest ordering, with "expiry" value switched with a sort order display to demonstrate correct ordering:
image

### DIFF
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -22,7 +22,7 @@ from model_bakery import baker
 
 from hackerspace_online.tests.utils import ViewTestUtilsMixin, generate_form_data
 from notifications.models import Notification
-from quest_manager.models import Category, Quest, QuestSubmission, CommonData
+from quest_manager.models import Category, CommonData, Quest, QuestSubmission, XPItem
 from quest_manager.views import is_staff_or_TA
 from siteconfig.models import SiteConfig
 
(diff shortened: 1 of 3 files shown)


src/quest_manager/tests/test_views.py
```diff
@@ -1481,6 +1481,27 @@ def test_show_hidden(self):
...
+        staff_response = self.client.get(reverse('quests:available'))
+
+        # log in as student and access available quests view
+        self.client.force_login(self.test_student)
+        student_response = self.client.get(reverse('quests:available'))
+
+        # get queryset from view context and order it as intended
+        for response in [staff_response, student_response]:
+            displayed_order = response.context['available_quests']
+            intended_order = displayed_order.order_by("sort_order", "-time_expired", "-date_expired", "name")
```
tylerecouture, line 1500: A code comment explaining where these order_by choices are coming from.

src/quest_manager/views.py
```diff
@@ -387,6 +387,11 @@ def quest_list(request, quest_id=None, template="quest_manager/quests.html"):
             available_quests = Quest.objects.get_available(request.user, remove_hidden)
         else:
             available_quests = Quest.objects.get_available_without_course(request.user)
+    
+    # this logic is in place as a temporary fix to issue #1162 (https://github.com/bytedeck/bytedeck/issues/1162)
+    # once a solution is found for the root cause of this issue, this line should be deleted.
+    # likely stems from PR #1158 (https://github.com/bytedeck/bytedeck/pull/1158)
+    available_quests = available_quests.order_by("sort_order", "-time_expired", "-date_expired", "name")
```
tylerecouture, line 394: Love it when code comments reference more details, thank you.

tylerecouture: Had an idea, what about using the model's ordering attribute here directly? That way we don't have it hard coded in 3 different places (model, here, and test)
I think you can go `available_quests.order_by(*Quest._meta.ordering)` or something like that... might have to treat the lists as args and expand with *args, not sure, just a guess

tylerecouture (changes requested):
One suggestions to DRY out the code a bit =)

src/quest_manager/tests/test_views.py
Dylan-RG-Patterson: switched to reference the XPItem model's meta option value for ordering and added context through code comment

src/quest_manager/views.py
Dylan-RG-Patterson: hard-coding removed and replaced with *XPItem._meta.ordering

tylerecouture (approved):

PR #1183, opened 2022-08-26
Removed extra "S" in quest detail tags display

Dylan-RG-Patterson:
Currently, there exists an extra "S" in the quest detail view. This is a small change that removes it.
image


Dylan-RG-Patterson:
i am so sorry

PR #1184, opened 2022-08-26
Fix: Tags by XP needs to account for quest.max_xp and questsubmission.xp_requested; Closes #1182

### LINKED ISSUE #1182: Tags by XP needs to account for quest.max_xp and questsubmission.xp_requested
Needs to pass these two new tests in `class Tag_total_xp_by_tags_Tests(TagHelper, TenantTestCase):`

```    
   def test_submission_with_xp_requested(self):
        """ 
            Verify that the xp_requested value is being used instead of quest.xp.
            This could happen when `quest.xp_can_be_entered_by_students = True`, which means the
            submissions might have an `xp_requested` value.
        """

        # Sanity check: should be no XP by this tag yet
        self.assertEqual(total_xp_by_tags(self.user, ["TEST"]), 0)

        quest = baker.make(Quest, xp=50, xp_can_be_entered_by_students=True)
        quest.tags.add("TEST")
        baker.make(
            QuestSubmission, 
            quest=quest, 
            xp_requested=100,  # Requesting 100 XP and it's been approved
            user=self.user, 
            is_completed=True, 
            is_approved=True, 
            semester=SiteConfig().get().active_semester,
        )

        # Method should recognize the 100 XP requested and approved, instead of quest.xp = 50:
        self.assertEqual(total_xp_by_tags(self.user, ["TEST"]), 100)

    def test_submission_with_max_xp_surpassed(self):
        """ 
            Verify that no more than quest.max_xp is counted toward a tag's XP when there are multiple submissions of the same quest.
            Create 3 submissions of a 50XP quest, but quest.max_xp=100 so this method should return 100
        """

        # Sanity check: should be no XP by this tag yet
        self.assertEqual(total_xp_by_tags(self.user, ["TEST"]), 0)

        quest = baker.make(Quest, xp=50, max_xp=100)  # Can earn no more than 100 XP for this quest
        quest.tags.add("TEST")
        baker.make(
            QuestSubmission, 
            quest=quest, 
            user=self.user, 
            is_completed=True, 
            is_approved=True, 
            semester=SiteConfig().get().active_semester,
            _quantity=3  # 3 submissions of this quest approved = 150 XP, but max counted should be 100
        )

        # Method should recognize a max of 100XP, instead of 150 (3 * quest.xp):
        self.assertEqual(total_xp_by_tags(self.user, ["TEST"]), 100)
```

9921622:


tylerecouture (approved):
Glad you figured it out!

PR #1187, opened 2022-08-27
Fix: Tag chart doesn't account for multiple submissions/assertions; Closes #1181

### LINKED ISSUE #1181: Tag chart doesn't account for multiple submissions/assertions

9921622:
Added a small fix in ``get_quest_submission_total_xp`` where values() does not squish submissions with different xp requested

Tag chart now accounts for multiple submissions + limits submission xp if assertions pointing to same quest is higher than quest.max_xp

repeat quest with new tag tag should be 18 without max_xp
image

total xp for new tag tag is 15 xp since map_xp > -1
image

submission with highest ordinal is cut to match max xp (reapeat (3) should have xp of 5 but changed to 2)
image

image


tylerecouture (approved):
Have not thoroughly reviewed, But since you're done for the summer I'm going to accept as is and will create new issues if anything comes up that needs fixin.  Thank you!

PR #1257, opened 2022-11-26
Fix #1154, Fix #1153

### LINKED ISSUE #1153: DAL widget: show more than one object per model

### LINKED ISSUE #1154: DAL Widget: "Results can not be loaded"
When first loading a page, the widget often says, "Results can not be loaded"

Access by clicking "Edit Prerequisites" at the top of a Quest or Badge edit view/form

e.g from `quest_manager.views.QuestPrereqsUpdate`:

```
class QuestPrereqsUpdate(ObjectPrereqsFormView):
    model = Quest
```

What's the deal with this?  Is there a cache timeout? or is it loading?  Need to fix this.
Peek 2022-08-10 14-49

bashu:


bashu:
cc @tylerecouture

tylerecouture (changes requested):
Thank you!  I'm going to test it out today.  One request: Can you add docstrings and unit tests to your new classes and methods in utilities.forms.py and utilities.forms.py?  More code comments would be appreciated, so less specialized developers understand why you are doing what you do in the code.

tylerecouture (changes requested):
Some behaviour changes in the formset that need to be fixed:

OLD Behaviour:  New empty form in the formset.  Clicking "Add Anther Prerequisite" button adds a new blank form:

Peek 2022-11-26 14-39


NEW Behaviour from PR.  It seems to cache a value for the fields instead of providing a blank form in the formset, and then clicking new field copies the last instead of creating a blank:

Peek 2022-11-26 14-38

bashu:
to @tylerecouture sure, I will fix that and add some tests

src/prerequisites/tests/test_forms.py
```diff
@@ -9,9 +9,7 @@ class PrereqFormsTest(TenantTestCase):
     def generate_prereq_model_choice(): 
         """ A list of tuples: [(model, model_search_field), ...] for DAL widgets"""
```
tylerecouture, line 10: This docstring is no longer accurate

src/prerequisites/tests/test_forms.py
```diff
@@ -9,9 +9,7 @@ class PrereqFormsTest(TenantTestCase):
...
         models = IsAPrereqMixin.all_registered_model_classes()
-        model_choices = [(model, model.dal_autocomplete_search_fields()) for model in models]
-        # print([(model.__name__, model.dal_autocomplete_search_fields()) for model in models])
-        return model_choices
+        return [model for model in models]
```
tylerecouture, line 12: What is this doing? models (from all_registered_model_classes()) is already a list.  This seems to be a list comprehension to generate an identical copy of itself?  If we can't just return `models` here, please comment why we need this. Thanks!

tylerecouture (commented):
Left a question for you, might have just been an oversight (or my misunderstanding).

tylerecouture:
I left a minor question for you in the review, but a major issue is that the new widget does not work in inline forms created by the "Add Another Prerequisite" button:
Peek 2022-11-28 19-59

bashu:
to @tylerecouture did the old worked with inlines?

tylerecouture:
> to @tylerecouture did the old worked with inlines?

Yes. You can see, just check out the develop branch and you'll see the old behaviour. I assume it's just a small change in the html that is now rendered that is breaking the js selectors in the form template.

tylerecouture:
Here's direct comparison between the develop branch and your PR. 
Peek 2022-11-29 18-02

src/prerequisites/tests/test_forms.py
bashu: I didn't wrote this tests, just fixed them to work with new patches.

bashu:
> Here's direct comparison between the develop branch and your PR.

Ok, now I see it

bashu:
to @tylerecouture offtopic, but what do you use to record gifs?

tylerecouture:
> to @tylerecouture offtopic, but what do you use to record gifs?

Peek (on Ubuntu, it might be Linux only?) https://github.com/phw/peek

bashu:
final bug was fixed, try again

bashu:
I found this https://github.com/onmyway133/GifCapture, will try it next time

tylerecouture:
It looks good now in development!  I'm going to accept PR and test in staging.

tylerecouture (approved):

PR #1265, opened 2022-12-05
Fix #1154, Fix #1153

### LINKED ISSUE #1153: DAL widget: show more than one object per model

### LINKED ISSUE #1154: DAL Widget: "Results can not be loaded"

bashu:


bashu:
to @tylerecouture, now select2 choices are cached for 5 minutes. If you want a better performance, consider django-cacheops for an  queryset auto caching and invalidation. 

My work is done here!

PR #1267, opened 2022-12-11
Do not generate email for inactive students

### LINKED ISSUE #1206: Notification emails being sent to old students
Notification emails that are scheduled to send daily to users who have opted in (via setting in their profile), should only be sent to currently active students. I.e. student registered in a course in the currently active semester.

I believe the user email list is NOT filtering based on active semester?  Need to verify, then fix if not working.

See notifications/tasks.py

Can also checkout out announcements/tasks.py for another app that also send emails (though not scheduled with celery-beat like the notifications are)

yujinyuz:
Closes #1206


tylerecouture (changes requested):

tylerecouture:
@yujinyuz following up here too. Thanks!

tylerecouture (approved):

PR #1272, opened 2022-12-13
Replacing DAL with django-select2

bashu:
don't merge it

### DIFF
--- a/src/prerequisites/forms.py
+++ b/src/prerequisites/forms.py
@@ -4,13 +4,13 @@ from crispy_forms import layout
 from django.core.exceptions import FieldDoesNotExist
 from django.contrib.contenttypes.fields import GenericForeignKey
 
-from django_select2.forms import Select2Widget
 from queryset_sequence import QuerySetSequence
 
 from utilities.forms import FutureModelForm
-from utilities.fields import CachedContentObjectChoiceField
+from utilities.fields import ContentObjectChoiceField
 
 from .models import Prereq
+from .widgets import CustomContentObjectSelect2Widget
 
 
 def popover_labels(model, field_strings):
@@ -47,15 +47,15 @@ def hardcoded_prereq_model_choice():
 class PrereqFormInline(FutureModelForm):
     """This form class is intended to be used in an inline formset"""
 
-    prereq_object = CachedContentObjectChoiceField(
+    prereq_object = ContentObjectChoiceField(
         queryset=QuerySetSequence(*[klass.objects.all() for klass in hardcoded_prereq_model_choice()]),
-        widget=Select2Widget,
+        widget=CustomContentObjectSelect2Widget(),
     )
 
-    or_prereq_object = CachedContentObjectChoiceField(
+    or_prereq_object = ContentObjectChoiceField(
         queryset=QuerySetSequence(*[klass.objects.all() for klass in hardcoded_prereq_model_choice()]),
         required=False,
-        widget=Select2Widget,
+        widget=CustomContentObjectSelect2Widget(),
     )
         
     class Meta:
--- a/src/prerequisites/widgets.py
+++ b/src/prerequisites/widgets.py
@@ -1,0 +1,36 @@
+from functools import reduce
+
+from django.db.models import Q
+
+from queryset_sequence import QuerySetSequence
+
+from utilities.widgets import ContentObjectSelect2Widget
+
+
+class CustomContentObjectSelect2Widget(ContentObjectSelect2Widget):
+
+    def filter_queryset(self, term, queryset=None, **dependent_fields):
+        """Return queryset."""
+        if queryset is None:
+            queryset = self.get_queryset()
+
+        queryset_models = []
+        for qs in queryset.get_querysets():
+            model = qs.model
+            if 'name' in [f.name for f in model._meta.get_fields()]:
+                filter_value = 'name'
+            elif 'title' in [f.name for f in model._meta.get_fields()]:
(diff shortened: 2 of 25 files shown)


bashu:
For example

select2

bashu:
@tylerecouture please test it

tylerecouture:
@bashu Can you explain how it works?

tylerecouture (commented):
This all looks great, as far as purging DAL from the repo.  Only concern is lack of unit tests for all the new widget code.

bashu:
to @tylerecouture pleas check new tests in my latest update

src/prerequisites/forms.py
```diff
@@ -47,15 +47,15 @@ def hardcoded_prereq_model_choice():
 class PrereqFormInline(FutureModelForm):
     """This form class is intended to be used in an inline formset"""
 
-    prereq_object = CachedContentObjectChoiceField(
+    prereq_object = ContentObjectChoiceField(
         queryset=QuerySetSequence(*[klass.objects.all() for klass in hardcoded_prereq_model_choice()]),
-        widget=Select2Widget,
+        widget=CustomContentObjectSelect2Widget(),
```
tylerecouture, line 52: can we make this the default widget for a ContentObjectChoiceField?

src/prerequisites/widgets.py
```diff
@@ -0,0 +1,36 @@
+from functools import reduce
+
+from django.db.models import Q
+
+from queryset_sequence import QuerySetSequence
+
+from utilities.widgets import ContentObjectSelect2Widget
+
+
+class CustomContentObjectSelect2Widget(ContentObjectSelect2Widget):
```
tylerecouture, line 10: Can you add a docstring to this class explaining why a customization of `ContentObjectSelect2Widget` is required?  What exactly is begin added or overridded for this widget? Thanks!

src/utilities/views.py
```diff
@@ -1,47 +1,126 @@
...
+    View that handles requests from heavy model widgets.
 
-    def get_queryset(self):
-        # Don't forget to filter out results depending on the visitor !
-        if not self.request.user.is_authenticated:
-            return self.model.objects.none()
-
-        # if model doesn't have a 'name' then try 'title'
-        search_field = 'name' if hasattr(self.model, 'name') else 'title'
+    The view only supports HTTP's GET method.
```
tylerecouture, line 34: This has the exact same docstring as the AutoResponseView from tags.views.  Can you add more detail specifically what this class is for?

src/tags/views.py
```diff
@@ -15,32 +16,55 @@
...
+User = get_user_model()
 
-#         return qs
 
-User = get_user_model()
+class AutoResponseView(AutoResponseView):
+    """
+    View that handles requests from heavy model widgets.
+
+    The view only supports HTTP's GET method.
```
tylerecouture, line 32: This is not very informative, can you be more specific what this class is used for and why?

tylerecouture (changes requested):
This is functioning excellent.  A few questions/comments inline to clarify documentation, and `prerequisites/widgets.py` is not tested

src/tags/views.py
bashu: fixed, see latest PR from me

src/utilities/views.py
bashu: fixed, see latest PR from me

src/prerequisites/widgets.py
bashu: The problem is that `QuerySetSequenc.filter(field_name=...)` will fail if some models in combined queryset doesn't have the same `field_name`, for example `Quest` and `Badge` models has field `name`, but not `title`. `Course` model has field `title`, but not field `name`.

```python
qs = QuerySetSequenc(Quest.objects.all(), Badge.objects.all(), Course.objects.all())
# this will raise an exception
qs.filter(name__icontains=...)
# this will also fail due to exception
qs.filter(title__icontains=...)
# ... and this will not work
qs.filter(title__icontains=..., name__icontains=...)
```

bashu: I wrote `ContentObjectSelect2Widget` to match the API of `ModelSelect2Widget` class, see https://github.com/codingjoe/django-select2/blob/main/django_select2/forms.py#L539

src/prerequisites/forms.py
bashu: see my comments bellow

src/prerequisites/widgets.py
bashu: As solution we can make it work like this,
```python
ContentObjectSelect2Widget(
  search_fields=[
    {Course: ["title__contains"]},
    "name__icontains",
  ]
)
```
and by changing `get_search_fields` and `filter_queryset` methods.

Let me know what do you think.

tylerecouture: DAL had this same issue, which we solved by forcing the models to implement a method to return the search field name.
```
class IsAPrereqMixin:
...
    @staticmethod
    def dal_autocomplete_search_fields():
        """ 
        For Django Autocomplete Light, should return a field on the model, 
        or something more complicated as described in
        https://django-autocomplete-light.readthedocs.io/en/master/gfk.html#form-example

        override this static method in the class to choose different search fields
        """
        return "name"
```

Models implementng IsAPrereqMixin that did not have a `name` field would override this method and return 'title', for example.

bashu: solved in https://github.com/bytedeck/bytedeck/pull/1276

PR #1276, opened 2022-12-29
feature: per-model `search_fields` for ContentObjectSelect2Widget class

bashu:
### What?

I've changed `ContentObjectSelect2Wiget` class to support per-model `search_fields` option.

### Why?

The problem is that `QuerySetSequence.filter(field_name=...)` will fail if some models in combined queryset doesn't have the same `field_name`, for example `Quest` and `Badge` models has field `name`, but not `title`. `Course` model has field `title`, but not field `name`.

### How?

New `search_fields` format

```python
search_field = {
   'myapp': {'mymodel': ['name__icontains']},
}
```

### Testing?

Yes

None

### Anything Else?

Happy New Year, I wish it will suck less than 2022!

### Review request
@tylerecouture


src/prerequisites/forms.py
```diff
@@ -49,13 +49,43 @@ class PrereqFormInline(FutureModelForm):
 
     prereq_object = ContentObjectChoiceField(
         queryset=QuerySetSequence(*[klass.objects.all() for klass in hardcoded_prereq_model_choice()]),
-        widget=CustomContentObjectSelect2Widget(),
+        widget=ContentObjectSelect2Widget(
```
tylerecouture, line 52: Can we DRY out this code by returning these search fields in a method that generates the dictionary based off of each model's "search field" method (previously called something like `dal_autocomplete_field` which existed in the IsAPrereqMixin class and was overridden by models that didn't have a `name` field.

tylerecouture (changes requested):
As discussed (please let me know if any of the requests below won't work for some reason):
* use a method that returns the search field, e.g. 'name' just like it did before with a method in IsAPrereqMixin that returns name and can be overridden by the models that use title. This makes it much easier to add new PreqModels in the future and (was) nicely covered by tests, you just need to test that all models returned by IsAPrereqMixin.all_registered_model_classes() has a field with the same name as the string returned by the method (previously that method was called dal_autocomplete_search_fields() 
* then, if you need to generate a dictionary to pass to ContentObjectSelect2Widget you can write a method that generates the dict and use it in both places instead of duplicate hardcoded dictionaries. Append '_icontains' to the string, etc. 
* don't use hardcoded_prereq_model_choice() that was a hack to get DAL to work, preferably get the models directly with IsAPrereqMixin.all_registered_content_types() or IsAPrereqMixin.all_registered_model_classes() If you MUST use hardcoded_prereq_model_choices()....fine, just explain
* Changing the above should allow you to remove all the parsing of app and models in your dict and return to the much simpler code you had before? I'm not really sure because it's not documented and has no comments so not sure why it's doing what its doing...

src/prerequisites/forms.py
bashu: see https://github.com/bytedeck/bytedeck/pull/1284

PR #1284, opened 2023-01-06
feature: per-model `search_fields` for ContentObjectSelect2Widget class

bashu:
### What?

I've changed `ContentObjectSelect2Wiget` class to support per-model `search_fields` option.

### Why?

The problem is that `QuerySetSequence.filter(field_name=...)` will fail if some models in combined queryset doesn't have the same `field_name`, for example `Quest` and `Badge` models has field `name`, but not `title`. `Course` model has field `title`, but not field `name`.

### How?

New `search_fields` format

```python
search_field = {
   'myapp': {'mymodel': ['name__icontains']},
}
```

### Testing?

Yes

None

### Anything Else?

Happy New Year, I wish it will suck less than 2022!

### Review request
@tylerecouture


src/djcytoscape/forms.py
```diff
@@ -37,12 +33,7 @@ class Meta:
 
     name = forms.CharField(max_length=50, required=False, help_text="If not provided, the initial quest's name will be used")
     
-    initial_content_object = ContentObjectChoiceField(
-        label='Initial Object',
-        required=True,
-        queryset=QuerySetSequence(*[klass.objects.all() for klass in get_model_options()]),
-        widget=ContentObjectSelect2Widget(search_fields=["name__icontains"]),
-    )
+    initial_content_object = AllowedContentObjectChoiceField(label='Initial Object')
```
tylerecouture, line 36: This will cause errors, especially in a new database, because content_types framework doesn't exist yet, but the field is being initialized when the code is read, and the __init__ above is trying to access ContentType

src/prerequisites/forms.py
```diff
@@ -31,32 +30,20 @@ def popover_labels(model, field_strings):
...
 
 
 class PrereqFormInline(FutureModelForm):
     """This form class is intended to be used in an inline formset"""
 
-    prereq_object = ContentObjectChoiceField(
-        queryset=QuerySetSequence(*[klass.objects.all() for klass in hardcoded_prereq_model_choice()]),
-        widget=CustomContentObjectSelect2Widget(),
-    )
+    prereq_object = AllowedContentObjectChoiceField()
```
tylerecouture, line 44: Similar to above, this will cause errors because content_types framework doesn't exist yet, but the field is being initialized when the code is read, and the __init__ above is trying to access ContentType via `Prereq.all_registered_model_classes()`

src/prerequisites/forms.py
```diff
@@ -31,32 +30,20 @@ def popover_labels(model, field_strings):
...
 
-    or_prereq_object = ContentObjectChoiceField(
-        queryset=QuerySetSequence(*[klass.objects.all() for klass in hardcoded_prereq_model_choice()]),
-        required=False,
-        widget=CustomContentObjectSelect2Widget(),
-    )
+    or_prereq_object = AllowedContentObjectChoiceField(required=False)
```
tylerecouture, line 46: Same here

tylerecouture (changes requested):
Part of the problem is that you are trying to access the content_types framework when the modules/app is loaded, but that doesn't exist yet (until the db is set up and migrations are completed).  In the previous code you may have noticed the models were hardcoded into a list (and then compared for accuracy in a test).  This was to avoid the problem of loading content_types before they exist.

I commented on the code that is causing the problem.  The problem is much more obvious if you delete your postres_data volume and recreate it fresh with:
`docker-compose run web bash -c "./src/manage.py initdb"`

You'll get errors like this: 
 `django.db.utils.ProgrammingError: relation "django_content_type" does not exist
 LINE 1: ..."."app_label", "django_content_type"."model" FROM "django_co...`

So there needs to be a way to either lazy load, or defer initializing the fields until they are actually being used, as opposed to initializing the fields when the modules/apps are loaded.

bashu:
to @tylerecouture should I fix that too?

tylerecouture:
> to @tylerecouture should I fix that too?

Yes please!  I'm sorry it took me so long to review this.  The main thing is, if you start with an empty database the errors because clear/obvious.

src/prerequisites/forms.py
bashu: fixed

src/prerequisites/forms.py
bashu: fixed

bashu:
to @tylerecouture should be fixed

tylerecouture:
Thank you!  This fixes the field initialization problem, and that weird failing test from before is gone.  But now there are 4 different fails. Can you take a look at the 4 failing tests?  Let me know if you need assistance with any of them.

tylerecouture (changes requested):
Thank you! This fixes the field initialization problem, and that weird failing test from before is gone. But now there are 4 different fails. Can you take a look at the 4 failing tests? Let me know if you need assistance with any of them.

bashu:
to @tylerecouture here is output of tests running on localhost:

https://pastebin.com/RArZ50Wk

Summary
```shell
...
Ran 632 tests in 1221.827s

OK
Destroying test database for alias 'default'...
```

bashu:
now, here is output from github actions:

```python
FAIL: test_ScapeGenerateMap__POST (djcytoscape.tests.test_views.ViewTests)
Assert a teacher can generate a map using ScapeGenerateMapView
  File "/app/src/djcytoscape/tests/test_views.py", line 111, in test_ScapeGenerateMap__POST
    self.assertTrue(CytoScape.objects.filter(name='New Name').exists())
AssertionError: False is not true

FAIL: test_ScapeUpdateView__POST (djcytoscape.tests.test_views.ViewTests)
Assert a teacher can update a map using ScapeGenerateMapView
  File "/app/src/djcytoscape/tests/test_views.py", line 134, in test_ScapeUpdateView__POST
    self.assertRedirects(response, reverse('djcytoscape:quest_map', args=[self.map.pk]))
  File "/usr/local/lib/python3.8/site-packages/django/test/testcases.py", line 357, in assertRedirects
    self.assertEqual(
AssertionError: 200 != 302 : Response didn't redirect as expected: Response code was 200 (expected 302)

FAIL: test_post_save_button__defaults (quest_manager.tests.test_views.QuestPrereqsUpdate)
Save button should redirect to the quest detail view, no changes to form data
  File "/app/src/quest_manager/tests/test_views.py", line 1143, in test_post_save_button__defaults
    self.assertRedirects(response, self.parent_quest.get_absolute_url())
  File "/usr/local/lib/python3.8/site-packages/django/test/testcases.py", line 357, in assertRedirects
    self.assertEqual(
AssertionError: 200 != 302 : Response didn't redirect as expected: Response code was 200 (expected 302)

FAIL: test_post_save_button__new_values (quest_manager.tests.test_views.QuestPrereqsUpdate)
New prereq data in form should change the prereqs for the parent quest.
  File "/app/src/quest_manager/tests/test_views.py", line 1207, in test_post_save_button__new_values
    self.assertRedirects(response, self.parent_quest.get_absolute_url())
  File "/usr/local/lib/python3.8/site-packages/django/test/testcases.py", line 357, in assertRedirects
    self.assertEqual(
AssertionError: 200 != 302 : Response didn't redirect as expected: Response code was 200 (expected 302)
```

tylerecouture:
Perhaps it just requires running our `initdb` management command in the test SetUp method, or even just `migrate`

bashu:
to @tylerecouture tell me how to run tests locally, so they fail for sure?

tylerecouture:
> to @tylerecouture tell me how to run tests locally, so they fail for sure?

Here's how I can do it:
1.  `docker-compose down` 
2. remove the postgres_volume (i.e. the django database) which I  did from VS Code's Docker panel
3. run the tests.  I only ran the djcytoscape tests for the demo cus I know one fails there

Peek 2023-03-24 22-47

Time skip...

Peek 2023-03-24 22-48

bashu:
to @tylerecouture I found and fixed mutual imports in tests

bashu:
is there a way to execute the test suite in 3-4 minutes, rather than in 30-40 minutes, like right now?

tylerecouture:
> is there a way to execute the test suite in 3-4 minutes, rather than in 30-40 minutes, like right now?

You can just run tests in a specific app instead of the entire project by indicating "src/announcements" or "src/djcytoscape" instead of "src" (e.g see my screen recording above)

Parallel and failfast options will cut time also.

tylerecouture:
Which test fails and in which app?

bashu:
> Which test fails and in which app?

these two : https://github.com/bytedeck/bytedeck/pull/1284#issuecomment-1490266825

tylerecouture:
> > Which test fails and in which app?
> 
> these two : https://github.com/bytedeck/bytedeck/pull/1284#issuecomment-1490266825

Thank you. I will investigate when I get time.

src/djcytoscape/forms.py
```diff
@@ -1,28 +1,21 @@
 from django import forms
-
-from queryset_sequence import QuerySetSequence
+from django.contrib.contenttypes.models import ContentType
 
 from utilities.forms import FutureModelForm
-from utilities.fields import ContentObjectChoiceField
-from utilities.widgets import ContentObjectSelect2Widget
+from utilities.fields import AllowedContentObjectChoiceField as ContentObjectChoiceField
```
tylerecouture, line 5: Please don't `import as` unless it's a standard convenion, it makes the code hard to maintain and read.  Explicit is better than Implicit!  Now there are two AllowedContentObjectChoice field classes.  Give the class below a new name that fits why you are extending it.

src/prerequisites/forms.py
```diff
@@ -4,13 +4,11 @@
 from django.core.exceptions import FieldDoesNotExist
 from django.contrib.contenttypes.fields import GenericForeignKey
 
-from queryset_sequence import QuerySetSequence
 
 from utilities.forms import FutureModelForm
-from utilities.fields import ContentObjectChoiceField
+from utilities.fields import AllowedContentObjectChoiceField as ContentObjectChoiceField
```
tylerecouture, line 9: See comment above, don't `import as`

src/prerequisites/forms.py
```diff
@@ -31,32 +29,18 @@ def popover_labels(model, field_strings):
...
 
 
-def hardcoded_prereq_model_choice():
-    """ Can't always dynamically load this list due to accessing contenttypes too early
-    So instead provide a hard coded list which is checked during testing to ensure it matches
-    what the dynamically loaded list would have produced """
-    from courses.models import Block, Course, Grade, Rank
-    from quest_manager.models import Category, Quest
-    from badges.models import Badge
+class AllowedContentObjectChoiceField(ContentObjectChoiceField):
```
tylerecouture, line 32: Call this `PrereqContentObjectChoiceField`, since that's what it is.

src/djcytoscape/forms.py
```diff
@@ -1,28 +1,21 @@
...
-        formatted for utilities.fields.ContentObjectModelField use
-        from prerequisites.forms:
-            Can't always dynamically load this list due to accessing contenttypes too early
-            So instead provide a hard coded list which is checked during testing to ensure it matches
-            what the dynamically loaded list would have produced
-    """
-    from quest_manager.models import Quest
-    from badges.models import Badge
-    from courses.models import Rank
+class AllowedContentObjectChoiceField(ContentObjectChoiceField):
```
tylerecouture, line 10: call this `CytoscapeContentObjectChoiceFiled` since that's what it is, specific to Cytoscape model.

src/djcytoscape/tests/test_fields.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib.contenttypes.models import ContentType
+
+from django_tenants.test.cases import TenantTestCase
+
+from djcytoscape.models import CytoScape
+
+
+class AllowedContentObjectChoiceFieldTest(TenantTestCase):
+
+    def test_queryset(self):
```
tylerecouture, line 10: docstring needed

src/prerequisites/forms.py
```diff
@@ -31,32 +29,18 @@ def popover_labels(model, field_strings):
...
 
-    return [
-        Category, Quest, Block, Course, Grade, Rank, Prereq, Badge,
-    ]
+    def get_allowed_model_classes(self):
```
tylerecouture, line 34: Needs docstring

tylerecouture (changes requested):
I still have to take a look at the tests.  But can you update the naming of `content_object`  it is not clear what that is referring to.  I think a better name would just be `GFK` for Generic Foreign Key.  So `AllowedContentObjectChoiceField` would become AllowedGFKChoiceField.  Just to a replace all for ContentObject and content_object etc.

src/djcytoscape/tests/test_fields.py
bashu: why do yoo need docstring in tests?

tylerecouture: It's helpful for other developers, it prints when a test fails, providing useful info and saving time, also see:
https://hynek.me/articles/document-your-tests/
https://jml.io/test-docstrings/

bashu:
to @tylerecouture check https://github.com/bytedeck/bytedeck/pull/1284/commits/210ca3a6f93739af3e260b9450118ca07e3d7e20

src/djcytoscape/tests/test_fields.py
tylerecouture: However, I realize this convention is not documented in the current contributing.md, so if you want to skip it, you can.

tylerecouture:
> to @tylerecouture check [210ca3a](https://github.com/bytedeck/bytedeck/commit/210ca3a6f93739af3e260b9450118ca07e3d7e20)

Still lots of `ContentObject` and `content_object`.  Can you do find/replace all?

tylerecouture:
Can you uncomment those two tests and push, and we can see if recent unrelated updates to the repo might have fixed things... Fingers crossed!

bashu:
> Still lots of `ContentObject` and `content_object`. Can you do find/replace all?

see https://github.com/bytedeck/bytedeck/pull/1284/commits/fb29e57932c1834f85fc1e0e491bb59d07aef174

bashu:
> Can you uncomment those two tests and push, and we can see if recent unrelated updates to the repo might have fixed things... Fingers crossed!

see latest commit

bashu:
> Can you uncomment those two tests and push, and we can see if recent unrelated updates to the repo might have fixed things... Fingers crossed!

one test failed, and what it has to do with this PR?

tylerecouture:
> > Can you uncomment those two tests and push, and we can see if recent unrelated updates to the repo might have fixed things... Fingers crossed!
> 
> one test failed, and what it has to do with this PR?

It is definitely weird.  I've asked @yujinyuz to take a look a this.

yujinyuz:
@bashu @tylerecouture 

I've commented on another ticket but will comment it here as well..

This PR https://github.com/bytedeck/bytedeck/pull/1326, I added a comment that it is failing during `14:20 - 14:40` UTC+8 (my time) probably it's somewhere between `14:00 and 15:00` but I've only managed to confirm it with `14:20 - 14:40`.

Try to re-run the tests again after this time and see that it would now succeed

bashu:
to @tylerecouture  and @yujinyuz rebased, but still failing (only one failure):
```python
  File "/app/src/quest_manager/tests/test_models.py", line 161, in test_active
AssertionError: True != False
```

yujinyuz:
@bashu try after an 30 mins or an hour. I think this test case will pass

bashu:
to @yujinyuz lol, it passed

tylerecouture (changes requested):
Find/Replace!!!

image

bashu:
to @tylerecouture

bashu:
not happening, these will retain its names, cuz GFKSelect2Widget and GFKChoiceField are build upon various mixins and subclasses.

Screenshot 2023-04-09 at 11 03 21 PM

PR #1292, opened 2023-02-18
Automatically fix ordinals on error

yujinyuz:
### What?
Temporary hack to fix incorrect ordinals
### Review request
@tylerecouture

### DIFF
--- a/src/quest_manager/models.py
+++ b/src/quest_manager/models.py
@@ -1,6 +1,5 @@
 import uuid
 import json
-import logging
 
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
@@ -12,7 +11,6 @@ from django.db.models.functions import Greatest
 # from django.templatetags.static import static
 from django.urls import reverse
 from django.utils import timezone, datetime_safe
-from django.forms import model_to_dict
 
 from siteconfig.models import SiteConfig
 
@@ -24,8 +22,6 @@ from tags.models import TagsModelMixin
 
 # from django.contrib.contenttypes.models import ContentType
 # from django.contrib.contenttypes import generic
-
-logger = logging.getLogger(__name__)
 
 
 class Category(IsAPrereqMixin, models.Model):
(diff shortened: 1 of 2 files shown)


tylerecouture (changes requested):
See inline comments.

src/quest_manager/models.py
```diff
@@ -954,13 +958,38 @@ def is_returned(self):
     def get_comments(self):
         return Comment.objects.all_with_target_object(self)
 
+    def _fix_ordinal(self):
+        # NOTE: There is a rare bug that we are unable to reproduce as of the moment where a QuestSubmission has the same ordinal.
+        # This is just a temporary hack where it will automatically fix the incorrect ordinal
+        broken_ordinal_start = self.ordinal - 1
+        submissions = QuestSubmission.objects.filter(quest=self.quest, user=self.user, ordinal__gte=self.ordinal - 1).order_by('time_completed')
+        logger.debug(
```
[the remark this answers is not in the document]
yujinyuz: If my memory is correct, this should be logged in one of the log files.

It seems that this will never be called since `LOGGING` is not configured here.. though this could provide useful information so that we could see which quest submissions were having incorrect ordinals

src/quest_manager/tests/test_models.py
```diff
@@ -381,6 +382,51 @@ def test_submission_get_previous(self):
...
+
+        fourth_sub = QuestSubmission.objects.create_submission(user=self.student, quest=repeat_quest)
+        self.assertEqual(fourth_sub.ordinal, 4)
+        fourth_sub.mark_completed()
+
+        # Make the 3rd submission's ordinal same with the second
+        old_third_submission_ordinal = third_sub.ordinal
+        third_sub.ordinal = second_sub.ordinal
+        third_sub.save()
+
```
tylerecouture, line 412: At this point, please check explicitly that  "MultipleObjectsReturned" exception is thrown when you try to `.get(ordinal=second_sub.ordinal)`

I see you are sort of checking that by mocking and verifying that _fix_ordinal is called, but the method could change in the future, and we want to verify why it's called, ie because MultipleObjectsReturned!

tylerecouture (changes requested):

src/quest_manager/tests/test_models.py
yujinyuz: ooh. that's what you meant. thanks for clarifying

PR #1300, opened 2023-03-11
Oauth setup

yujinyuz:

### What?

* Added a new sign in option via Google.
* This will automatically allow signing up for an account using just a Google account though students would still need to provide the access code given by their teachers.
* For existing accounts, they can still use the both signing in via their username OR via google sign in provided that they added their email address in their profile and is using the same email address with their google account.
* In case a student has an existing account but has not set up an email address in their profile, and tries to login via google, it will be assumed that the student logging in is a new student and a new account will be created if they completed the sign up process.
* Added a new field `SiteConfig.enable_google_signin` for enabling or disabling google sign in

### Why?

For enabling google sign in

### How?

* This is done by leveraging the `django-allauth` package, using the google provider `allauth.socialaccount.providers.google`, by adding it into the `INSTALLED_APPS`.
* django-allauth also makes use of the `django.contrib.sites` since adding the consumer keys, tokens, etc. required by a social provider are added in the database. You can read more details here: https://django-allauth.readthedocs.io/en/latest/overview.html#architecture-design
* Some settings were added:
    * Set `ACCOUNT_EMAIL_VERIFICATION = "optional"` so an email is sent for verifying emails after sign up
    * Set `ACCOUNT_UNIQUE_EMAIL = True` to prevent duplicate emails since django-allauth also has their own way of handling emails
    * Added `SOCIALACCOUNT_AUTO_SIGNUP = False` to prevent bypassing the sign up page where a student has to enter an access code before being able to enter the site. When this is set to `True`, they will be automatically redirected to the homepage of a normal student without having to enter the access code.
    * Added `SOCIALACCOUNT_EMAIL_REQUIRED = True` so that signing up via email address is required when using google sign in.
    * Added `SOCIALACCOUNT_ADAPTER = "hackerspace_online.adapter.CustomSocialAccountAdapter"` for automatically merging accounts

### Testing?

Not sure yet on how to test since this is just leveraging django-allauth package

Kapture 2023-03-11 at 20 26 06

https://user-images.githubusercontent.com/10972027/224485219-39d4f930-9fbd-4d2b-8b63-a47fc746a778.mp4

### Anything Else?

When testing this PR, you need to obtain google credentials here: https://developers.google.com/workspace/guides/create-credentials#oauth-client-id

### Review request
@tylerecouture


tylerecouture:
Thanks @yujinyuz, this looks great!

Could you provide some more info in the What and How sections before I dig into the code?  

- What settings were added or changed and what do we need to configure for different environments (dev vs production)
- Are there any new dependencies?
- What happens if a user registers via OAuth, but tries to sign in with their account name?
- What happens if a local registration does NOT click the link to verify the email address?
- Are email addresses still verified if a uer changes it later by editing their profile?
- How does the OAuth registration handle it when there is already another account that uses the gmail address (and can this be tested?)
- At there very beginning of the video, when you clicked "sign up" from the sign in page, I don't see the Google option that I can see from your GIF, why is that?
- Can you take a look at the Password reset email template (it's simple and lame...) and use that?
- Since we haven't been using all-auth for much, can you ensure we are using the latest version of it and that it's compatible?

Tests
- Can we test that we don't get 404 or server error when hitting the google sign up and google log in?
- Can we test what happens when we get conflicting emails (local account already using a gmail, then they try to OAuth register with that gmail? As one example.
- Since the OAuth and local forms are similar, are the fields tested to prevent them from diverging in the future (e.g. someone edits one form but doesn't realize there are two and both need to be edited.)

yujinyuz:
@tylerecouture 

> Could you provide some more info in the What and How sections before I dig into the code?

Updated the description

> What settings were added or changed and what do we need to configure for different environments (dev vs production)

Added in the description


> Are there any new dependencies?

Just the allauth package


> What happens if a user registers via OAuth, but tries to sign in with their account name?

If by account name, you mean username, they won't be able to login since by default they don't have a usable password. Though they can login via google and change their password so they can now login via password and via google


> What happens if a local registration does NOT click the link to verify the email address?

Nothing as of the moment. They just won't have a verified email.

> Are email addresses still verified if a uer changes it later by editing their profile?

Emails need to be re-verified when it is changed. However, due to how django-allauth handles this where they have a separate model for storing emails, I think we should create a separate task for this one.

What I mean is that if I change my email via `User.email`, the `EmailAddress` model won't be changed. We have to update both in order for it to take effect.

I think this issue quite similar: https://github.com/pennersr/django-allauth/issues/2694


> How does the OAuth registration handle it when there is already another account that uses the gmail address (and can this be tested?)

via the CustomSocialAccountAdapter class. I think it can be tested and might involve mocking..

> At there very beginning of the video, when you clicked "sign up" from the sign in page, I don't see the Google option that I can see from your GIF, why is that?

I'm not quite sure what you mean. But sign in and sign up is the same buttom for google.


> Can you take a look at the Password reset email template (it's simple and lame...) and use that?

Use it for what?

> Since we haven't been using all-auth for much, can you ensure we are using the latest version of it and that it's compatible?

Sure.

---

Tests

> Can we test that we don't get 404 or server error when hitting the google sign up and google log in?

What do you mean by this? Do you mean by actually sending an http request to the google server?

> Can we test what happens when we get conflicting emails (local account already using a gmail, then they try to OAuth register with that gmail? As one example.

If a local account is using gmail, and they use oauth register with that gmail, it will merge and they can now use google sign in and normal sign in

> Since the OAuth and local forms are similar, are the fields tested to prevent them from diverging in the future (e.g. someone edits one form but doesn't realize there are two and both need to be edited.)

Haven't tested yet. Do you think that maybe just duplicating the fields would be better?

yujinyuz:
Looks like the failing test is caused by the upgrade that was introduced django-allauth 0.48.0 release related to `ACCOUNT_RATE_LIMITS` https://django-allauth.readthedocs.io/en/latest/release-notes.html#section-5

README.md
```diff
@@ -169,8 +169,49 @@ Using pgadmin4 we can inspect the postgres database's schemas and tables (helpfu
...
 
-## Contributing
+### Enabling Google Sign In (Optional)
 
-See [CONTRIBUTING.md](https://github.com/bytedeck/bytedeck/blob/develop/CONTRIBUTING.md) if you plan to contribute code!
 
+Here are the steps, assuming that you now have a functional tenant:
+
+1. Obtain Google credentials: https://developers.google.com/workspace/guides/create-credentials#oauth-client-id
+2. Make sure that in the Authorized URIs, add `http://hackerspace.localhost.net:8000/accounts/google/login/callback/`. We will explain why we are using `localhost.net` later but for now, just add this.
```
tylerecouture, line 178: Won't this only work for a single tenant?  Do we need to set up Authorized URIs for every tenant?  This implementation needs to work for any bytedeck tenant (subdomain) and should be agnostic to the tenant using google sign in.

README.md
```diff
@@ -169,8 +169,49 @@ Using pgadmin4 we can inspect the postgres database's schemas and tables (helpfu
...
+3. Go to Social Applications: http://hackerspace.localhost:8000/admin/socialaccount/socialapp/
```
tylerecouture, line 179: Every tenant has to do this?

README.md
```diff
@@ -169,8 +169,49 @@ Using pgadmin4 we can inspect the postgres database's schemas and tables (helpfu
...
+4. Click Add Social Application
+5. Fill in `Client Id` and `Secret Key`. And then add the `Available Sites` to `Chosen Sites`
+6. Click Save
+7. Go to your Site Configuration: http://hackerspace.localhost:8000/config/ and click `Enable sign-in via Google`
+8. Done
+
+When you are developing locally, Google won't allow you to add `http://hackerspace.localhost:8000/accounts/google/login/callback/` in the Authorized URIs. So we need a way to bypass this in our local machine by mapping
+our localhost to `localhost.net` so we can access our tenant via `http://hackerspace.localhost.net:8000`.
+
+We need to modify our hosts file aka `/etc/hosts`.
```
tylerecouture, line 189: Where is this?  in docker config somewhere?

README.md
```diff
@@ -169,8 +169,49 @@ Using pgadmin4 we can inspect the postgres database's schemas and tables (helpfu
...
+
+Add the following, preferably at the bottom of the file:
```
tylerecouture, line 191: Please number sequential steps (see the rest of readme for examples)

README.md
tylerecouture: This seems like something that should be automated when a tenant owner turns the setting on in their SiteConfig.  This is not something we can rely on tenant owners (i.e. lay people) to do.

tylerecouture (changes requested):
I haven't looked at much code yet, just reviewing the Notes and readme and have some questions/comments.

In the notes, can you articulate what setting "enable_google_signin=True" does?  Does it just work? or does an admin on a tenant still need to go through all that stuff in the Readme?  That doesn't seem like a realistic ask of users in production.

src/hackerspace_online/settings.py
```diff
@@ -497,6 +500,7 @@
 
 
 # AUTHENTICATION ##################################################
+SESSION_COOKIE_AGE = env("SESSION_COOKIE_AGE", default=int(60 * 60 * 24 * 7 * 8))  # 8 Weeks
```
tylerecouture, line 503: Where does this come from and how is it used?  Please provide info in PR Notes.

src/siteconfig/models.py
```diff
@@ -152,6 +155,12 @@ class SiteConfig(models.Model):
         help_text="Set up at least one Mark Range in admin for this to do anything."
     )
 
+    enable_google_signin = models.BooleanField(
+        verbose_name="Enable sign-in via Google",
+        default=False,
+        help_text="Used to enable or disable google sign-in"
```
tylerecouture, line 161: help_text is a repeat of the verbose name and redundant.  It should give useful info or remove.

README.md
yujinyuz: I think it's possible that we only have one client/secret settings and one social app instance by moving the `allauth.socialaccount` to the `SHARED_APPS` only.

Also, every time a tenant wants to use Google OAuth, they should be manually added to the Authorized Redirect URIs

image


We can't use wildcards as per documentation:

> Users will be redirected to this path after they have authenticated with Google. The path will be appended with the authorization code for access, and must have a protocol. It can’t contain URL fragments, relative paths, or wildcards, and can’t be a public IP address.

README.md
yujinyuz: Gotcha. As per my comment above, the solution here would be to move `allauth.socialaccount` to `SHARED_APPS` so we only need to configure it in the public admin

README.md
yujinyuz: I'll update the docs to clarify. This is only done when developing locally and should be done in the host machine

README.md
tylerecouture: > I think it's possible that we only have one client/secret settings and one social app instance by moving the `allauth.socialaccount` to the `SHARED_APPS` only.

Ok, let's try that.  But if this is the case (individual tenants must be added to the URIs), either:

1. We need to be able to add them programmatically via API somehow, OR
2. The setting needs to be read only on the form  With help-text indicating that if the deck owner wants to enable this setting, they must contact Bytedeck; and then a simple management command to enable it on the tenant (i.e. turning  the SiteConfig field to True and doing whatever other configuration needs to be done on the tenant, if any).  The management command should also include instructions on what needs to be done FIRST (i.e. add "{{tenant}}.bytedeck.com/.../.../  to the list of authorized urls.

README.md
yujinyuz: Seems like we can't just move the `allauth.socialaccount` to `SHARED_APPS` only since each tenant has different users.

Instead, I did a hack that automatically copies the social app configuration from admin to tenant when they enable google sign in.

README.md
yujinyuz: Google documentation doesn't provide an obvious way to add redirect uris programatically..

Also [this](https://stackoverflow.com/questions/73467957/google-console-how-to-add-authorized-redirect-uris-via-api) stack overflow issue mentions that it's not possible.

Can you explain further with the 2nd approach? Do you mean like there is just a management command and then it just prints out the instructions on what to do?

tylerecouture: > Google documentation doesn't provide an obvious way to add redirect uris programatically..
> 
> Also [this](https://stackoverflow.com/questions/73467957/google-console-how-to-add-authorized-redirect-uris-via-api) stack overflow issue mentions that it's not possible.
> 
> Can you explain further with the 2nd approach? Do you mean like there is just a management command and then it just prints out the instructions on what to do?

Let me try what you've done so far and then I will get back to you =)

tylerecouture:
@yujinyuz Can you provide more thorough, step by step documentation about adding Google credentials?

yujinyuz:
Did you follow the instructions here? https://developers.google.com/workspace/guides/create-credentials#oauth-client-id

image

---

Here's what it looks like on my end and I think you're in a different page

image

tylerecouture:
> Did you follow the instructions here? https://developers.google.com/workspace/guides/create-credentials#oauth-client-id
> 

I tried this too late at night and my brain was mush...I totally missed that.

tylerecouture (changes requested):
Thank you for this!  It looks great.  Because it involves authentication and emails, I want to ensure we have all possibilities covered.  I think we need a flow chart to cover all possible sign-up methods and email changing.

Here are a few things I tested that will be problems we need to handle.  

- After logging in with Google, I could change my email address in my profile to anything.  The Google log-in still worked, but which email are notifications going to?
- After logging in, I could change my email address in my profile to another user's Gmail address! (since they changed it to something else)
- Does "remember me" still work with Google sign in? This is important.

Can we merge to two email fields?  Having two seems like it will be a problem.  Or maybe we just need tighter controls on changing the email address within a profile.  I think ALL email addresses should be verified, can django-allauth handle that?

src/hackerspace_online/adapter.py
```diff
@@ -1,8 +1,27 @@
...
 class CustomAccountAdapter(DefaultAccountAdapter):
 
     def clean_username(self, username, shallow=False):
         username = super().clean_username(username, shallow)
         return username.lower()
+
+
+class CustomSocialAccountAdapter(DefaultSocialAccountAdapter):
+
+    def pre_social_login(self, request, sociallogin):
```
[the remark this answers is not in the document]
yujinyuz: > Thank you for this! It looks great. Because it involves authentication and emails, I want to ensure we have all possibilities covered. I think we need a flow chart to cover all possible sign-up methods and email changing.
> 
> Here are a few things I tested that will be problems we need to handle.
> 
> * After logging in with Google, I could change my email address in my profile to anything.  The Google log-in still worked, but which email are notifications going to?
> * After logging in, I could change my email address in my profile to another user's Gmail address! (since they changed it to something else)
> * Does "remember me" still work with Google sign in? This is important.
> 
> Can we merge to two email fields? Having two seems like it will be a problem. Or maybe we just need tighter controls on changing the email address within a profile. I think ALL email addresses should be verified, can django-allauth handle that?




> After logging in with Google, I could change my email address in my profile to anything. The Google log-in still worked, but which email are notifications going to?

This will use the email in the profile. The google account is still associated to that account

> After logging in, I could change my email address in my profile to another user's Gmail address! (since they changed it to something else)

Yeah.. this isn't something we could control. Maybe we could disable changing emails? But we need to think about a workflow for this


> Can we merge to two email fields? Having two seems like it will be a problem. Or maybe we just need tighter controls on changing the email address within a profile. I think ALL email addresses should be verified, can django-allauth handle that?

What do you mean by merging? Also can you give me more details on ALL email addresses should be verified?

src/siteconfig/forms.py
```diff
@@ -15,7 +16,19 @@ class Meta:
...
     def __init__(self, *args, **kwargs):
         is_deck_owner = kwargs.pop('is_deck_owner', False)
+
         super().__init__(*args, **kwargs)
-        
+
         if not is_deck_owner:
             self.fields['deck_owner'].disabled = True
+
+    def clean_enable_google_signin(self):
```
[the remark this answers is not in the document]
yujinyuz: Yup. Can make this read only. I think we can just stick to a simple workflow for now

src/templates/socialaccount/connections.html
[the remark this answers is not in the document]
yujinyuz: Yup. You'd be able to see connected google accounts to your current account but I haven't actually added any links to access this page since I wasn't sure if we were gonna expose this page.

It can be accessed here: http://tenant.localhost.net:8000/accounts/connections/

src/templates/socialaccount/login.html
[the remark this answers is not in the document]
yujinyuz: You can access this in the login page when you click the google sign in button and then you will get redirected to a page which is using this template and if you click proceed, you will be redirected to the actual google sign in page

src/hackerspace_online/adapter.py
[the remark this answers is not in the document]
tylerecouture: > > After logging in, I could change my email address in my profile to another user's Gmail address! (since they changed it to something else)
> 
> Yeah.. this isn't something we could control. Maybe we could disable changing emails? But we need to think about a workflow for this

How does django-allauth handle this?  Currently the email field is the one on django's User model.  But I know allauth does email verification, so how do we use that? 

What are these settings actually doing in our project?

> Set ACCOUNT_EMAIL_VERIFICATION = "optional" so an email is sent for verifying emails after sign up
> Set ACCOUNT_UNIQUE_EMAIL = True to prevent duplicate emails since django-allauth also has their own way of handling emails


"an email is sent for verifying emails after sign up" <-- I haven't experienced this in any of the sign-up workflows (local or Google), so how do we use it?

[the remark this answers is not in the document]
yujinyuz: > > > After logging in, I could change my email address in my profile to another user's Gmail address! (since they changed it to something else)
> > 
> > 
> > Yeah.. this isn't something we could control. Maybe we could disable changing emails? But we need to think about a workflow for this
> 
> How does django-allauth handle this? Currently the email field is the one on django's User model. But I know allauth does email verification, so how do we use that?

django-allauth has a separate `EmailAddress` model that stores the one you used during registration. I think it would only start doing this after this setting `ACCOUNT_EMAIL_VERIFICATION = "optional"` was enabled


> What are these settings actually doing in our project?
> 
> > Set ACCOUNT_EMAIL_VERIFICATION = "optional" so an email is sent for verifying emails after sign up
> > Set ACCOUNT_UNIQUE_EMAIL = True to prevent duplicate emails since django-allauth also has their own way of handling emails
> 
> "an email is sent for verifying emails after sign up" <-- I haven't experienced this in any of the sign-up workflows (local or Google), so how do we use it?

From the docs:

> When set to “mandatory” the user is blocked from logging in until the email address is verified. Choose “optional” or “none” to allow logins with an unverified e-mail address. In case of “optional”, the e-mail verification mail is still sent, whereas in case of “none” no e-mail verification mails are sent.

Have you tried signing up as a new student? As you can see in this recording, a new tab opened after I signed up as a student.. I'm currently using [django-naomi](https://pypi.org/project/django-naomi/) to automatically open emails on browser but I think you could also see it inside the `_sent_mail` folder

https://user-images.githubusercontent.com/10972027/229258193-b4949e12-c39d-4c74-a38e-2d23cab18b6f.mp4

src/templates/account/email_confirm.html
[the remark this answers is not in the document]
yujinyuz: This is sent during sign up as an email. You can check the `_sent_mail` folder

src/hackerspace_online/adapter.py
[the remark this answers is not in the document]
yujinyuz: @tylerecouture 

I created a somewhat guide (for myself as well) using [gherkin](https://cucumber.io/docs/gherkin/reference/) syntax regarding the sign in. Hope this makes sense and let me know what you think:

```cucumber
Feature: New User sign in with Google

  Scenario: A new student decides to sign in to a deck account using their Google Account
    Given a new user with a google account: student@gmail.com

    When a user clicks the google sign in button

    Then they should be redirected to the google sign in page and enter their credentials

    When they enter the correct credentials

    Then they should be redirected to the sign up page and they should enter the required fields and correct access code

    When they hit submit

    Then they should now be logged in
----

Feature: User Email Confirmation

  Background:
    A student previously signed in via google with account student@gmail.com

  Scenario: User changes their email from student@gmail.com to new_student@gmail.com

    Given a user that signed in via google with email: student@gmail.com

    When they go to the profile page and change their email to new_student@gmail.com

    Then they hit submit

    Then they should receive a confirmation email to their new email new_student@gmail.com

    Then their social account (Google Sign in) should still work even after they changed their email since a user who signed in using a social provider does not necessarily mean that they cannot use any more. The only way where they won't be able to sign in is when they disconnect their social account from their bytedeck account

    Then they should still be able to sign in via google using student@gmail.com google account
```


I've tested social login with stackoverflow.. I signed up via email and connected my Github account. Changed my email and my Github is still connected. I think we should do the same with bytedeck to retain the google account they used for logging in even if they changed their email. All emails they receive will go to the new one though.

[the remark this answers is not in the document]
tylerecouture: > "Choose “optional” or “none” to allow logins with an unverified e-mail address. " 

We want to allow logins with NO email address, but if there IS an address it needs to be verified.  Is that an option?

Unverified email addresses are a problem for lots of reasons. E.g You can put someone else's email addy, or you might have a typo and then we get rejected notification/announcement emails etc.

tylerecouture (changes requested):
This is starting to look really good.  The big issue that remains is how to handle existing email addresses, and ensuring that all emails are verified.

I agree with your stack overflow example.  There's no need for the Google Auth email to be the same as the email in their profile.  But Stack Overflow still requires you to verify that different email address.

src/tenant/initialization.py
```diff
@@ -71,24 +77,24 @@ def set_initial_icons(object_list):
             with open(filepath, 'rb') as f:
                 object.icon.save(filepath, File(f), save=True)
                 object.save()
-        except IOError:
+        except OSError:
```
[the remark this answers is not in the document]
yujinyuz: This was because of the `pyupgrade` pre-commit hook..

https://stackoverflow.com/a/29347946/3788603

Also, IOError is just an alias of OSError

src/hackerspace_online/adapter.py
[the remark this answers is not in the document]
yujinyuz: > I think we were not talking about the same thing. My concern is existing users that have unverified emails in their User.email. There needs to be a way to migrate/convert/refactor these old accounts to the new system. We can't have two different email fields. All existing users in production will have this problem, so it needs to be resolved somehow.
> 
> Which of the two emails appears in their profile when they view it? When they change their email, which one is being changed?

When they view and change it, it's still the `User.email` that they will see and change.. I'm planning to create a post_save callback everytime a user changes their email to automatically send a confirmation email. In order to utilize allauth for sending confirmation emails, when the user changes their email, I just have to call `EmailAddress.objects.add_email(...).set_as_primary(...)`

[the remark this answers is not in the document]
yujinyuz: > We want to allow logins with NO email address, but if there IS an address it needs to be verified. Is that an option?
> 
> Unverified email addresses are a problem for lots of reasons. E.g You can put someone else's email addy, or you might have a typo and then we get rejected notification/announcement emails etc.

Unfortunately, this isn't an option but I'll try if I can hack this so we meet this requirement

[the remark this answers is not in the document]
tylerecouture: > > I think we were not talking about the same thing. My concern is existing users that have unverified emails in their User.email. There needs to be a way to migrate/convert/refactor these old accounts to the new system. We can't have two different email fields. All existing users in production will have this problem, so it needs to be resolved somehow.
> > 
> > Which of the two emails appears in their profile when they view it? When they change their email, which one is being changed?
> 
> When they view and change it, it's still the `User.email` that they will see and change.. I'm planning to create a post_save callback everytime a user changes their email to automatically send a confirmation email. In order to utilize allauth for sending confirmation emails, when the user changes their email, I just have to call `EmailAddress.objects.add_email(...).set_as_primary(...)`

So maybe to prevent confusion, anytime "email" is used in commits or docs or code comments or docstrings, it needs to specify which one: user.email or socialaccount.email? My preference would be to remove one from use completely.

What is the purpose of social account email? What is it currently used for? My understabding is that it's better than User.email because it comes with verification out of the box?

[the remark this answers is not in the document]
yujinyuz: It's only inside of this method that the `socialaccount.email` is used. This is the email we get from a social provider which is google in this case.

Then we just check if the one we got from google is being used as an email in `User.email`. If they do match, that's when we attempt to automatically connect a SocialAccount to a User.

Does that answer your questIon?

tylerecouture:
Yes! Ok, that makes sense.

So, then what email is being verified by all-auth? Is it User.email that is
being verified on new accounts?

src/hackerspace_online/adapter.py
```diff
@@ -16,6 +16,18 @@ def clean_username(self, username, shallow=False):
...
+        """
+        Invoked just after a user successfully authenticates via a
+        social provider, but before the login is actually processed
+        (and before the pre_social_login signal is emitted).
+
+        ---
+
+        Currently using this to automatically connect a google account
+        to an existing user provided that they used the same email address
+        for logging in
```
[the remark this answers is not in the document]
yujinyuz: The email address on the `User.email` is still important and what we will still be using.

I added new code so that every time User.email is changed, a confirmation email will be sent. Of course, they will be marked as unverified if they do so

src/profile_manager/forms.py
```diff
@@ -70,12 +80,13 @@ def save(self):
         # update self.instance since ProfileForm changes User model vars
         # super().save() not saving email correctly so we only update the model fields in UserForm.fields
         # This code only updates the UserModel fields that is in self._meta.fields
-        # required since Userform saves over user.email field which is also a field saved in ProfileForm. 
+        # required since Userform saves over user.email field which is also a field saved in ProfileForm.
         # This prevents email field from being overwritten in Userform
         user = self.instance
-        user_fields = list(set(self._meta.fields) & set([field.name for field in user._meta.fields]))
+        user_fields = list(set(self._meta.fields) & {field.name for field in user._meta.fields})
```
yujinyuz, line 86: this was suggested by pre-commit

yujinyuz:
> Yes! Ok, that makes sense.
>
> So, then what email is being verified by all-auth? Is it User.email that is
> being verified on new accounts?

* If a new user uses the Google sign in, the email they used is automatically verified and they need not to confirm their email
* On new accounts with emails, the email used for verifying is the `User.email`
* When an existing user with or without an email updates their email address, a confirmation email will be sent to `User.email` to verify their email address


Hope this makes sense

tylerecouture:
> * If a new user uses the Google sign in, the email they used is automatically verified and they need not to confirm their email
> * On new accounts with emails, the email used for verifying is the `User.email`
> * When an existing user with or without an email updates their email address, a confirmation email will be sent to `User.email` to verify their email address
> 
> 
> Hope this makes sense

Yes, ok, that all makes sense! Sorry for all the questions.

I have a much better understanding now. I think this will work as is, and no need to hack a solution for "either verified or no email"

Instead, it would be much simpler to
1. Add a django messages reminder, on login, to verify the email address, and
2. When sending bulk emails (notifications and announcements) only send to verified emails...filter out the unverified.

Does that make sense to you?

yujinyuz:
@tylerecouture Let me know if you have anything else you would like to add. 

I think this PR is good and has a nice test coverage

tylerecouture (changes requested):
I'm slowly working my way through the code and testing stuff.  One bug I think I came accross:
1. a new deck with sample data installed (all users have first.last@example.com email addresses on their profile)
2. The deck is NOT enabled for Google sign in (default)
3. Edit a user profile then save (no changes needed, I did it from admin account editing a student profile)
4. Stacktrace below.
5. If you can reproduce, please create a test to catch and then repair. Thank you!
````Python
Internal Server Error: /profiles/85/edit/
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/profile_manager/views.py", line 190, in dispatch
    return super().dispatch(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/tenant/views.py", line 37, in wrapper
    return f(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/tenant/views.py", line 46, in dispatch
    return super().dispatch(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/profile_manager/views.py", line 199, in post
    return self.form_valid(forms)
  File "/home/tylere/Developer/bytedeck/src/profile_manager/views.py", line 233, in form_valid
    form.save()
  File "/home/tylere/Developer/bytedeck/src/profile_manager/forms.py", line 57, in save
    send_email_confirmation(
TypeError: send_email_confirmation() got an unexpected keyword argument 'email'
Internal Server Error: /profiles/85/edit/
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/profile_manager/views.py", line 190, in dispatch
    return super().dispatch(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/utils/decorators.py", line 43, in _wrapper
    return bound_method(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/tenant/views.py", line 37, in wrapper
    return f(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/tenant/views.py", line 46, in dispatch
    return super().dispatch(*args, **kwargs)
  File "/home/tylere/Developer/bytedeck/venv/lib/python3.8/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/home/tylere/Developer/bytedeck/src/profile_manager/views.py", line 199, in post
    return self.form_valid(forms)
  File "/home/tylere/Developer/bytedeck/src/profile_manager/views.py", line 233, in form_valid
    form.save()
  File "/home/tylere/Developer/bytedeck/src/profile_manager/forms.py", line 57, in save
    send_email_confirmation(
TypeError: send_email_confirmation() got an unexpected keyword argument 'email'
````

File: README.md [hunk 12]
```diff
@@ -129,8 +129,48 @@ New tenants will come with some basic initial data already installed, but if you
...
+```
+
+2. We need to update the`ALLOWED_HOSTS` in our .env file:
 
+```bash
+ALLOWED_HOSTS=.localhost,.localhost.net
+```
+
+3. For the final step, we need to let `django-tenants` know that `hackerspace.localhost.net` is also a valid domain.
+Run `$ ./src/manage.py shell` and type in the following commands
```
tylerecouture, line 164: I'm confused about this step. Do I need to access the shell in production every time a deck owner wants Google Sign in? I thought this was done automatically... or we could do from public tenant admin or something? I remember us discussing this but now I'm confused again =)

tylerecouture:
> I'm slowly working my way through the code and testing stuff. One bug I think I came across:
I'm getting this error even after I enable Google Sign in on the domain.  I must have screwed something up or missed a step.

README.md
tylerecouture: This is just for dev environment isn't it?

yujinyuz: Ahh. This must've happened when I was fixing a merge conflict. Let me fix this first

README.md
```diff
@@ -129,8 +129,48 @@ New tenants will come with some basic initial data already installed, but if you
...
 
+Here are the steps, assuming that you now have a functional tenant:
+
+1. Obtain Google credentials: https://developers.google.com/workspace/guides/create-credentials#oauth-client-id
+2. Make sure that in the Authorized URIs, add `http://hackerspace.localhost.net:8000/accounts/google/login/callback/`. We will explain why we are using `localhost.net` later but for now, just add this.
+3. Go to Social Applications: http://localhost:8000/admin/socialaccount/socialapp/
+4. Click Add Social Application
+5. Fill in `Client Id` and `Secret Key`. And then add the `Available Sites` to `Chosen Sites`
+6. Click Save
+7. Go to your Site Configuration: http://hackerspace.localhost:8000/config/ and click `Enable sign-in via Google`
```
tylerecouture, line 143: This step needs to be updated

src/announcements/tasks.py
```diff
@@ -35,12 +33,14 @@ def send_notifications(user_id, announcement_id):
 def get_users_to_email():
     students_to_email = list(
         CourseStudent.objects.all_users_for_active_semester()
+                             .filter(emailaddress__verified=True, emailaddress__primary=True)
```
tylerecouture, line 36: What's emailaddress__primary?

src/hackerspace_online/forms.py
```diff
@@ -37,11 +47,19 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields['username'].help_text = 'Ask your teacher what you should be using for your username. Username is not case sensitive'
```
tylerecouture, line 48: While you are in here, can you update the help text?
`self.fields['username'].help_text = 'Username is not case sensitive.'

src/hackerspace_online/signals.py
```diff
@@ -24,3 +26,16 @@ def change_domain_urls(sender, *args, **kwargs):
                 tenant.save()
             public_tenant.domain_url = kwargs['instance'].domain
             public_tenant.save()
+
+
+def handle_tenant_site_domain_update(tenant, **kwargs):
```
tylerecouture, line 31: Docstring: what and why

src/profile_manager/forms.py
```diff
@@ -18,14 +19,28 @@ class Meta:
...
         super().__init__(*args, **kwargs)
 
         self.fields['grad_year'] = forms.ChoiceField(
             choices=Profile.get_grad_year_choices()
         )
-        
         self.fields['email'].initial = self.instance.user.email
 
+        user = self.instance.user
+        email_address = user.emailaddress_set.filter(email=user.email).first()
```
tylerecouture, line 32: Maybe a comment explaining why `first()`

src/profile_manager/models.py
```diff
@@ -423,6 +428,32 @@ def post_delete_user(sender, instance, *args, **kwargs):
         instance.user.delete()
 
 
+@receiver(email_confirmed, sender=EmailConfirmationHMAC)
+def email_confirmed_handler(email_address, **kwargs):
+    """
+    After a user has confirmed their email, this will handle converting that email to the primary email
+    and therefore making it the User.email.
+
+    Then, we delete the non-primary email addresses of that user
```
tylerecouture, line 437: What is the non-primary email address?

src/profile_manager/models.py
```diff
@@ -448,7 +479,7 @@ def smart_list(value, delimiter=",", func=None):
     model signal to ensure that you always get a list back from the field.
 
     """
-    if value in ["", u"", "[]", u"[]", u"[ ]", None]:
+    if value in ["", "", "[]", "[]", "[ ]", None]:
```
tylerecouture, line 482: Hmm... this looks sus.  Now it's just repeating values.  Obviously `u""` isn't tested, or this change would fail, but unless you have a specific reason for changing this, please leave as is.

src/profile_manager/models.py
```diff
@@ -423,6 +428,32 @@ def post_delete_user(sender, instance, *args, **kwargs):
...
+    Then, we delete the non-primary email addresses of that user
+    """
+
+    with transaction.atomic():
+        email_address.set_as_primary()
+        EmailAddress.objects.filter(user=email_address.user, primary=False).delete()
+
+
+@receiver(user_logged_in, sender=User)
+def user_logged_in_verify_email_reminder_handler(request, user, **kwargs):
```
tylerecouture, line 446: I don't see a test for this, but I may have missed it.

src/profile_manager/views.py
```diff
@@ -276,6 +280,31 @@ def get_success_url(self):
...
 
 
+@non_public_only_view
+@login_required
+def profile_resend_email_verification(request):
+
+    user = request.user
+    email_address = EmailAddress.objects.filter(email=user.email).first()
+
+    if not email_address or not user.email:
```
tylerecouture, line 290: Under what circumstances would this occur?  Maybe a comment to explain.

src/profile_manager/views.py
```diff
@@ -276,6 +280,31 @@ def get_success_url(self):
...
+        messages.error(request, "User does not have an email")
+        return redirect_to_previous_page(request)
+
+    if email_address.verified:
```
tylerecouture, line 294: Under what circumstances would this occur?  Maybe a comment to explain.

src/announcements/tasks.py
yujinyuz: django-allauth has a `EmailAddress` model which has a relationship to `User`. We don't want that feature where a user can have multiple email address so everytime they update their email via `ProfileForm`, there is a receiver `email_confirmed_handler` that calls `email_address.set_as_primary()` everytime an email is confirmed

src/profile_manager/models.py
yujinyuz: For context: https://github.com/bytedeck/bytedeck/pull/1300/files#r1161206616

When a user adds or updates their email address in the profile page, it creates new record for the allauth model `allauth.account.models.EmailAddress`.

So, after they confirmed their email, the new one becomes the primary and the old one becomes the non-primary, so we just delete that

src/profile_manager/models.py
yujinyuz: https://github.com/bytedeck/bytedeck/pull/1300/files#diff-d20ce8cc823f496ca1b2f26080ff2dc503de9450495c6efb9f35c79fb338babfR431

src/templates/account/login.html
```diff
@@ -1,23 +1,20 @@
...
 
 {% block head_title %}{% trans "Sign In" %} | {% endblock %}
 
 {% block heading %}{% trans "Sign In" %}<i class="pull-right fa fa-sign-in"></i>{% endblock %}
 
 {% block content %}
-  {% if socialaccount.providers %}
-    <p>{% blocktrans with site.name as site_name %}Please sign in with one
-    of your existing third party accounts. Or, <a href="{{ signup_url }}">sign up</a>
+  {% if config.enable_google_signin %}
```
tylerecouture, line 11: I think these instruction could flow more logical, something like this?  Also, probably shouldn't say "Google" before the list that could in the future have other providers
````
  <p>Sign in via:</p>

    <div class="socialaccount_ballot">

      <ul class="socialaccount_providers">
        {% include "socialaccount/snippets/provider_list.html" with process="login" %}
      </ul>
      <a href="{% provider_login_url 'google' process='login' %}" class="btn btn-success"><i class="fa fa-google"></i>oogle</a>

      <div class="login-or">{% trans 'or' %}</div>

    </div>
  {% include "socialaccount/snippets/login_extra.html" %}
  {% endif %}

    <p>{% blocktrans %}Sign in with your username and password. If you have not created an account yet, then please
    <a href="{{ signup_url }}">sign up</a> first.{% endblocktrans %}</p>
  {% endif %}
  <form class="login" method="POST" action="{% url 'account_login' %}">
````

File: src/profile_manager/models.py (hunk 19 above)
Comment (yujinyuz) [reply src/profile_manager/models.py:482]: I think this is from one of the pre-commit hooks and I think the `u` is not relevant anymore as of python 3

File: src/profile_manager/views.py (hunk 22 above)
Comment (yujinyuz) [reply src/profile_manager/views.py:294]: Will add a comment. But this would only happen if they happen to have access to the resend verification url or click it twice

File: src/tenant/admin.py [hunk 24]
```diff
@@ -109,5 +112,49 @@ def get_queryset(self, request):
...
+            with tenant_context(tenant):
+                config = SiteConfig.get()
+                with transaction.atomic():
+                    config._propagate_google_provider()
+                    config.enable_google_signin = True
+                    config.save()
+
+        enabled_count = queryset.count()
+        self.message_user(request, ngettext(
+            "%d tenant google signin was enabled successfully. Please ensure that the tenant domain is added in the Authorized Redirect URIs",
```
tylerecouture, line 136: `... Ensure that the {tenant_domain} is added in the Authorized Redirect URIs on Bytedeck's Google credentials.`

Can you put the {tenant_domain} here, whatever it is for easy copy/paste?

src/templates/account/login.html
tylerecouture, line 1: We need to update the Sign Up template too, currently it doesn't mention Google option even if Google is turned on:
image

src/profile_manager/models.py
tylerecouture: Can you add all that into the docstring?

src/profile_manager/models.py
tylerecouture: You're right! In that case, we can remove the repeated list items.

src/profile_manager/models.py
tylerecouture: Excellent.  make sense to have this covered in an integrated test.  Can you add a comment with the method name at this line, like:
# test of method: user_logged_in_verify_email_reminder_handler()

And if you are integrating tests of other methods, do the same thing?  Makes it easier to find with CTRL+F, and is one of the reason we use test naming convention:

`test_method_name__specific_details_being_tested`

Hmm... I guess that isn't documented anywhere.  I need to add that under contributing.

src/tenant/admin.py
yujinyuz: Does this look okay-ish?

image

README.md
yujinyuz: > > I'm slowly working my way through the code and testing stuff. One bug I think I came across:
> 
> I'm getting this error even after I enable Google Sign in on the domain. Maybe I screwed something up or missed a step?

Try upgrading django-allauth to the one specified in the requirements file. This happened to me when I was using the old version

src/templates/account/login.html
yujinyuz: Does this look right?

image

src/tenant/admin.py
tylerecouture: Yes! But please add "..on Bytedeck's Google account credentials." to the end, because the person doing this will not be a developer and the prompt will help, especially if we don't get this request (to turn on Google sign in) very often.

src/templates/account/login.html
tylerecouture: Needs some spacing after the Google button, and an "or" after the button.

tylerecouture: Does Remember me work for Google login as well?

README.md
tylerecouture: Is there a way we can automatically catch outdated dependancies?  I don't know what kind of hook or test or..something that would be.  Something that automatically re-checks or attempts to reinstall requirements.txt.  Any ideas?

src/profile_manager/forms.py
```diff
@@ -18,14 +19,32 @@ class Meta:
...
+
+        # It is possible that the user registered without an email so instead of using
+        # user.emailaddress_set.get(...), let's just use .first() since it would return
+        # None instead of raising a DoesNotExist error
+        email_address = user.emailaddress_set.filter(email=user.email).first()
+
+        if (user.email and email_address is None) or (email_address and email_address.verified is False):
+            resend_email_url = reverse('profiles:profile_resend_email_verification', args=[self.instance.pk])
+            self.fields['email'].help_text = (
+                '<i class="fa fa-info"></i> Not yet verified. '
```
tylerecouture, line 41: use:
`'<i class="fa fa-ban text-danger" aria-hidden="true"></i> Not yet verified. '`

src/profile_manager/forms.py
```diff
@@ -18,14 +19,32 @@ class Meta:
...
+                f'<a href="{resend_email_url}">Re-send verification link</a>'
+            )
+
+        if email_address and email_address.verified:
+            self.fields['email'].help_text = '<i class="fa fa-check info"></i>Verified'
```
tylerecouture, line 46: `'<i class="fa fa-check text-success" aria-hidden="true"></i>&nbsp;Verified'`

src/templates/account/signup.html
tylerecouture, line 1: Something funky is going on here:
image

src/profile_manager/models.py
tylerecouture, line 1: I got an error when I tried to change an email to one that already exists in another profile.  Please cover this with a test and fix.  This should invalidate the form shouldn't it?
image

tylerecouture, line 1: Which leads to other interesting stuff...
image

tylerecouture (changes requested):
This is really cool.  I actually got to play around with it a bit and I'm very excited for this!!

I think I found one bug and left a few other comments.

src/templates/account/login.html
yujinyuz: Nope. I think it will remember you as long as your google account is still signed in to the same browser you're using

src/profile_manager/models.py
yujinyuz: Thanks for finding this. Will fix this

yujinyuz:
@tylerecouture could you test this again and see if everything works?

I've resolved the ones you posted

tylerecouture (changes requested):
Button missing?  Also, maybe change the "or" to use p instead of div to get proper margins:
image

Sign up page also missing google option.  Isn't there a test that check if "Google" or something similar appears (a link?) if Google Signon is turned on for a deck?
image

yujinyuz:
@tylerecouture thanks. should be fixed now

yujinyuz:
seems like tests are failing. looking into it

tylerecouture (changes requested):
The "or" is in the wrong spot for the Sign up page.  It should be "Sign up with Google OR ... the completed the Sign up form:

image

Also, now that we're almost ready to merge, I don't like the custom/non standard button.  I think we should just use the standard button (dark version) and follow Google's branding guide for it, it looks much more professional:
https://developers.google.com/identity/branding-guidelines

tylerecouture:
If I try to edit the email field in a profile and then save:
image

tylerecouture:
The Profile form needs a test to cover changing the email!

yujinyuz:
@tylerecouture can you update your allauth package version?

yujinyuz:
@tylerecouture I could add a test for this but then if you upgrade the package, it would never fail and then there would be no way to test the older version so I'm not too sure whether we should actually add tests for this

tylerecouture:
No, I'm sorry, I didn't notice the package upgrade 🤦‍♂️

tylerecouture:
If I change my email and verify, then change it back to my original google Auth sign in email, it still asks me to verify the original email address. Shouldn't it be able to tell that this is my google OAuth sign in email and thus already verified by comparing to my authentication email address?

tylerecouture:
When I sign up for a new account and provide an email address, it shows me two different verification messages (though it only sent one email)
image

Can we remove one of them in this case?

tylerecouture:
I signed up with a local account and entered a gmail address in the optional email field (not using Google OAuth).  I then signed out, and tried to sign in using Google OAuth with that same UNVERIFIED email address.  It connected the accounts.  I don't know if that's what we want?

It should probably only auto link to accounts with matching VERIFIED address, or ask before linking an account with a matching, but unverified email.  What do you think?

The person using OAuth obviously owns that email.  But the local account who used the email, we don't know if they own it because they didn't verify it.  Does that make sense?

Because this involves accounts and authentication, I wan't to make sure we consider these edge cases.

I don't know if this is important or not, I just want to make sure we considered the case.

yujinyuz:
> If I change my email and verify, then change it back to my original google Auth sign in email, it still asks me to verify the original email address. Shouldn't it be able to tell that this is my google OAuth sign in email and thus already verified by comparing to my authentication email address?

yeah but I added a code to delete previous email addresses entered. I think that doesn't make sense with this scenario so I'll remove that instead

yujinyuz:
> When I sign up for a new account and provide an email address, it shows me two different verification messages (though it only sent one email) image
> 
> Is this intentional or can we remove one of them in this case?

This is unintentional. I didn't think signing up as a new user and then it automatically logs you in. We have a signal whenever a user logs in to remind them to verify their email. Not quite sure yet how to prevent this but will take a look

yujinyuz:
> I signed up with a local account and entered a gmail address in the optional email field (not using Google OAuth). I then signed out, and tried to sign in using Google OAuth with that same UNVERIFIED email address. It connected the accounts. I don't know if that's what we want?
> 
> It should probably only auto link to accounts with matching VERIFIED address, or ask before linking an account with a matching, but unverified email. What do you think?
> 
> The person using OAuth obviously owns that email. But the local account who used the email, we don't know if they own it because they didn't verify it. Does that make sense?
> 
> Because this involves accounts and authentication, I wan't to make sure we consider these edge cases.
> 
> I don't know if this is important or not, I just want to make sure we considered the case.

Yeah. This is gonna add another extra step. Maybe I should just add a message.info that your existing account has been merged to your google and you can login with both

yujinyuz:
I think the best way to handle when email is verified and they login via google = automatically merge.

But if both unverified, then it should be redirected to a new page that has some text like

```
We noticed you already have an unverified account and attempted to sign in via Google. Would you like to merge your accounts? Proceed | Cancel
```

If they click proceed, then it should automatically merge accounts. But I'm not too sure about clicking `Cancel`.

---

I also haven't experienced this auto merging of accounts in other sites though.. at least I haven't tried so I'm not entirely sure what's the proper way of implementing this

tylerecouture:
> But I'm not too sure about clicking `Cancel`.

How about this, instead of saying cancel, just say "Yes" and "No" and if they click No, then remove the unverified email from that other account (unverified emails aren't used for anything anyway)

tylerecouture:
Here: 

> We noticed another account with username "{{other_account_username}}" also using {{email_address}} in their profile, but they have not verified the email address to that account. Would you like to merge the existing "{{other_account_username}}" account with this one? If you choose No, then {{email}} will be removed from the  "{{other_account_username}}" profile. Yes | No

yujinyuz:
@tylerecouture

Sounds good. But if the user also owns that account and chooses No, then what happens?

Coz if that user logged in via gmail, gets redirected to that prompt and chooses `No`: the email will be removed but then gets added back as a verified email. The behavior is still the same as the current implementation

tylerecouture:
> @tylerecouture
> 
> the email will be removed but then gets added back as a verified email. The behavior is still the same as the current implementation

Can you clarify what you mean here?

yujinyuz:
> > @tylerecouture
> > the email will be removed but then gets added back as a verified email. The behavior is still the same as the current implementation
> 
> Can you clarify what you mean here?

I think I just misunderstood something here.. I thought we were gonna let them login whether they choose yes or no.

---

Let me clarify:

So, if I own `test_student` account, has an unverified email of `test_student@gmail.com`,

I log in via google sign in using `test_student@gmail.com`, then I get redirected to that page with clicking Yes|No.

If I click yes, then accounts gets merged.

But if I click No, the `test_student@gmail.com` email is removed. BUT since we are logging in via google and I click No, do I get redirected to the sign up page instead?

tylerecouture:
Yes, that's what I was imagining. 

If they click "No" then they'll be creating a new account.

yujinyuz:
Alright. That makes more sense. Thanks

tylerecouture:
Yes, that's what I was imagining. 

If they click "No" then they'll be creating a new account.

To be clear, the unverified email is removed from the account that had the email (unverified) in it's profile.

tylerecouture:
The problem is, if the email is unverified, the app doesn't actually know if it's your account or not, it could be anyone who entered that email.

yujinyuz:
@tylerecouture finished adding the new oauth merge account page

tylerecouture:
@yujinyuz did you see my comment somewhere about using the official Google sign in image instead of our own custom thing?  I can't find it now, but this looks much more professional:
https://developers.google.com/identity/branding-guidelines

yujinyuz:
@tylerecouture I haven't seen your comment about it but I can check this out

yujinyuz:
@tylerecouture Which button design would you prefer though?

Google provides multiple options

tylerecouture:
Blue option at the top

yujinyuz:
https://user-images.githubusercontent.com/10972027/232382358-411c8838-5552-4481-a2ac-9fc79b7b3a01.mp4

Issue #1298, opened 2023-03-08
Keep users signed in for 8 Weeks


tylerecouture:
- When user logs in the "Remember Me" appears to currently does nothing.  It should keep the user logged in (via authentication cookie or other method)
- Default expiry of authentication cookies to 8 weeks.

![image](https://user-images.githubusercontent.com/10604391/223629216-dda2533e-4800-4c14-8fa8-ebeca66f17ba.png)

PR #1306, opened 2023-03-22
Set login session expiry to 8 weeks

yujinyuz:
### What?

I've added a fix that made the `Remember Me?` checkbox not work as it relies on the `SESSION_COOKIE_AGE`.

### Why?

These changes complete the requirements for issue #1298 

### How?

django-allauth uses the builtin `SESSION_COOKIE_AGE` of Django and uses that variable for setting the expiry of a cookie. Previously, since it is unset, it would just set it to `0`.

You can take a look at their implementation here: https://github.com/pennersr/django-allauth/blob/99b67e8db3795ff1ebc5eb27b4a843f7c2c39591/allauth/account/forms.py#L206-L207

### Testing?

Added a test that the session automatically expires depending on the value of `SESSION_COOKIE_AGE`.

It's quite hard to use seconds during tests since there might be some discrepancies as to when was the `User.last_login` set and when was the session's expiry is set.

First part is leaving the `Remember me` unchecked and closing the browser automatically logged them out.

Second part is checking the `Remember me` and closing the browser still leaves them logged in.

Tested on Firefox

https://user-images.githubusercontent.com/10972027/227086943-b9771777-6ee9-4b79-be8a-c5b925542b8a.mp4

### Review request
@tylerecouture

### DIFF
--- a/src/hackerspace_online/settings.py
+++ b/src/hackerspace_online/settings.py
@@ -497,6 +497,7 @@ SILENCED_SYSTEM_CHECKS += ['captcha.recaptcha_test_key_error']
 
 
 # AUTHENTICATION ##################################################
+SESSION_COOKIE_AGE = env("SESSION_COOKIE_AGE", default=int(60 * 60 * 24 * 7 * 8))  # 8 Weeks
 
 AUTHENTICATION_BACKENDS = (
 
--- a/src/hackerspace_online/tests/test_forms.py
+++ b/src/hackerspace_online/tests/test_forms.py
@@ -1,3 +1,6 @@
+from django.conf import settings
+from django.test.utils import override_settings
+from datetime import timedelta
 from django import forms
 from django.contrib.auth import get_user_model
 from django.shortcuts import reverse
@@ -87,7 +90,7 @@ class CustomSignUpFormTest(TenantTestCase):
 class CustomLoginFormTest(TenantTestCase):
 
     def setUp(self):
-        User.objects.create_user(username='testuser', password='testuser')
+        self.user = User.objects.create_user(username='testuser', password='testuser')
 
     def test_init(self):
         CustomLoginForm()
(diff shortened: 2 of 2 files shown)


tylerecouture (changes requested):
Even if you can't unit test this, can you manually test it?  How does this connect to the checkbox when signing in, and does it work?  Can you set it to something lower locally, like 30 seconds, and verify it works as expected (and then describe the manual testing you did to verify).

You can also test current production (master branch) behaviour on our testing tenant: https://testing.bytedeck.com/accounts/login/  (course code 314159).

I just tried production and the tick box does something, because if I don't check it I'm not logged in after closing browser, but if I do check it, I am logged in.  But no idea for how long.

tylerecouture (changes requested):
Before I close this issue, can you verify the behaviour of the remember me button?  Your video is good in that it auto-logs out the user after the session cookie expires.... but what about the remember me button...what is that doing?  

This is the behaviour I'm expecting:
* If you tick "remember me" when you log in, then close your browser:
    * keeps you logged in next time you open the browser... but for how long does it "remember" you are logged in?
* If you DO NOT tick "remember me" when you log in, then close your browser, it definitely should not be logged in when you re-open the browser.

Is this the current behaviour? And if not, what is it, and does it makes ssense or deoes it need fixing?  How does the "Remember Me" tick box relate to the SESSION_COOKIE_AGE, if it is at all?

yujinyuz:
@tylerecouture 

Oh okay. I think I get what you're saying and probably had a misunderstanding on my end. I guess the PR title is quite misleading since this doesn't fix the remember me but rather sets the `Remember me` functionality to remember it for 8 weeks. The default is 2 weeks that's why it was working when you tested it

I updated the description and the video that tests the remember me function and it works as expected.


---

Based on the django documentation

image


So, when the `Remember me` is left unchecked, `django-allauth` calls `request.session.set_expiry(0)` and this is what it means for Django

> If value is 0, the user’s session cookie will expire when the user’s Web browser is closed.

Issue #1302, opened 2023-03-15
Implement two Summernote fields: 1. escaped/safe, and 2. unescaped


tylerecouture:
Note that  I am going to use the `Quest.instructions` field for all examples, and the form at `quest_manager.forms.QuestForm`, although the problem exists everywhere we use this widget:
![image](https://user-images.githubusercontent.com/10604391/225453389-10879543-068e-4dea-9c74-82502b6aa5ba.png)

### Problems
Currently we use the `SummernoteInplaceWidget()` from [django-summernote ](https://github.com/summernote/django-summernote )for most of our `models.TextField` fields.

This provides us with a simple, expandable, configurable wysiwyg editor for our forms.

####  1. Undesired escaping/stripping behaviour: 
The current implementation has some major issues, demonstrated in the screen recording using `<script>alert("Hello")</script>`:

1. Enter the content: `<script>alert("Hello")</script>` into the summernote widget in the form
2. When saved, the text appears as expected.
3. However, when you edit the form, it executes the code!  You can see the alert pop up.
4. The text is also converted into code, so that it is no longer visible unless you switch to "Code View"
5. When switching OUT of "Code View", it appears to do some escaping and strips out the `<script>` tags

![Peek 2023-03-15 14-45](https://user-images.githubusercontent.com/10604391/225451196-d190f532-3afc-465b-a962-2aff0112041a.gif)

#### 2. iFrame version of the widget is bad:
A potential solution is to consider summernote's iFrame version of the widget, From [django-summernote](https://github.com/summernote/django-summernote) docs: 

> Warning: Please mind, that the widget does not provide any escaping. If you expose the widget to external users without taking care of this, it could potentially lead to an injection vulnerability. Therefore you can use the SummernoteTextFormField or SummernoteTextField, which escape all harmful tags through mozilla's package bleach:

However, in the past, SummernoteTextFormField or SummernoteTextField were only usable on the widget that used an iframe, which had other problems that made the widget unusable or undesirable, as shown here: (scroll bars and popup menus that are useless):
![image](https://user-images.githubusercontent.com/10604391/225454538-fad6c69e-7d19-454f-9b3c-9d6ccebe669a.png)

#### 1. Ability to choose escaping/stripping or not on a per-widget basis:

In our app, we need two versions of this widget: 

1. A "Safe" widget that escapes content entered in the text view (e.g. `<` are converted in the code to `&lt;`) and also strips script tags entered in the HTML/code view, 
[...]

PR #1307, opened 2023-03-23
WIP: implement SummernoteInplaceSafeWidget

bashu:
### What?

I've implemented `ByteDeckSummernoteInplaceWidget` class as substitute for `SummernoteInplaceWidget` from `django-summernote`

### Why?

See https://github.com/bytedeck/bytedeck/issues/1302

### How?

append `bytedeck_summernote` right after `django_summernote` in INSTALLED_APPS
```python
INSTALLED_APPS = [
...
'django_summernote',
'bytedeck_summernote',
]
```

update urls.py, replace `django_summernote.urls` with `bytedeck_summernote.urls`:

```python
urlpatterns += [
...
    # bytedeck summernote
    url(r'^summernote/', include('bytedeck_summernote.urls')),
]
```

Replace imports:
```python
from django_summernote.widgets import SummernoteInplaceWidget, SummernoteWidget

# and
from django_summernote.admin import SummernoteModelAdmin, SummernoteInlineModelAdmin
```
with

```python
from bytedeck_summernote.widgets import ByteDeckSummernoteSafeWidget, ByteDeckSummernoteAdvancedWidget

# and
from bytedeck_summernote.admin import ByteDeckSummernoteModelAdmin, ByteDeckSummernoteInlineModelAdmin

```

### Testing?
Yes

summernoteinplacewidget

### Anything Else?

fork django-summernote under your project/company account, name the fork as bytedeck-summernote or django-summernote-bytedeck

### Review request
@tylerecouture

### DIFF
--- a/src/announcements/forms.py
+++ b/src/announcements/forms.py
@@ -4,7 +4,8 @@ from django import forms
 from django.utils import timezone
 
 from bootstrap_datepicker_plus.widgets import DateTimePickerInput
-from django_summernote.widgets import SummernoteInplaceWidget
+
+from bytedeck_summernote.widgets import ByteDeckSummernoteSafeWidget
 
 from .models import Announcement
 
@@ -26,7 +27,7 @@ class AnnouncementForm(forms.ModelForm):
         # > If you don't like <iframe>, then use inplace widget
         # > Or if you're using django-crispy-forms, please use this.
         widgets = {
-            'content': SummernoteInplaceWidget(),
+            'content': ByteDeckSummernoteSafeWidget(),
             'sticky_until': DateTimePickerInput(format='%Y-%m-%d %H:%M'),
             'datetime_released': DateTimePickerInput(format='%Y-%m-%d %H:%M'),
             'datetime_expires': DateTimePickerInput(format='%Y-%m-%d %H:%M'),
--- a/src/bytedeck_summernote/templates/bytedeck_summernote/widget_common.html
+++ b/src/bytedeck_summernote/templates/bytedeck_summernote/widget_common.html
@@ -1,0 +1,170 @@
+<script>
+ function initSummernote_{{ id_safe }}() {
+     {{ config.jquery }}(document).ready(function($) {
+{% if iframe %}
+         var frameId = window.frameElement.id;
+         var id = frameId.substring(0, frameId.length - 7);  // without `_frame`
+         var iframe = window.parent.document.getElementById(frameId);
+         var origin = window.parent.document.getElementById(id);
+         var csrftoken = '{{ csrf_token }}';
+         var settings = window.parent.settings_{{ id_safe }};
+         var $sn = $('#summernote');
+         $(iframe).height(settings.height);
+{% else %}
+         var origin = window.document.getElementById('{{ id }}');
+         var csrftoken = getCookie('{{ CSRF_COOKIE_NAME }}');
+         var settings = {{ settings|safe }};
+         var $sn = $(origin);
+{% endif %}
+{% comment %} Start changes from source at https://github.com/summernote/django-summernote/blob/main/django_summernote/templates/django_summernote/widget_common.html
+
+Changes because: fix django-summernote/issues/465 issue, make it possible to use XSS protection for non-iframe widgets
+{% endcomment %}
+{% if iframe %}
+         $sn.html(origin.value);
+         $(origin).hide();
+{% endif %}
+{% comment %} End changes {% endcomment %}
+         var $nEditor, $nCodable, $nImageInput;
+
+         function recalcHeight(e) {
+             var nEditable = e.find('.note-editable');
+             var nEditor = $('.note-editor');
+             var height = parseInt(
+                 parseInt(settings.height)  // default
+                 - e.find('.note-toolbar').outerHeight()
+                 - e.find('.note-status-output').outerHeight()
+                 - e.find('.note-statusbar').outerHeight()
+                 - (nEditor.outerHeight() - nEditor.innerHeight())  // editor's border
+             );
+             nEditable.outerHeight(height);
+         }
+
+         var initEditor = function() {
+             $sn.summernote($.extend(settings, {
+{% if iframe %}
+                 disableResizeEditor: true,
+{% endif %}
+                 callbacks: {
+                     onInit: function() {
+                         $nEditor = $sn.next();
+                         $nCodable = $nEditor.find('.note-codable');
+                         recalcHeight($nEditor);
+
+{% if iframe %}
+                         // Resize again when the height of note-toolbar changed
+                         // This enhances calculation behaviors on Safari
+                         new ResizeSensor($nEditor.find('.note-toolbar'), function() {
+                             recalcHeight($nEditor);
+                         });
+                         // Add an event handler for fullscreen
+                         $nEditor.find('.btn-fullscreen').on('click', {editor: $nEditor}, function (e) {
+                             var $nEditor = e.data.editor;
+                             if ($nEditor.hasClass('fullscreen')) {
+                                 $(iframe).addClass('note-fullscreen');
+                             } else {
+                                 $(iframe).removeClass('note-fullscreen');
+                             }
+                         });
+{% else %}
+                         // Apply attributes with wrapper div
+                         var $wrap = $($nEditor).wrap(
+                             '<div class="summernote-div"></div>'
+                         ).parent();
+                         {% for k, v in attrs.items %}
+                         $wrap.attr('{{ k }}', '{{ v }}');
+                         {% endfor %}
+{% endif %}
+
+                         // Move dropdown dialog when it exceeds the bound of the editor
+                         $nEditor.find('.note-btn-group').on('shown.bs.dropdown', function () {
+                             var nDropdown = $(this).find(".dropdown-menu");
+                             var offset = nDropdown.offset();
+                             var width = nDropdown.width();
+
+                             var windowWidth = $(window).width();
+                             var margin = 15;
+
+                             if( offset.left + width + margin > windowWidth ) {
+                                 var left = windowWidth - width - offset.left - margin;
+                                 nDropdown.css("left", left + "px");
+                             }
+                         });
+                     },
+                     onBlur: function() {
+                         origin.value = $sn.summernote('code');
+                     },
+                     onBlurCodeview: function() {
+                         origin.value = $sn.summernote('code');
+                     },
+                     {% if not config.disable_attachment %}
+                     onImageUpload: function(files) {
+                         // custom attachment data
+                         var attachmentData = origin.dataset;
+                         $nImageInput = $nEditor.find('.note-image-input');
+                         $nImageInput.fileupload();
+                         var jqXHR = $nImageInput.fileupload('send',
+                                                             {
+                                                                 files: files,
+                                                                 formData: $.extend({csrfmiddlewaretoken: csrftoken}, attachmentData),
+                                                                 url: settings.url.upload_attachment,
+                         })
+                                                 .done(function (data, textStatus, jqXHR) {
+                                                     $.each(data.files, function (index, file) {
+                                                         $sn.summernote("insertImage", file.url);
+                                                     });
+                                                 })
+                                                 .fail(function (jqXHR, textStatus, errorThrown) {
+                                                     // if the error message from the server has any text in it, show it
+                                                     var msg = jqXHR.responseJSON;
+                                                     if (msg && msg.message) {
+                                                         alert(msg.message);
+                                                     }
+                                                     // otherwise, show something generic
+                                                     else {
+                                                         alert('Got an error while uploading images.');
+                                                     }
+                                                 });
+                         // Summernote will replace ImageInput when next the dialog opens.
+                         // But we have to reset it to prevent additional uploading by form itself.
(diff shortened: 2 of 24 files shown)


tylerecouture (changes requested):
For the SafeWidget: 

This shouldn't strip tags when they are entered in the wysiwyg editor, instead they should be escaped.

**WYSIWYG entry:**  `<script>alert("Hello")</script>`

In code view, after escaping:
* `&lt;script&gt;alert("Hello")&lt;/script&gt;`
or 
* `&lt;script&gt;alert(&quot;Hello&quot;)&lt;/script&gt;`

Rendered as: `<script>alert("Hello")</script>`

**Code view entry:**  `<script>alert("Hello")</script>`

This must be stripped of the script tags, or all content can be deleted. Importantly, this must not be executed when saving or toggling back to the wysiwyg editor:

After stripping: 
* `alert("Hello")` <-- current default behaviour of the widget IF you switch out of code view before saving!  But if you save the form while still in code view, it executes the code.

bashu:
to @tylerecouture so it has nothing to do with django, nor python at all. fine I will fix it right in editor source code

tylerecouture:
> to @tylerecouture so it has nothing to do with django, nor python at all. fine I will fix it right in editor source code

I'm sorry, I don't know! That's why we hired you =)

bashu:
to @tylerecouture just to clarify, you are fine with first scenario, it's  what you have expected:

variant01


but second one, should be fixed, plus avoid running script when loading data from db

variant02

tylerecouture:
> to @tylerecouture just to clarify, you are fine with first scenario, it's  what you have expected:
> 
> variant01
> 

Yes, for both the "Safe" widget and the "Unsafe" widget, this is good behaviour.


> 
> but second one, should be fixed, plus avoid running script when loading data from db
> 
> variant02
> 

For the "Safe" widget, yes, the script tags should be stripped and should never execute.

bashu:
to @tylerecouture how about now?

final

tylerecouture (changes requested):
The behaviour is perfect.  Can you advise: what is the most sustainable way of including your changes in the widget_common.html template?  Should we fork the django-summernote repo?  Or even better, can you make a PR to the summernote widget and see if they pick up the changes so we can continue to use the original library?  What do you suggest?

src/templates/django_summernote/widget_common.html
tylerecouture, line 1: Please be careful about violating open source licenses.  Attribution and license requirements need to be followed.

https://github.com/summernote/django-summernote/blob/main/LICENSE

tylerecouture (changes requested):
See notes above, also failing tests.

tylerecouture (changes requested):
One more thing to check on behaviour:  What happens when you save the form, and then go to edit the form again, does any additional processing occur?

bashu:
> Should we fork the django-summernote repo? Or should you make a PR to the django-summernote repo and see if they pick up the changes so we can continue to use the original library? What do you suggest?

fork django-summernote under your project/company account, name the fork as bytedeck-summernote or django-summernote-bytedeck. You will use this fork, until I manage pull my changes into official repo, accepting and merging it can take a time.

bashu:
> One more thing to check on behaviour: What happens when you save the form, and then go to edit the form again, does any additional processing occur?

Yes, all the tags, attributes and styles listed here https://github.com/summernote/django-summernote/blob/main/django_summernote/settings.py are passed unchanged, everything else (including script and style tags) are purified

bashu:
> See notes above, also failing tests.

Was too excited about fix, so forgot to update tests from previous iteration

src/announcements/forms.py
```diff
@@ -4,7 +4,8 @@
 from django.utils import timezone
 
 from bootstrap_datepicker_plus.widgets import DateTimePickerInput
-from django_summernote.widgets import SummernoteInplaceWidget
+
+from utilities.widgets import SummernoteInplaceSafeWidget as SummernoteInplaceWidget
```
tylerecouture, line 8: I would prefer to use the new class name to prevent confusion, as opposed to `as SummernoteInplaceWidget`.  It makes the code more clear and easier to read. Explicit is better than implicit =)

Since we will be having two different widget types (safe and unsafe) the code will not be clear which one a form is using just by reading the code, unless you know to check the import statements.

src/comments/forms.py
```diff
@@ -1,6 +1,6 @@
 from django import forms
 
-from django_summernote.widgets import SummernoteWidget
+from utilities.widgets import SummernoteWidget
```
tylerecouture, line 3: If we are extending the SummernoteWidget, lets give it a new name to prevent confusion.  How about `BDSummernoteWidget` (ByteDeck Summernote Widget)

src/hackerspace_online/settings.py
```diff
@@ -619,6 +619,9 @@
             'theme': 'monokai',
         },
 
+        # XSS protection for CodeView
+        'codeviewFilter': True,  # set this to true to filter entities.
```
tylerecouture, line 623: What's an "entity" ?

src/portfolios/forms.py
```diff
@@ -1,7 +1,7 @@
 from bootstrap_datepicker_plus.widgets import DatePickerInput
 from django import forms
-from django_summernote.widgets import SummernoteInplaceWidget
 
+from utilities.widgets import SummernoteInplaceSafeWidget as SummernoteInplaceWidget
```
tylerecouture, line 4: See previous comment

src/quest_manager/forms.py
```diff
@@ -8,11 +8,11 @@
 from crispy_forms.bootstrap import Accordion, AccordionGroup
 from crispy_forms.helper import FormHelper
 from crispy_forms.layout import HTML, Div, Layout
-from django_summernote.widgets import SummernoteInplaceWidget
 from django_select2.forms import ModelSelect2MultipleWidget, ModelSelect2Widget
 from tags.forms import BootstrapTaggitSelect2Widget
 
 from utilities.fields import RestrictedFileFormField
+from utilities.widgets import SummernoteInplaceSafeWidget as SummernoteInplaceWidget
```
tylerecouture, line 15: See previous comment

src/utilities/forms.py
```diff
@@ -2,8 +2,8 @@
 from django.core.exceptions import ValidationError
 from django.contrib.flatpages.models import FlatPage
 from django.contrib.flatpages.forms import FlatpageForm
-from django_summernote.widgets import SummernoteInplaceWidget
 
+from utilities.widgets import SummernoteInplaceSafeWidget as SummernoteInplaceWidget
```
tylerecouture, line 6: See previous comment

src/utilities/widgets.py
```diff
@@ -1,11 +1,15 @@
...
+import html
 import operator
 from functools import reduce
 
 from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
 
+import bleach
+from bleach.css_sanitizer import CSSSanitizer
+from django_summernote.widgets import SummernoteWidget, SummernoteInplaceWidget as SummernoteWidgetBase  # noqa
```
tylerecouture, line 12: I'd prefer not to rename Classes unless it's a common standard.  It makes code harder to read.  Let's just stick with `SummernoteInplaceWidget`  Because developers will have to dig through less code to understand what it means.

src/utilities/widgets.py
```diff
@@ -191,3 +195,27 @@ class MyForm(forms.Form):
         if you just drop in the widget for a ContentObjectChoiceField field.
     """
     pass
+
+
+class SummernoteInplaceSafeWidget(SummernoteWidgetBase):
+
+    def value_from_datadict(self, data, files, name):
```
tylerecouture, line 202: This method will need a test and a docstring.

For docstring, please state the reason why the method is being overridden.  i.e. what problem is being solved that the original method does not handle.

src/templates/django_summernote/widget_common.html
tylerecouture: I think a code comment linking to the original is fine.  But also, to make it easier for us to keep this code up to date, clear comment indicators anywhere you've changed something, and comments WHY it was changed.  This will help future-proof the widget and prevent us from removing your changes by accident.

src/comments/forms.py
tylerecouture: This is confusing me.  I don't see a SummernoteWidget in utilities.widgets in this PR?  I think I'm getting confused by all the class renaming that's going on.

tylerecouture (changes requested):
I've added a more thorough code review.  Please mark each comment as resolved as you resolve them!

tylerecouture (changes requested):
Before I review, what happened to the safe widget/field?  Since we are going to have two widgets/fields, can you add the "safe" back into the names, so it is explicit which field/widget we are using?

For clarity, lets use this:
* "Safe": for the safe/stripping/escaping field/widget, e.g. `SummernoteInplaceSafeWidget`
* "Advanced": for the unsafe widget that does not modify code entered into the HTML editor view, e.g. 'SummernoteInplaceAdvancedWidget' (whcih I know you haven't started on yet)
* Common: for base classes or common stuff used by both versions

And then use these terms throughout where it makes sense.

I'm a bit surprised you seemed to overhaul everything since your last PR.  I was just expecting some renaming.  If you need clarification or any of my requests don't make sense, please ask!

It's also helpful for you to comment on my review notes, so I know you've considered them and so I don't have to waste time digging to figure out whether the note was actually resolved or not. Reviewing takes a lot of time, and the more info you can provide the easier and faster I can complete a review.

tylerecouture:
> fork django-summernote under your project/company account, name the fork as bytedeck-summernote or django-summernote-bytedeck

Is this still necessary?  It seems like this is not required any more.

src/quest_manager/forms.py
```diff
@@ -81,9 +81,9 @@ class Meta:
         }
 
         widgets = {
-            'instructions': SummernoteInplaceWidget(),
-            'submission_details': SummernoteInplaceWidget(),
-            'instructor_notes': SummernoteInplaceWidget(),
+            'instructions': ByteDeckSummernoteInplaceWidget(),
+            'submission_details': ByteDeckSummernoteInplaceWidget(),
+            'instructor_notes': ByteDeckSummernoteInplaceWidget(),
```
tylerecouture, line 86: These three fields will use the the "advanced" (unsafe) widget, once you get there.

src/bytedeck_summernote/templates/bytedeck_summernote/widget_common.html
tylerecouture, line 1: What has been changed here from the original?  Can you clearly indicate what you've changed from the parent repo?  For all of these templates, can you wrap your changes in: comment tags somehow?

maybe:
{% comment %} Start changes from source at https://github.com/summernote/django-summernote/blob/main/django_summernote/templates/django_summernote/widget_common.html 

Changes becuase: REASONS
{% endcomment %}
Moded code
{% comment %} End changes {% endcomment %}

Alternately, we create our own fork of the repo so that it remains connected to the upstream. I'm starting to think that would be a better approach.

tylerecouture (changes requested):
This is looking great. What happened to the safe widget/field?  Since we are going to have two widgets/fields, can you add the "safe" back into the names, so it is explicit which field/widget we are using?

For clarity, lets use this:
* "Safe": for the safe/stripping/escaping field/widget, e.g. `SummernoteInplaceSafeWidget`
* "Advanced": for the unsafe widget that does not modify code entered into the HTML editor view, e.g. 'SummernoteInplaceAdvancedWidget' (whcih I know you haven't started on yet)
* Common: for base classes or common stuff used by both versions

And then use these terms throughout where it makes sense.

src/hackerspace_online/settings.py
tylerecouture: How was this resolved?

bashu:
> "Safe": for the safe/stripping/escaping field/widget, e.g. SummernoteInplaceSafeWidget
"Advanced": for the unsafe widget that does not modify code entered into the HTML editor view, e.g. 'SummernoteInplaceAdvancedWidget' (whcih I know you haven't started on yet)

widget can be either inplace (iframe=False) or not (iframe=True). what about "safe" and "advanced" ? should both of them be inplace, or be inside HTML iframe or should I left the option for end user how to use them ?

tylerecouture:
> widget can be either inplace (iframe=False) or not (iframe=True). what about "safe" and "advanced" ? should both of them be inplace, or be inside HTML iframe or should I left the option for end user how to use them ?

I don't think the iframes work, see the github issue #1302 for the problems we are having with summernote iframe widget.

Leaving the option to the end user, which widget to use would be ideal.  However currently we are using the inplace widget because of the problems with the iframe widget.

tylerecouture:
@bashu with each update, can you provide a description of what you did to resolve each of my comments, under my comment?  A thumbs up acknowlegment is good, but I can't review a thumbs up, haha...  As well, an overall description of the changes you've made with each new push or update would help immensly.  This is too hard to review otherwise, because you force pushed instead of adding commits, it's not easy for me to see what's changed since the last review.  This complicates code review a lot and I can't just drop in to quickly review, I have to set aside a big block of time, which further delays the review.

Here's an example, I have no idea how to interpret this. Was it resolved? How was it resolved? Was it removed? Did you add a comment to explain it? etc.
image

tylerecouture:
Here's an option thatight make things easier:

1. Close this PR and start a fresh one.
2. Provide a full description that's up to date with the current status of the code changes.

PR #1328, opened 2023-04-10
feature: implement Safe & Advanced SummernoteWidget(s)

bashu:
### What?

I've implemented `ByteDeckSummernoteSafeWidget` and `ByteDeckSummernoteAdvancedWidget` classes as substitute for widgets from `django-summernote` package.

### Why?

See https://github.com/bytedeck/bytedeck/issues/1302

### How?

append `bytedeck_summernote` right after `django_summernote` in INSTALLED_APPS
```python
INSTALLED_APPS = [
...
'django_summernote',
'bytedeck_summernote',
]
```

update urls.py, replace `django_summernote.urls` with `bytedeck_summernote.urls`:

```python
urlpatterns += [
...
    # bytedeck summernote
    url(r'^summernote/', include('bytedeck_summernote.urls')),
]
```

Replace imports:
```python
from django_summernote.widgets import SummernoteInplaceWidget, SummernoteWidget

# and
from django_summernote.admin import SummernoteModelAdmin, SummernoteInlineModelAdmin
```
with

```python
from bytedeck_summernote.widgets import ByteDeckSummernoteSafeWidget, ByteDeckSummernoteSafeInplaceWidget

# and
from bytedeck_summernote.admin import ByteDeckSummernoteSafeModelAdmin, ByteDeckSummernoteSafeInlineModelAdmin

```

### Testing?
Yes

final

### Anything Else?

Q: Why there are a lot of templates?
A: To avoid clashing with django_summernote stock templates

Q: What was changed there?
A: Either "include" or "extends" tags, and two lines of code in "widget_common.html" file

Q: Why did you copy-pasted `render` methods of widget classes?
A: Because original developers hardcoded template names, instead of using easy to override "template_name" attribute.

Q: Why all classes have prefix "ByteDeck" ?
A: To avoid confusion with stock `django_summernote`, but please suggest better prefixes.


tylerecouture (changes requested):
This is looking great.  I will test out a live version of this when I get a chance.  For now a question for you. My biggest concern is the sustainability of the code.  What happens when django-summernote updates?  What is the workflow for keeping our hacked version up to date?  You've got the changes well documented now, so perhaps that's all we need.

Would it make more sense for use to move your work into a fork of the actual Summernote widget? It seems like that would be easier to maintain, since the code will maintain it's connection via git to the upstream repo.  Then we could also suggest PRs to the original as well, reducing our delta from the upstream.  However I'm not sure how we install a fork in the app?

I'm asking for your professional advice on this.

src/quest_manager/forms.py
```diff
@@ -81,9 +81,9 @@ class Meta:
         }
 
         widgets = {
-            'instructions': SummernoteInplaceWidget(),
-            'submission_details': SummernoteInplaceWidget(),
-            'instructor_notes': SummernoteInplaceWidget(),
+            'instructions': ByteDeckSummernoteSafeWidget(),
+            'submission_details': ByteDeckSummernoteSafeWidget(),
```
tylerecouture, line 85: `instructions` and `submission_details` fields should both use the AdvancedWidget

src/bytedeck_summernote/widgets.py
```diff
@@ -0,0 +1,93 @@
...
+from django.template.loader import render_to_string
+from django.urls import reverse
+from django.utils.safestring import mark_safe
+
+from bleach.css_sanitizer import CSSSanitizer
+from django_summernote.utils import get_config
+from django_summernote.widgets import SummernoteInplaceWidget, SummernoteWidget
+
+
+class ByteDeckSummernoteAdvancedWidget(SummernoteWidget):
```
tylerecouture, line 16: Docstring needed for this class describing the purpose of the AdvancedWidget (what does it add/change from it's base class?)

src/bytedeck_summernote/widgets.py
```diff
@@ -0,0 +1,93 @@
...
+            "width": summernote_settings["width"],
+            "height": summernote_settings["height"],
+        }
+
+        # using customized template here...
+        html += render_to_string("bytedeck_summernote/widget_advanced.html", context)
+        return mark_safe(html)
+
+
+class ByteDeckSummernoteSafeWidget(SummernoteInplaceWidget):
```
tylerecouture, line 47: Docstring needed for this class describing the purpose of the SafeWidget (what does it add/change from it's base class?)

src/utilities/forms.py
```diff
@@ -140,7 +141,7 @@ class Meta:
         exclude = ('enable_comments', 'template_name',)
 
         widgets = {
-            'content': SummernoteInplaceWidget(),
+            'content': ByteDeckSummernoteSafeWidget(),
```
tylerecouture, line 144: This can use AdvancedWidget

bashu:
> This is looking great. I will test out a live version of this when I get a chance. For now a question for you. My biggest concern is the sustainability of the code. What happens when django-summernote updates? What is the workflow for keeping our hacked version up to date? You've got the changes well documented now, so perhaps that's all we need.

See my latest commits, I tried to resolve it. In a near future all you need to is:

a) if this one got fixed https://github.com/summernote/django-summernote/issues/465, then replace its content of `bytecode_summernote/widget_commont.html` with:
```html+django
{% extends "django_common/widget_common.html" %}
```

b) pin newer version in `requirements.txt` and until there are no backward incompatibilities in django-summernote nor with django itself, thats all you need.

> Would it make more sense for use to move your work into a fork of the actual Summernote widget? It seems like that would be easier to maintain, since the code will maintain it's connection via git to the upstream repo. Then we could also suggest PRs to the original as well, reducing our delta from the upstream. However I'm not sure how we install a fork in the app?

with my latest patches, there are no benefits to maintain separate fork. However if we manage to accept some of our changes (templates and more generic implementation of admin+widgets classes) into upstream version of `django_summernote`, then we can reduce "bytedeck's" specific implementatin to three-five lines of code.

bashu:
to @tylerecouture and @yujinyuz once again tests fail, and once again it's unrelated to this Pull Request.

tylerecouture:
> TODO: What is the workflow for keeping our hacked version up to date? 
>  
> a) if this one got fixed [summernote/django-summernote#465](https://github.com/summernote/django-summernote/issues/465), then replace the content of `bytecode_summernote/widget_common.html` file with:
> 
> ```
> {% extends "django_common/widget_common.html" %}
> ```
> 
> yes, one line of code instead of hundred and half
> 
> b) pin newer version in `requirements.txt` and until there are no backward incompatibilities in django-summernote nor in django itself, that's all you need.
> 
> with my latest patches, there are no benefits to maintain separate fork. However if we manage to accept some of our changes (templates and more generic implementation of admin+widgets classes) into upstream version of `django_summernote`, then we can reduce "bytedeck" specific implementation to three-five lines of code.

Excellent.  Can you paste the above content into one of the docstrings or a block comment at the top?  Whatever location you think is most obvious.

bashu:
> @bashu This looks related? All three test failures have this same error:

this happened after I triggered "github actions" with forced push, after my last comment. Yeah they are related, but again in happens only on github, works normally on my machine, to @yujinyuz any ideas how to solve it?

bashu:
to @tylerecouture how about README.md, will be automatically rendered by github as index.html when you navigate the folder, see

Screenshot 2023-04-11 at 6 48 57 PM

src/portfolios/forms.py
```diff
@@ -1,16 +1,18 @@
...
+from .models import Portfolio, Artwork
 
 
 class PortfolioForm(forms.ModelForm):
     class Meta:
         model = Portfolio
         fields = ['description', 'listed_locally', 'listed_publicly', ]
         widgets = {
-            'description': SummernoteInplaceWidget(),
+            'description': ByteDeckSummernoteSafeWidget(),
```
tylerecouture, line 15: ByteDeckSummernoteSafeInplaceWidget?

tylerecouture (changes requested):
I'm not sure if those are why the tests are failing, or maybe you're missing a migration?

tylerecouture:
> > @bashu This looks related? All three test failures have this same error:
> 
> this happened after I triggered "github actions" with forced push, after my last comment. Yeah they are related, but again it happens only on github, works normally on my machine, to @yujinyuz any ideas how to solve it?

Try it with an empty database, that's the only difference between github actions and local testing.  Last feature, you were trying to access the db when the app loads, and I suspect it's the same issue here.  You need to lazy load things or access after migrations.

src/announcements/forms.py
```diff
@@ -26,7 +27,7 @@ class Meta:
         # > If you don't like <iframe>, then use inplace widget
         # > Or if you're using django-crispy-forms, please use this.
         widgets = {
-            'content': SummernoteInplaceWidget(),
+            'content': ByteDeckSummernoteSafeWidget(),
```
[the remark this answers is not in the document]
bashu: was it fixed by later commits, no?

src/portfolios/forms.py
bashu: see above

tylerecouture: This is current code on github,

bashu: fixed

bashu:
to @tylerecouture the problem with test in "notifications" module,

if you run this:
```
docker-compose run web bash -c "python ./src/manage.py test src/bytedeck_summernote src/notifications --parallel --failfast"
```
tests fill fail, but if you test in different order like this:
```
docker-compose run web bash -c "python ./src/manage.py test src/notifications src/bytedeck_summernote --parallel --failfast"
```

everything is working

bashu:
> Can you run tests locally with an empty database and ensure you aren't trying to access the db before migrations run? That was a problem with the GFK widget at one point

this thing again:
```python
FAIL: test_active (quest_manager.tests.test_models.QuestTestModel)
The active method of the Quest model's parent "XPItem" should return the correct values based on a quest object's settings.
  File "/app/src/quest_manager/tests/test_models.py", line 161, in test_active
AssertionError: True != False

Ran 645 tests in 1330.235s

FAILED (failures=1)
Destroying test database for alias 'default'...
```

tylerecouture (changes requested):
Everything is looking great!  You just have one method that isn't tested.  Can you cover that and then we should be done!!!

image

tylerecouture:
Do you know why scripts run twice when you load an form in Django Admin?

For example, put `<script>alert("hello")</script>` in an advanced widget in Django Admin, save, then load the form.  You get two popups.

bashu:
> Everything is looking great! You just have one method that isn't tested. Can you cover that and then we should be done!!!

I see it already fixed by @TimberlineHackerspace

src/bytedeck_summernote/templates/bytedeck_summernote/widget_iframe.html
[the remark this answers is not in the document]
bashu: please look into thin PR https://github.com/bytedeck/bytedeck/pull/1335 it answers your question

bashu:
> Do you know why scripts run twice when you load a form in Django Admin?
> 
> For example, put `<script>alert("hello")</script>` in an advanced widget in Django Admin, save, then load the form. You get two popups.

Can confirm, will fix it https://github.com/bytedeck/bytedeck/issues/1336

tylerecouture:
> > Everything is looking great! You just have one method that isn't tested. Can you cover that and then we should be done!!!
> 
> I see it already fixed by @TimberlineHackerspace

That is me from a different computer =)

Issue #1334, opened 2023-04-13
Add indicators to "Advanced" Summernote widgets


tylerecouture:
We need a visual indicator for "Advanced" summernote widgets:

- [x] Make the codeview icon red or yellow (or otherwise different)
- [x] Change the the hover-text to this: "Advanced Code View: scripts will run; with great power comes great responsibility!"

bashu:
can not assign it to myself, but working on it

tylerecouture:
Closed by 1338

PR #1338, opened 2023-04-14
feature: add indicators to "Advanced" Summernote widgets

bashu:
### What?

- [x] Make the codeview icon red or yellow (or otherwise different)
- [x] Change the the hover-text to this: "Advanced Code View: scripts will run; with great power comes great responsibility!"

### Why?

As requested in #1334 

### How?

Using customized CSS files:
```python
SUMMERNOTE_CONFIG = {
    ...
    # You can add custom css/js for SummernoteWidget.
    'css': (
        'path/to/custom_summernote_iframe.css',
    ),

    # You can also add custom css/js for SummernoteInplaceWidget.
    # !!! Be sure to put {{ form.media }} in template before initiate summernote.
    'css_for_inplace': (
        'path/to/custom_summernote_widget.css',
    ),
}
```

### Testing?

Nope

Screenshot 2023-04-14 at 8 12 57 PM

### Anything Else?

Q: How to change colours?
A: Edit/override `custom_summernote_iframe.css` and/or `custom_summernote_widget.css` for inplace variant.

Q: How to change tooltips?
A: Summernote WYSIWYG editor stores text for tooltips in its own language file, edit `bytedeck_summernote-advanced-<language code>.min.js` file.

### Review request
@tylerecouture


bashu:
> @bashu does this include the hover text?

No it doesn't, it will be included in a next commit. 

> Also question: is there a reason not to just add a simple style rule in our existing stylesheet (custom.css), rather than customizing the widget itself so much?

No reason, it's just an example. I can move everything into custom.css if you insist

tylerecouture:
> No reason, it's just an example. I can move everything into custom.css if you insist

If it is more simple than overriding, then yes.

Can you change it to be highlighted on :hover instead of all the time?  I realize looking at the screenshot its annoying and most users won't care, haha.

bashu:
to @tylerecouture see it in action:

tooltip

tylerecouture:
> to @tylerecouture see it in action:
> 
That looks sweet!!

src/bytedeck_summernote/static/bytedeck_summernote/lang/bytedeck_summernote-advanced-en-US.min.js
tylerecouture, line 1: What's the purpose of including this file locally? Why aren't we using the default and modifying with css or jquery?

bashu: it contains the translations, ie. the line "Advanced Code View..." is here. You can't modify localization with CSS, as for JS, they (developers of summernote) choose to use this format, for example:
Screenshot 2023-04-17 at 9 39 24 AM

PR #1342, opened 2023-04-21
fix: safe widget is stripping some html tags (closes #1340)

### LINKED ISSUE #1340: Summernote safe widget is stripping/escaping some html tags entered in codeview
Summernote safe widget should only strip script tags.  Currently it's also stripping many html tags when entered in code view.

Safewidget should not strip or escape any html entered into CodeView, only script tags.

To reproduce:
1. Run the dev environment and got to a tenant site (e.g. hackerspace.localhost:8000) and log in as admin/password
2. Go to the Announcements tab, and click New (hackerspace.localhost:8000/announcements/create/)
3. Give a title
4. Enter some content in the Content field that uses the SummernoteSafeWidget.
5. Style some html content:
    - using various style buttons 
    - Embed a video using the "Embed Video" button
    - type html directly into codeview, e.g. <small> tags
8. Note how toggling in and out of Codeview works as expected (no HTML is stripped or escaped)
9. Save the Announcement
10. Click the title of the new announcement to expand the accordian and view content: observe how content is escaped or stripped.

Some examples:
ezgif com-optimize

MoreEscaping
SafeEscaping

bashu:
### What?

Fixes issue with so-called "safe" widget is stripping/escaping some html tags entered in codeview 

### Why?

See https://github.com/bytedeck/bytedeck/issues/1340

Closes #1340 

### How?

It's a bug in upstream "django_summernote" package, but I fixed it in bytedeck's version, see diffs

### Testing?

Manually, yes.

makesummernotegreatagain

### Anything Else?

Q: WTF you just did?
A: Upstream "django_summernote" has an issue, they didn't updated their list of ALLOWED_TAGS and STYLES for years, it was out-of-sync with the version of SummerNote Editor they use. I fixed it locally.

### Review request
@tylerecouture

### DIFF
--- a/src/bytedeck_summernote/css_sanitizer.py
+++ b/src/bytedeck_summernote/css_sanitizer.py
@@ -1,0 +1,13 @@
+import html as htmllib
+
+from bleach import css_sanitizer
+
+
+class CSSSanitizer(css_sanitizer.CSSSanitizer):
+    """Override default `bleach.css_sanitizer.CSSSanitizer` class (mandatory for ByteDeck project)"""
+
+    def sanitize_css(self, style):
+        """
+        Override `sanitize_css` method, unescape styles before sanitization (fix #1340)
+        """
+        return super().sanitize_css(htmllib.unescape(style))
--- a/src/bytedeck_summernote/settings.py
+++ b/src/bytedeck_summernote/settings.py
@@ -1,0 +1,15 @@
+ALLOWED_TAGS = [
+    'a', 'div', 'p', 'span', 'img', 'em', 'i', 'li', 'ol', 'ul', 'strong', 'br',
+    'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
+    'table', 'tbody', 'thead', 'tr', 'td',
+    'abbr', 'acronym', 'b', 'blockquote', 'code', 'strike', 'u', 'sup', 'sub',
+]
+
+STYLES = [
+    'background-color', 'font-size', 'line-height', 'color', 'font-family'
+]
+
+ATTRIBUTES = {
+    '*': ['style', 'align', 'title', ],
+    'a': ['href', ],
+}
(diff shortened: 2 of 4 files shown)


bashu:
to @tylerecouture let me add some unittests...

bashu:
hey @tylerecouture you may asks, why do we need to override CSSSanitizer from bleach package? It's because how "font family" picker implemented in Summernote WYSIWYG editor, they insert "escaped" CSS values, which get stripped by stock "bleach.clean" method and you lose your fancy fonts after saving/loading.

tylerecouture:
> hey @tylerecouture you may ask, why do we need to override `CSSSanitizer` class from **bleach** package? It's because how "font family" picker implemented in Summernote WYSIWYG editor, it insert "escaped" CSS values which get stripped by stock `bleach.clean` method and you lose your fancy fonts after saving/loading.

This is not the case in the original summernote widget.  Wasn't bleach.clean method a package you introduced?

bashu:
> This is not the case in the original summernote widget. Wasn't bleach.clean method a package you introduced?

Except it was, https://github.com/summernote/django-summernote/blob/main/django_summernote/fields.py#L18 but from my understanding, original authors didn't give a damn about, and this feature was never supported  as you and I wanted to

tylerecouture:
@bashu is it possible to just use your AdvancedWidget, since it is working great, but strip out the script tags for the safe widget?

bashu:
> @bashu is it possible to just use your AdvancedWidget, since it is working great, but strip out the script tags for the safe widget?

Do you want me to scrap everything above?

tylerecouture:
> > @bashu is it possible to just use your AdvancedWidget, since it is working great, but strip out the script tags for the safe widget?
> 
> Do you want me to scrap everything above?

I need this as a Hotfix because production is currently broken.  If you are satisfied with this, then I'll take it.  But why is there a True/False discrepancy I commented about above?

bashu:
without "bleach" library, it will become a little non-trivial task. I think it was a reason why authors of django_summernote choose it, instead of implementing their own parser. 

And simple removal of all "script" occurencies from input, will no solve it.

PR #1360, opened 2023-05-13
feature: getting rid of "bleach" library

bashu:
### What?
Getting rid of "bleach" library
### Why?
Fixing previously made "over engineered" solution
### How?
Escape `script` tags and only it
### Testing?
Yes
No
### Anything Else?
This is alternative solution for https://github.com/bytedeck/bytedeck/pull/1359
### Review request
@tylerecouture


tylerecouture:
Thanks @bashu I'm going to leave the over engineered solution for now, buit may come back to this, so please keep your branch!

PR #1367, opened 2023-05-16
Fixed missing closing brace for pluralization

Enprogames:
### What?
This fixes a small issue which exposed front-end code. It had to do with adding an "s" to the end of a word if a quest was repeatable more than once e.g. "repeatable 1 time" vs "repeatable 6 times".
### Why?
This will add to the overall polish of the website. Front-end code should not be exposed like this.
### How?
To fix this, all I had to do was add a closing brace to the code in preview_content_quests_avail.html.
### Testing?
I set the number of allowed repeats for one of the default quests to ensure that this issue was fixed. There didn't appear to be any exposed front-end code.

This issue seems much to simple to write a dedicated unit test for.
image

### Anything Else?
In the future, there might be some kind of test we could add to prevent this? Maybe something that scans each Django template file to ensure it contains valid code?
### Review request
@tylerecouture

### DIFF
--- a/src/quest_manager/templates/quest_manager/preview_content_quests_avail.html
+++ b/src/quest_manager/templates/quest_manager/preview_content_quests_avail.html
@@ -17,7 +17,7 @@
       <li>Repeatable: {% if q.max_repeats == 0 %}no
              {% else %}
                {% if q.max_repeats == -1 %} Unlimited
-               {% else %} {{q.max_repeats}} time{{q.max_repeats|pluralize} max
+               {% else %} {{q.max_repeats}} time{{q.max_repeats|pluralize}} max
                {% endif %}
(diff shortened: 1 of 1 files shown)


tylerecouture (approved):
I like your suggestion for a django template linter/verifier.  I wonder if there's a pre-commit hook for that!

Enprogames:
> I like your suggestion for a django template linter/verifier. I wonder if there's a pre-commit hook for that!

I looked into that, and there doesn't seem to be anything like that.

There's one that verifies that the Django templates compile correctly without errors, but that wouldn't have helped us in this case. And there are other ones that verify only the HTML, but that still wouldn't have helped here. The only thing I can think is to write a script that uses regular expressions to ensure that any curly braces are closed correctly lol.

PR #1368, opened 2023-05-16
fix: alphabetically order quest maps in list view; Closes #1346

### LINKED ISSUE #1346: Quest Maps in Quest Map List Should Be Alphabetical
Would make this page look neater and when there are a lot of maps it is also easier to navigate

Dylan-RG-Patterson:
### What?
This is a change in ordering in map objects, which are now ordered by alphabet/name. Maps in the list view are now ordered as such, and because the change in ordering takes place at the model level, any additional archives outside of the list view will now also be displayed alphabetically.

### Why?
An alphabetical list is far more aesthetically pleasing, and it makes parsing larger lists for specific maps easier too.

### How?
The model class for map (cytoscape) objects has had its meta class ordering value set to 'name'. 

### Testing?
A test to assert that the default object list is ordered correctly has been created. It creates three new maps with names A, B and C and then compares it to the same list, but manually ordered by name.

image

### Anything Else?
Creating a new map with the same initial quest as another currently raises an error instead of a warning message or being disallowed. If there isn't an active issue regarding this, it would be good to create one.

### Review request
@tylerecouture


tylerecouture (approved):
@Dylan-RG-Patterson Please add an issue for the bug you described, and steps to reproduce.  Thank you!

PR #1376, opened 2023-05-30
Custom deck css and js, fix #1365

### LINKED ISSUE #1365: Custom deck css and js
Allow deck owners to upload a custom css file and a custom js file.

- These should be two new fields in the SiteConfig model, and should only be editable by the deck owner (similar to the deck owner field itself)
- This should be in a collapsible, "Advanced" section of the form along with the deck owner field (similar to how the Quest form that has a collapsible Advanced section)
- There should be some simple check on the field that the files or of the correct type (js and css, file extensions at least)
- The CSS should be inserted LAST (i.e. highest priority) in the css stack in `templates/head_css.html`.  For an example of similar behaviour, see this code in the same file:

```{% if config.get_favicon_url %}
      <link rel="icon" href="{{ config.get_favicon_url }}">
{% else %}
```
- The JS file should be loaded, if it exists, at the bottom of templates/javascript.html
- A simple test that provides a (blank?) css file to the test tenant siteconfig and verifies it is included in the html returned from the a view
- A simple test that provides a (blank?) js file to the test tenant siteconfig and verifies it is included in the html returned from a view

bashu:
### What?
Allow deck owners to upload a custom css file and a custom js file.

### Why?

Closes #1365 

### How?

adding custom model fields and extending form / validation

### Testing?

Yes

Screenshot 2023-05-24 at 8 35 06 PM

### Anything Else?

Q: What you did in "clean_custom_stylesheet" method?
A: Used "cssutils" library for simple (stupid) content validation, see relevant tests.

Q: Why do we need this method?
A: You don't, but you may like it, I can remove / disable it.

### Review request
@tylerecouture


bashu:
to @tylerecouture, here is an issue:

Helper function `generate_form_data` from `hackerspace_online.tests.utils` module doesn't take into account custom form layout created with `FormHelper` from `crispy` library,

So if you forgot to list new field in `form.helper.layout` list, it wouldn't be rendered by `{% crispy form %}` template tag, and in a case of `custom_name_for_student` which was *required* (non-blank) field, the `SiteConfigForm` will fails to submit, and "crispy" can't render errors for an excluded (even they are *requided*) fields.

So the problem has nothing to do with "custom_stylesheet" file submission at all, the problem with "FormHelper". Why do we need it? You have requested to put "custom_stylesheet", "custom_javascript" and "deck_owner" fields into "accordion" thats why.

Here how you can reproduce it yourself:

1) go to https://github.com/bytedeck/bytedeck/blob/develop/src/quest_manager/forms.py#L126
2) comment out some required fields, like "xp" and "name"
3) navigate to http://hackerspace.localhost:8000/quests/ and try to update any quests, nothing will happens
4) now try to run tests, they will succeed without any exceptions (function `generate_from_data` doesn't take into account that "form.helper.layout" has its own list of "included" form fields). yes tests from `quest_manager.test.test_forms" doesn't work and doesn't catch failures too.

tylerecouture:
> to @tylerecouture, here is an issue:

Thank you for the explanation, excellent detail. This makes sense.

So what we are left with is, we just need a generic test that the form can be successfully posted with a valid payload (redirects to the detail view as opposed to back to the form view).  It won't be super specific, but at least that would let us know that something is missing somewhere?  That should catch and prevent new bugs in this form.

Do you think this would work?

tylerecouture (changes requested):
Just needs a generic test with full payload to ensure the form can be successfully posted.  This will prevent the bugs in previous PRs from slipping through the cracks.

bashu:
to @tylerecouture please check two last commits, I hope it's what you wanted

Issue #1378, opened 2023-05-31
Implement bootstrap-table Filter Control on Available Quests List


tylerecouture:
Example and code:
https://examples.bootstrap-table.com/#extensions/filter-control.html

Docs:
https://bootstrap-table.com/docs/extensions/filter-control/

Quest: Search (and remove the main search widget)
Campaign: Filter
Tags: Filter (but take into account comma seperated values with filterDataCollector or filterControlMultipleSearchDelimiter?)

tylerecouture:
No longer required.

Issue #1369, opened 2023-05-17
Refactor quest list to use bootstrap-table


tylerecouture:
Try to mimic existing quest list as close as possible, but utilizing [bootstrap-table](https://examples.bootstrap-table.com/index.html?bootstrap3) so that the list can be sorted, filtered, etc.

For some examples of standard bootstrap-table usage, see the Student list at profiles_manager/profile_list.html and the tag list at tags/list.html

The difficult part will be the accordion, i.e. collapsing each row.
- Here's a proof of concept I made a long time ago, that could be used for ideas: https://jsfiddle.net/mpr63ezm/40/
- Some more discussion here: https://github.com/wenzhixin/bootstrap-table/issues/964

To start, just added the default search (on name) and sorting (on all columns).  We'll be adding filters in a new issue.

I recommend creating a duplicate quest list view for this with it's own url we can access manually.  Once the feature is complete we can swap it out for the main quest list.

Enprogames:
I just saw that you've assigned this to me. I will get started working on it right away.

Enprogames:
I'm very new to JavaScript, but I think I understand the JSFiddle you provided. I created a new one that more closely mirrors the behaviour of the hackerspace website: https://jsfiddle.net/enprogames/cv8f2Lgh/10/

I've also spent quite a bit of time diving into the Django templates that currently exist for this, and I'm a bit confused about them. My current understanding of the `quest_manager` templates that relate to this functionality goes as follows:
- quests.html: Main page for displaying the quests page. Shows the "Available", "In Progress", "Completed", and "Draft" tabs depending on who is using the site, and uses each of the "tab_*" templates to display a list of quests depending on which tab is currently selected.
- quests2.html: A previous attempt at redoing the quests.html page? It doesn't seem to be using bootstrap-tables though.
- tab_quests_approvals.html: Displays quests awaiting approval, allowing staff or TAs to review and approve quests?
- tab_quests_available.html: Displays the quests that are currently available for a user to undertake.
- tab_quests_available2.html: An attempt at improving tab_quests_available.html?
- tab_quests_submission.html: Displays quests that are currently in progress or have been completed by the user?

It seems like each of these tabs do essentially the same thing, so it looks like to complete this issue, all of these will need to be refactored to use bootstrap-table? Also, maybe we should create an additional issue to consolidate these views in the future?

tylerecouture:
[...]

PR #1389, opened 2023-06-05
Refactor all quest lists

Enprogames:
### What?
This is a full attempt at refactoring all of the quest lists 

### Why?
Once this is working properly, it will allow more dynamic actions to be performed on the lists of quests, such as sorting, filtering, etc. See #1369.

### How?
I updated all of the quest list views. This required understanding bootstrap tables, writing new JavaScript code, and wrestling with a combination of custom CSS styles and bootstrap ones. 

To get to this point, I had to work through several issues:
- Some old, experimental stylesheets and templates were removed, including:
    - `src/static/js/bootstrap-table-accordion.js`
    - `src/static/css/bootstrap-table-accordion.css`
    - `src/announcements/templates/announcements/list2.html`
    - `src/quest_manager/templates/quest_manager/flagged.html`

- Removing small gap at bottom of an expanded row (in some cases). This seems to be due to some style in `src/static/bootstrap-table-accordion.css`, applying to list items.
     small gap image
<del>
 -  It took a long time to figure out what was causing this, but adding the following CSS seemed to fix it:
</del>

<del>
- I was able to fix the small gap appearing by changing the max-height of .list-group-item-buttons elements from 55px to 65px:
</del>

- We ended getting rid of the small gap by changing the margin-bottom of `.list-group-item-buttons > .pull-right` to 10px, and the max-height of .list-group-item-buttons elements from 55px to 60px:
```css
/*
The .list-group-item-buttons rule sets a minimum height of 65px.
This is necessary to eliminate a small gap that appears between an expanded
bootstrap-table accordion row and the subsequent row. This gap is caused by a
15px margin-bottom on the .pull-right child element of .list-group-item-buttons.
*/
.list-group-item-buttons {
    min-height: 60px;
}

.list-group-item-buttons > .pull-right {
    margin-left: 15px;
    margin-bottom: 10px;
}
```

- Styling not showing up for "danger", "warning", "muted", and "info" rows.
    - <del>I tried using the bootstrap-table contextual classes, but these wouldn't work on my table. It seems that these only exist in later versions of bootstrap. I also found that "row-danger", "row-warning", etc classes were already created in the `src/static/bootstrap-table-accordian.css stylesheet`, but these didn't work either. I ended up making my own "table-warning", "table-danger", etc classes, and this worked well.</del>
    - I was able to use `danger`, `warning`, `info`, etc to style the rows. For the expanded rows, despite my best efforts, I was unable to find bootstrap classes that could effectively style them (except for with muted, `panel-muted` worked). 
<details>

  <summary>Table Row Styling Light Mode</summary>

  table row styling light mode

</details>

<details>

  <summary>Table Row Styling Dark Mode</summary>

  table row styling dark mode

</details>

- The black bar mentioned [here](https://github.com/bytedeck/bytedeck/issues/1369#issuecomment-1569367199) has been removed.
- Tags now appear on both the `tab_quests_available.html` and `tab_quests_submission.html` templates.

❗Note: All of the old views were made accessible by adding `/old` to the page, ❗**but not the category_detail_content.html view**

### Testing?
Views on each of the following pages were refactored:
- tab_quests_available.html
    - http://hackerspace.localhost:8000/quests/available/
- tab_quests_submission.html
    - http://hackerspace.localhost:8000/quests/inprogress/
    - http://hackerspace.localhost:8000/quests/completed/
- tab_quests_approvals.html
    - http://hackerspace.localhost:8000/quests/approvals/
    - http://hackerspace.localhost:8000/quests/approvals/returned/
    - http://hackerspace.localhost:8000/quests/approvals/approved/1/
    - http://hackerspace.localhost:8000/quests/approvals/flagged/
- category_detail_content.html
    - `http://hackerspace.localhost:8000/quests/campaigns/<pk>/` e.g. http://hackerspace.localhost:8000/quests/campaigns/1/

<details>

<summary>And `/old/` versions were created for each page (except for `category_detail_content.html`)</summary>

- tab_quests_available_old.html
    - http://hackerspace.localhost:8000/quests/available/old/
- tab_quests_submission_old.html
    - http://hackerspace.localhost:8000/quests/inprogress/old/
    - http://hackerspace.localhost:8000/quests/completed/old/
- tab_quests_approvals_old.html
    - http://hackerspace.localhost:8000/quests/approvals/old/
    - http://hackerspace.localhost:8000/quests/approvals/returned/old/
    - http://hackerspace.localhost:8000/quests/approvals/approved/old/
    - http://hackerspace.localhost:8000/quests/approvals/flagged/old/

</details>

<details>

<summary>Example: Expanded row, tags, no black bar</summary>

image

</details>

<details>

<summary>Table Row Styling Light Mode</summary>

table row styling light mode

</details>

<details>

<summary>Table Row Styling Dark Mode</summary>

  table row styling dark mode

</details>

### Anything Else?
- ❗ Since this PR is already huge, I will cover #1378 in my next PR (if we decide to implement it).

### Review Request
@tylerecouture


src/static/css/bootstrap-table-accordion.css
```diff
@@ -1,123 +1,310 @@
...
+
+  /* Removing border from the table header */
+  .fixed-table-container,
+  .bootstrap-table .table > thead > tr > th,
+  .table > thead > tr > th,
+  .table > tbody > tr > th {
+    border: none;
+  }
+
+  /* Removing spacing between expanded accordion rows for a cleaner flow */
```
tylerecouture, line 56: I don't think this extra CSS is necessary.  The gap is caused by the buttons have a bottom margin of 15px (caused by `.list-group-item-buttons > .pull-right` in custom_common.css), but the containing element only having a bottom padding of 10px (caused by .list-group-item class and provided by bootstrap).

So if the content is only a single line of text, the margin will push 5px past the padding, causing the gap.  If there is more than one line of text, then the container will be large enough to contain the button margin

I think a more simple fix is to just add 5px to the minimum-height of the containing element, OR reduce the margin of the buttons by 5px.  The margin is important to give a gap below the buttons in case there is lots of text that will flow below the button. Both of these styles are in our custom_common.css.

I think the better solution is to set margin-bottom: 10 px; on the buttons, via classes shown here:

image

src/static/css/bootstrap-table-accordion.css
```diff
@@ -1,123 +1,310 @@
...
-*
-************************************/
-
-/* restore thicker bootstrap header bottom border removed in bootstrap-table */
+/*=============================================================
+=            Bootstrap Table Accordion Styling Fixes          =
+=         https://github.com/wenzhixin/bootstrap-table        =
+
+=  In Bootstrap 3, many important styling rules for the
+=  tables are missing, such as danger, warning, info etc.
```
tylerecouture, line 6: I don't think this is the case.  For example, the danger styling works fine in the existing profile list view that uses bootstrap-table.  Better to use existing styles if possible to remove unnecessary code and styling (everything we customize makes maintaining the project a little bit harder)

image


The color is coming from default bootstrap css:

image

src/quest_manager/templates/quest_manager/tab_quests_submission.html
```diff
@@ -8,81 +8,91 @@
...
+          <img class="img-responsive panel-title-img img-rounded" src="{{ s.quest.get_icon_url }}" alt="icon"/>
+        </td>
+        <td class="col-sm-5 col-xs-8">{{s.quest_name}}</td>
+        <td class="col-sm-1 col-xs-2 text-center">
+          {% if s.quest.xp_can_be_entered_by_students %}{{s.xp_requested}}*{% else %}{{s.quest.xp}}{% endif %}
+        </td>
+        <td class="col-sm-2 hidden-xs"><small>
+          {% if s.quest.campaign %}{{ s.quest.campaign }} {% else %}-{% endif %}
+        </small></td>
+        <td class="col-sm-1 hidden-xs text-center"><small>{{s.quest.tags.all|join:", "}}</small></td>
```
tylerecouture, line 51: Can you add a default of "-" to tags, if their are none? This adds a nice visual indicator, similar to how Campaigns display, rather than whitespace.

Also, there is a cleaner way to do this, than what I did with campaigns: 
`{% if s.quest.campaign %}{{ s.quest.campaign }} {% else %}-{% endif %}`

I think can just be done like this:
`{{ s.quest.campaign|default:"-" }}

I think you can chain filters, since you're already using join, but I'm not sure which order you'll need to out them in.

Please update the campaign code too while you're in here!


image

tylerecouture (changes requested):
I haven't had a chance to look at everything yet, but I have a few comments for you to consider for now.

src/quest_manager/templates/quest_manager/tab_quests_submission.html
Enprogames: Done

src/static/css/bootstrap-table-accordion.css
Enprogames: You were right. Bootstrap had `danger`, `warning`, `info`, and `success` classes for table rows. It looks like it's missing muted classes, but `panel-muted` ended up working perfectly. Not sure why I can't get `panel-warning`, `panel-danger`, etc to work.

I also couldn't figure out a way to have the danger/warning/info border around an expanded row, so I kept the styles for that:
```css
  /* Updating border-color based on class */
  /* Specify the colors for each class */
  .accordian-table > tbody > tr {
    --accordian-table-border-color: transparent;
  }

  .accordian-table > tbody > tr > .default {
    /* make what it would be normally */
    --accordian-table-border-color: #337ab7;
  }

  .accordian-table > tbody > tr > .muted {
    --accordian-table-border-color: #333;
  }

  .accordian-table > tbody > tr > .warning {
    --accordian-table-border-color: #8a6d3b;
  }

  .accordian-table > tbody > tr > .danger {
    --accordian-table-border-color: #a94442;
  }

  .accordian-table > tbody > tr > .info {
    --accordian-table-border-color: #31708f;
  }


  /* Updating list-group-item border-color to match the row */
  tr.detail-view > td > div > ul > .list-group-item,
  tr.detail-view > td > div > .list-group-item {
    border-left-color: var(--accordian-table-border-color);
    border-right-color: var(--accordian-table-border-color);
  }

  tr.detail-view > td > div > ul > .list-group-item:last-child,
  tr.detail-view > td > div > .list-group-item:last-child  {
    border-bottom-color: var(--accordian-table-border-color);
  }
  ```

File: src/static/css/bootstrap-table-accordion.css (hunk 1 above)
Comment (Enprogames) [reply src/static/css/bootstrap-table-accordion.css]: You were right! I tried changing the min-height on a few other elements, but I never got it to work. But changing it on that `.list-group-item-buttons` class seemed to do the trick.

```css
/* custom_common.css */

/*
The .list-group-item-buttons rule sets a minimum height of 65px.
This is necessary to eliminate a small gap that appears between an expanded
bootstrap-table accordion row and the subsequent row. This gap is caused by a
15px margin-bottom on the .pull-right child element of .list-group-item-buttons.
*/
.list-group-item-buttons {
    min-height: 65px;  /* used to be 55px */
}

.list-group-item-buttons > .pull-right {
    margin-left: 15px;
    margin-bottom: 15px;
}
```

This allowed me to remove the classes you mentioned.

src/static/css/custom_common.css
```diff
@@ -664,8 +664,14 @@ to overlap borders ****/
...
 
+/*
+The .list-group-item-buttons rule sets a minimum height of 65px.
+This is necessary to eliminate a small gap that appears between an expanded
+bootstrap-table accordion row and the subsequent row. This gap is caused by a
+15px margin-bottom on the .pull-right child element of .list-group-item-buttons.
+*/
 .list-group-item-buttons {
-    min-height: 55px;
+    min-height: 65px;
```
tylerecouture, line 674: As I mentioned in my comment:

> I think the better solution is to set margin-bottom: 10 px; on the buttons, via classes shown here:
> 
image


Setting the min height to 65, although it solves the gap problem just makes everything unnecessarily bigger than it needs to be and leaves a big gap below if there is only a small amount of text.

tylerecouture (changes requested):

PR #1399, opened 2023-06-12
Admin action on public tenant to send emails, closes #1386

### LINKED ISSUE #1386: Admin action on public tenant to send emails
e.g. https://stackoverflow.com/questions/18619577/using-django-admin-actions-to-send-bulk-emails.

Create an Tenant model admin action (public tenant only) that will send emails to selected tenant "owners"

- owner emails taken from the "owner" user (see SiteConfig.owner) for each deck.
- Field for email title
- use Summernote Safe Widget to draft email body
- email sent via celery task

bashu:
### What?

This PR is for feature requested in https://github.com/bytedeck/bytedeck/issues/1386 where website administrators can selectively compose and send email message to Deck owners via simple admin "action".

### Why?

This feature was requested in https://github.com/bytedeck/bytedeck/issues/1386, probably becaues there are no other way to communicate with deck owners and this is easisist solution.

### How?

A custom two-step admin action was implemented for registered `TenantAdmin` class, named [message_selected](https://github.com/bytedeck/bytedeck/pull/1399/files#diff-7e0a16a1951c3fc0fb90ab6c5b3ef0e7d299074ca491b8346c192e510d389a87R133) (see screenshots).

If you look at `Tenant` change list in the admin, you’ll see this feature in action; it's called “send an email message to all owners for selected tenant(s)” action.

Once you select objects and run the action, it will try to extract relevant recipient information, such as name and email address, from related `SiteConfig` (see `deck_owner` object) model.
  
**Note**: only verified email addresses will be selected, also if "full name" is missing, "username" will be selected.

On second (or intermediate) step you will compose the message (subject, body) for a list of selected recipients and once you finish the draft and confirm sending, it will be sent via async celery task: [send_email_message](https://github.com/bytedeck/bytedeck/pull/1399/files#diff-f3322bc520a62a8c028ebda7c6b36ab7028fb35da53801d2ce32241683a79368) 

**Note**: all the recipients will be listed in "Bcc" header when sending the email.

Message from intermediate step will be renderer as HTML using [tenant/templates/admin/tenant/email/message.txt](https://github.com/bytedeck/bytedeck/pull/1399/files#diff-1239c84a6797afd654a37cd7859e4d4d1bc79dfbb8e15e9792d7374d71f34b79), plain text copy will be created using `html2text` library. 
 
### Testing?

Test coverage for "message_selected" action of the admin views has been done, including testing of each separate features:

* celery task
* custom utilities, see `textify`

admin actions for selected objects:

Screenshot 2023-07-10 at 11 47 03 AM

compose form for selected recipients:

Screenshot 2023-07-10 at 12 22 17 PM

notification about successful sending, note that real "sending" is done by asynchronous task:

Screenshot 2023-07-10 at 12 22 35 PM

### Anything Else?

New dependency: `html2text` library.

`html2text` is a Python script that converts a page of HTML into clean, easy-to-read plain ASCII text. Better yet, that ASCII also happens to be valid Markdown (a text-to-HTML format).

[textify](https://github.com/bytedeck/bytedeck/pull/1399/files#diff-f543f37cf3b5d5a10f5f419eade37af775f2c486540b3c29094c9df82618c5b0) function uses `html2text` library to convert htmlized message, back to text message.

### Review request
@tylerecouture


bashu:
to @tylerecouture few questions, before finising this PR:

1) Do you want to send message as-is, or insert into template (with defined header, footer)?

2) Where to put `textify` function, maybe `utilities.text` ? Don't worry I will add tests. Why there such function `textify`?, I think it's nice to have "textual" version, for "htmlized" email message created in Summernote rich-text editor.

3) Where to put `send_html_mass_email` function, maybe `utilities.email` module?
What is that? That's a fork of builtin `django.core.mail.send_mass_mail` rewritten to use `EmailMultiAlternatives`. If you don't need it, I can simply use `send_email` in loop.
I will add test, if you decide to keep it.

4) I'm going to add tests, and also update template from intermediate step to match Grappelli style in next updates.

tylerecouture:
> 1. Do you want to send message as-is, or insert into template (with defined header, footer)?

Yes there should be a basic template with a signature block, but it doesn't need more than that. 

>     2. Where to put `textify` function, maybe `utilities.text` ? Don't worry I will add tests. Why there such function `textify`?, I think it's nice to have "textual" version, for "htmlized" email message created in Summernote rich-text editor.

Yes, utilities app seems fine. Put it in `utilities.email`

>     3. Where to put `send_html_mass_email` function, maybe `utilities.email` module?
>        What is that? That's a fork of builtin `django.core.mail.send_mass_mail` rewritten to use `EmailMultiAlternatives`. If you don't need it, I can simply use `send_email` in loop.
>        I will add test, if you decide to keep it.

Yes! Sounds great.
 
>     4. I'm going to add tests, and also update template from intermediate step to match Grappelli style in next updates.

Don't worry about styling this or making it look good, minimal code is easier to maintain, and this will only be used by admins, so I don't care if it's ugly.  Function is more important =)

bashu:
to @tylerecouture do you know how to test "admin" views with django_tenant, cuz I don't

bashu:
to @tylerecouture please review and test it,  next I will squash and rebase this PR

src/tenant/admin.py
```diff
@@ -115,6 +127,79 @@ def get_queryset(self, request):
...
 
+    @admin.action(description="Send an email message to all owners for the selected tenant(s)")
+    def send_email_message(self, request, queryset):
+        """
+        Send an email message to all owners for selected tenant(s).
+
+        This action first displays an intermediate page which shows compose message form and
+        a list of all recipients.
+
+        Next, it send email message to all selected users and redirects back to the change list.
```
tylerecouture, line 138: I appreciate the detailed docstring. Excellent.

src/tenant/admin.py
```diff
@@ -115,6 +127,79 @@ def get_queryset(self, request):
...
+                schema_name=get_public_schema_name())
+
+        # get a list of all owners (recipients) for selected tenant(s)
+        owners = []
+        for tenant in objects:
+            with tenant_context(tenant):
+                config = SiteConfig.get()
+                if config.deck_owner is not None:
+                    owners.append(config.deck_owner)
+        owners = set(owners)
```
tylerecouture, line 154: An explicit comment here would be helpful why you are using set() (to remove duplicates I assume?)

src/tenant/admin.py
```diff
@@ -115,6 +127,79 @@ def get_queryset(self, request):
...
+        if request.POST.get("post"):  # if admin pressed 'post' on intermediate page
+            form = SendEmailAdminForm(data=request.POST)
+            if form.is_valid():
+                cleaned_data = form.cleaned_data
+
+                recipient_list = [o.email for o in owners if o.email]
+                # run background task that will send email messages
+                send_email_message.apply_async(
+                    # subject, message and recipient_list (emails)
+                    args=[cleaned_data["subject"], cleaned_data["message"], recipient_list],
```
tylerecouture, line 165: is the recipient list bcc?  It should be bcc, and can send direct to the project's admin email (in setting.py)

src/tenant/admin.py
```diff
@@ -115,6 +127,79 @@ def get_queryset(self, request):
...
+
+        request.current_app = self.admin_site.name
+        # we need to create a template of intermediate page with form
+        return TemplateResponse(
+            request,
+            "admin/tenant/send_email_message.html",
+            context={
+                **self.admin_site.each_context(request),
+                "title": "Write your message here",
+                'adminform': adminform,
```
tylerecouture, line 193: single quotes mixed in here?

src/tenant/tasks.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+    # load and render base template with message content
+    msg = get_template("admin/tenant/email/message.txt").render(context={
+        "message": message,
+    })
+
+    # the subject, text, html, from_email and recipient_list parameters are required.
+    send_mass_mail([
+        (subject, textify(msg), msg, settings.DEFAULT_FROM_EMAIL, recipient_list)
+    ], fail_silently=True)
```
tylerecouture, line 22: why fail silently?  What does failing non-silently do?

src/utilities/html.py
```diff
@@ -0,0 +1,15 @@
...
+# html2text is a python script that converts a page of HTML into clean, easy-to-read plain ASCII text
+import html2text
+
+
+def textify(html):
+    """
+    Generate a plain text version of an html content using html2text library.
+    """
+    h = html2text.HTML2Text()
+    # don't ignore links anymore, I like links
```
tylerecouture, line 13: me too =)

src/utilities/tests/test_mail.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertEqual(len(message.get_payload()), 2)
+        self.assertEqual(message.get_payload(0).get_payload(), "Content")
+        self.assertEqual(message.get_payload(0).get_content_type(), "text/plain")
+        self.assertEqual(message.get_payload(1).get_payload(), "HTML Content")
+        self.assertEqual(message.get_payload(1).get_content_type(), "text/html")
+
+
+class BaseEmailBackendTests:
+    """
+    Copy-pasted from `django.tests.mail` without changes.
```
tylerecouture, line 36: What's the purpose of these?

src/utilities/tests/test_mail.py
```diff
@@ -0,0 +1,92 @@
...
+            1,
+            "Expected exactly one message, got %d.\n%r"
+            % (len(mailbox), [m.as_string() for m in mailbox]),
+        )
+        return mailbox[0]
+
+
+class LocmemBackendTests(BaseEmailBackendTests, MailTests, SimpleTestCase):
+    """
+    Copy-pasted from `django.tests.mail` without changes.
```
tylerecouture, line 80: What's the purpose of these tests?

src/utilities/tests/test_html.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+class TestUtilsText(SimpleTestCase):
+    """
+    Various tests for `utilities.text` module.
+    """
+
+    def test_textify(self):
+        """
+        Generate a plain text version of an html content using html2text library.
```
tylerecouture, line 13: links?

tylerecouture (changes requested):
BCC is important for mass mailings, and a few minor comments/questions added.  I haven't actually checked out the code to test yet but I will try to soo.

bashu:
> BCC is important for mass mailings, and a few minor comments/questions added. I haven't actually checked out the code to test yet but I will try to soo.

mailhog

src/tenant/admin.py
tylerecouture: It would be helpful and easier/faster for me to review comments if you replied with details instead of just marking them as resolved.  Otherwise I have to dig into the code which can delay my review for days.

In this case, thanks for the additional comment, except you say "...It first removes the duplicates and returns a dictionary which has....".  This is not correct. `set()` does not return a dictionary, it returns a `set` data structure.  You may have been confused because they both use {curly braces} in their syntax.

image

src/tenant/admin.py
tylerecouture: Please reply to my comment.  I'm not sure why this was marked as resolved.

src/tenant/admin.py
tylerecouture: Ignored or fixed?

src/utilities/tests/test_mail.py
tylerecouture: ?

src/utilities/tests/test_html.py
tylerecouture: ?

tylerecouture (changes requested):
I unresolved some comment.  Please provide a reply so I don't have to dig through the code I've already reviewed.

Also, why does it try to send emails to non-existance email addresses?  If a tenant doesn't have an owner email field, it should not be included or I will get spammed with undeliverable messages or other email errors.

image

tylerecouture:
@bashu also, it doesn't seem to work at all?  Your PR description is too sparse, so don't have the info for what or how you tried to do this, but when I try it:

image

If this isn't how it works, then more info in your PR would be helpful so I don't waste more time on this:

image

tylerecouture:
Examples of a kick ass PR descriptions for your reference: #1404, #1406, #1375

bashu:
to @tylerecouture what do you have:

a) user model with "first name", "last name" and "email" address

Screenshot 2023-07-07 at 8 31 05 PM

b) profile model with "first name", "last name" and "prefered name"

Screenshot 2023-07-07 at 8 30 48 PM

c) also Tenant model has "owner name" and "email"

Screenshot 2023-07-07 at 8 30 35 PM

Questions
a) what should I use as owner's "preferred" name and what to use as fallback (in what order)
b) what should I use as owner's "primary" email address and what to use as fallback (in what order)

tylerecouture:
@bashu we need to merge the public tenant's list of owner names and emails in the tenant model, with the `owner` user info in each tenant's SiteConfig model...I will contact you about this but it is not part of this current issue.

Issue #1386 mentions to use the Siteconfig.owner, so if this is what you are currently doing, awesome. My mistake for looking at the fields in the tenant model!  (Note that more detailed PR descriptions will help with this in the future =)

Answers:
a) Ignore the Profile model, just use `user.get_full_name` or whatever that field/method is
b) For email, use only verified emails.  See `get_users_to_email()` in announcements/tasks.py  as reference, you'll need something like this:
```
your_owner_user_queryset.filter(
                      .filter(emailaddress__verified=True, emailaddress__primary=True)
                      .exclude(email='')
                      .values_list('email', flat=True)
)
```

src/utilities/mail.py
tylerecouture, line 1: Is this module used anymore?  It appears you are using ` send_email_message.apply_async()` instead of this, so can this be removed?

src/tenant/admin.py
```diff
@@ -115,6 +129,102 @@ def get_queryset(self, request):
...
+                full_name_or_username = owner.get_full_name() or owner.username
+
+                # get the email address, but only primary and verified
+                for primary_email_address in EmailAddress.objects.filter(user=owner, primary=True, verified=True):
+                    # make sure it's primary email for real
+                    if primary_email_address.email == user_email(owner):
+                        email = owner.email
+
+                if len(email):  # skip, if email address is empty
+                    recipient_list.append(f"{full_name_or_username} <{email}>")
```
tylerecouture, line 170: Could you include the tenant in the email? That would be good feedback for the user of the form to confirm the expected decks are included, for example:
f"{tenant.name} - {full_name_or_username} <{email}>"
But please confirm the proper field on the tenant model.

tylerecouture (changes requested):
Nice work! Please see a few minor changes requested, and please provide details in each review comment before marking them as resolved in Github.  Thank you!

tylerecouture (changes requested):
Please provide a comment on each of the four issues explaining what you did to resolve it, before you mark them as resolved.  Review is too hard otherwise.

bashu:
> Please provide a comment on each of the four issues explaining what you did to resolve it, before you mark them as resolved. Review is too hard otherwise.

I've provided a comment on each of the four issues above

tylerecouture (changes requested):
If I select only a tenant without an owner email address, I get an error page:

image

image

There is a missing test: this should be caught by your test suite for this feature.  Maybe because you create the email in the SetUp method so it is never captured by your test.

bashu:
to @tylerecouture fixed it latest two commits, see fix https://github.com/bytedeck/bytedeck/pull/1399/commits/b0d6e6af6f7c1a6a467345a6983caf3aa46d24c0 and test https://github.com/bytedeck/bytedeck/pull/1399/commits/9c71ef130ea0b932a8e6d67e218ec83f8847207e

tylerecouture (changes requested):
The page no longer errors, but now if I select a tenant with no owner email, I get aweird situation like this, and when I send the email, it says "0 users emailed successfully".

Is it possible to reject the case when there are no valid email address? or at least put this information somewhere in the form so it's clear to the user (it doesn't have to be pretty, but it can't be confusing for a layperson to use)

image

Issue #1407, opened 2023-06-19
quest bootstrap-table icons don't load - refactor columns


tylerecouture:
It appears the jquery for placing these icons is broken.  Probably just need to fix the selector?  It shoudl show some icons, like repeatable and prerequisite, etc.

![image](https://github.com/bytedeck/bytedeck/assets/10604391/aa6a761b-ad7d-401c-8dfa-928d52a23d6d)

tylerecouture:
I noticed they ARE loading, twice sometimes, for students:

![image](https://github.com/bytedeck/bytedeck/assets/10604391/b44c1a6d-586e-45e1-9087-4d613383e7f6)

tylerecouture:
@Enprogames let's re-purpose this column:

- [ ] Remove the javascript that populates it.
- [ ] Only include the repeatable icon if quest.repeatable != 0 (i.e. check directly, no javascript), including the existing hovertext
- [ ] Also include the blocking icon with appropriate hovertext, and a good icon for it.
- [ ] Change the column width to col-xs-1
- [ ] Give the extra width to the Tags column (increase from 1 to 2)

And while you are redistributing columns, please make a couple other minor changes:

In Progress and Completed tabs:
- [ ] Increase Tag columns by 1. (col-sm-2)
- [ ] Reduce Quest column by 1 (col-sm-4)
- [ ] Leave the xs columns as is

PR #1410, opened 2023-06-21
Fixed Quest List Bugs: Duplicate IDs, Border Styling, Infinite Loading Indicator

### LINKED ISSUE #1403: bootstrap-table bugs with muted / skipped submissions
When a staff member skips a submission by hitting the SKIP button instead of the APPROVE button, it appears in the Approved (for staff) and Completed (for students) and are styled as muted / grayed out.

However, there are two issues:

1. In the student view (at least in light theme) there is no border for the muted panel when expanded
2. In the teacher's expanded view, their is a big blank gap for some reason, and the quick reply form doesn't appear.
image

### LINKED ISSUE #1409: bootstrap-table top border appears in mobile view
Narrwo browser / mobile view (-xs in bootstrap)
image

Enprogames:
## What?
There were still a few major issues with the bootstrap-table quest lists that needed to be worked out. These included:

- **Duplicate IDs**: When a quest was clicked and the detailed content was shown, the content was duplicated from another source. Since the HTML DOM elements contained IDs (which are supposed to be kept unique), the JavaScript code wasn’t sure which elements to target. As a result, it would simply target elements that were closer to the top of the tree. This resulted in several issues, including the loading indicator being shown forever.
- **Reduce JavaScript Includes**: Some JavaScript includes (bootstrap-table.min.js and bootstrap-table-accordion.js) were done multiple times in multiple files, so they were moved to base.html.
- **Bad Table Styling**: Some table styles related to borders were improved.

## Why?
These issues would reduce user satisfaction and make the project more difficult to maintain in the long term.

### Expand/Collapse JavaScript Refactoring
Perhaps the biggest change was to the bootstrap-table-accordion.js, which is responsible for handling expanding/collapsing of quest list rows when the user clicks them. Previously, when content was being inserted underneath a quest row, it was duplicated from a hidden DIV element that existed elsewhere on the page. The hidden DIV was meant to hold the content while the row was collapsed, allowing for the ajax_content_loading.html script to load content while a row was still collapsed.

This duplication of HTML code resulted in unexpected issues, such as the element IDs existing in more than one place. This created several problems, such as the loading indicator being displayed forever if a row was expanded before it had its content inserted, then had its content inserted before being collapsed/expanded again. The reason was that the ajax_content_loading.html script would target the duplicated element. It would insert the content into the duplicated code that was inserted underneath the row, rather than in the original hidden DIV element.

The solution to this was to "move" the hidden DIV underneath the row, rather than duplicate it. This greatly simplified the expanding/collapsing process. Not implementing it this way in the first place was a definite oversight on my part.

### Table Border Styling Fixed for Muted and Collapsed Rows, and Table in Mobile View (#1403 and #1409)
I made some changes to the bootstrap-table-accordion CSS. Firstly, the border around the table in the mobile view is removed. Second, the border is made invisible while a table row is collapsing. And third, the border was removed around the "muted" expanded rows.

```css
/* Fixes #1409 */
.accordian-table {
  border: transparent;
}
.accordian-table > tbody > tr,
.accordian-table > tbody > tr > .collapsing {
  --accordian-table-border-color: transparent !important;
}
/* Fixes #1403 */
.accordian-table > tbody > tr > .muted {
  --accordian-table-border-color: transparent;
}
```

### JavaScript Includes Moved to `quest_manager/base.html`
Since several of the templates in the quest_manager app make use of the bootstrap-table library, I moved the script includes to `quest_manager/base.html`:
```html
{% block js %}
    {% comment %} ### Import Bootstrap Table JS ### {% endcomment %}
    <script src="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-table/1.18.3/bootstrap-table.min.js"
            integrity="sha384-H2wFlf4pp5vxYPba0+XWp9zlTzEKO5Syk7ZZlbgOgZqj5GWvazhyWvq9zHs0INqj"
            crossorigin="anonymous"></script>

    <script src="{% static 'quest_manager/js/bootstrap-table-accordion.js' %}"></script>
{% endblock %}
```

Since other scripts included the `ajax_content_loading.html` script in their JS blocks, I had to use the `{{ block.super }}` tag in some of the templates, such as here in `quest.html`:
```
{% block js %}
  {% comment %} Include from quest_manager/base.html {% endcomment %}
  {{ block.super }}

  {% with quests_available_page=True %}
    {% include "quest_manager/ajax_content_loading.html" %}
  {% endwith %}

{% endblock %}
```

## Testing?
I spent time clicking through the various options that were previously broken, such as the quick reply form, the "Please read submission instruction more carefully..." button, the flag button, etc.

It would also be helpful if the reviewer checked out these changes to make sure nothing is broken.

<details>

<summary>Top Border Fixed in Mobile View</summary>

image

</details>

<details>

<summary>Expanded content for muted rows is no longer muted</summary>

image

image

</details>

<details>

<summary>While a row is collapsing, the border is removed</summary>

image

</details>

## Anything Else?
I looked into fixing #1407, but I'm not sure if it is still issue. I'm also not entirely sure of the intended behavior. Based on the code and comments in `quest_manager/ajax_content_loading.html`, it looks like the hidden labels are only meant to show up in the quests available tab? But some comments seem to indicate that they should also show up in the quest approvals tab, They do show up in the quests available tab for me, but not in the approvals tab. I may need a bit longer to investigate this.

## Review request
@tylerecouture

### DIFF
--- a/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
+++ b/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
@@ -26,29 +26,29 @@ $(document).ready(function () {
       var row_html_id = row._id;
       var quest_id = parseInt(row_html_id.match(/\d+$/)[0], 10);
 
-      // Get the hidden span element
-      var $hiddenSpan = $("#collapse-quest-" + quest_id);
+      // Get the hidden div element
+      var $hiddenDIV = $("#collapse-quest-" + quest_id);
 
       $detail.html('<div class="detail-container" style="display:none;"></div>');
-      // Add all classes from the hidden span to the tr.detail-view element
-      $detail.addClass($hiddenSpan.attr('class'));
+      // Add all classes from the hidden div to the tr.detail-view element
+      $detail.addClass($hiddenDIV.attr('class'));
 
       // Get the detail container div
       var $detailContainer = $detail.find('div.detail-container');
 
       // Create a function to update the row's content
       var updateContent = function () {
-        var detailContent = $hiddenSpan.html();
+        var detailContent = $hiddenDIV.html();
         $detailContainer.html(detailContent);
       };
 
       // Update the row's content initially
       updateContent();
 
-      // Create an observer that updates the row's content when the span's content changes
+      // Create an observer that updates the row's content when the div's content changes
       // This is activated when the ajax script updates the quest content
       var observer = new MutationObserver(updateContent);
-      observer.observe($hiddenSpan[0], { childList: true, subtree: true });
+      observer.observe($hiddenDIV[0], { childList: true, subtree: true });
 
       // When the row is expanded, slide down the detail container
       $detailContainer.slideDown();
@@ -58,6 +58,9 @@ $(document).ready(function () {
       // If the row is already expanded, collapse it
       if ($tr.next().is('tr.detail-view')) {
         $tr.removeClass('active');
+        // Add a class to the row before it starts collapsing
+        var $detailViewRow = $tr.next();
+        $detailViewRow.find('td').addClass('collapsing');
         // send slideUp callback to collapseRow. This causes the currently expanded row to collapse
         $tr.next().find('div.detail-container').slideUp(function () {
           $table.bootstrapTable('collapseRow', $tr.data('index'));
@@ -68,6 +71,8 @@ $(document).ready(function () {
         $table.find('tr.active').each(function () {
           var $expandedRow = $(this);
           $expandedRow.removeClass('active');
+          // Add a class to the row before it starts collapsing
+          $expandedRow.next().find('td').addClass('collapsing');
           $expandedRow.next().find('div.detail-container').slideUp(function () {
             $table.bootstrapTable('collapseRow', $expandedRow.data('index'));
           });
--- a/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
+++ b/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
@@ -11,76 +11,73 @@
     - Adding "active" and "collapse" classes to rows when they are expanded or collapsed
 
 */
-$(document).ready(function () {
-  // Changed from id selector to class selector to target all tables with the class '.accordian-table'
-  var $tables = $('.accordian-table');
 
+/**
+     * Collapses the provided row, animates the collapse, moves the content back to its original location,
+     * and updates the Bootstrap table to reflect the row's collapsed state.
+     *
+     * @param {JQuery} $expandedRow - The row to collapse. Should be a JQuery object.
+     * @param {JQuery} $table - The Bootstrap table in which the row is located. Should be a JQuery object.
+     */
+function collapseRow($expandedRow, $table) {
+  $expandedRow.removeClass('active');
+  $expandedRow.next().find('td').addClass('collapsing');
+  const $detailViewRow = $expandedRow.next();
+  const expanded_row_html_id = $expandedRow.attr("id");
+  const expanded_quest_id = parseInt(expanded_row_html_id.match(/\d+$/)[0], 10);
+  const $expanded_hiddenDIV = $detailViewRow.find("#collapse-quest-" + expanded_quest_id);
+
+  $expanded_hiddenDIV.slideUp(function() {
+    // After animation complete, hide the div and append it back to its original parent
+    $expanded_hiddenDIV.hide().appendTo($expanded_hiddenDIV.data('originalParent'));
+    $table.bootstrapTable('collapseRow', $expandedRow.data('index'));
+  });
+}
+
+$(document).ready(function () {
+  // class selector to target all tables with the class '.accordian-table'
+  const $tables = $('.accordian-table');
+
+  // For each table, initialize the Bootstrap Table and add event listeners for expanding/collapsing rows
   $tables.each(function() {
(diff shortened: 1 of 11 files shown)


tylerecouture (changes requested):
Seem to be some border inconsistancies:
InconsistantBorders

And:
image

Enprogames:
> Seem to be some border inconsistancies: InconsistantBorders
> 
> And: image

That's strange. I wasn't able to replicate this. I removed some styles that might have been causing issues with this though.

tylerecouture:
"Finally, I added a timestamp span element to the submitted_status.html snippet:"  Those are going to need a code comment because there is no context for why they are there and will easily get deleted in some future code cleanup.

tylerecouture:
"I removed some styles that might have been causing issues with this though."  Were they causing the problem or not?  Do you know why those styles were there?  Detail view doesn't sound like the accordion view, it sounds like the quest detail view. Are you sure those styles weren't for the detail view?

Enprogames:
> "I removed some styles that might have been causing issues with this though." Were they causing the problem or not? Do you know why those styles were there? Detail view doesn't sound like the accordion view, it sounds like the quest detail view. Are you sure those styles weren't for the detail view?

From what I've been able to uncover with the git blame, it looks like most of those styles are from 6 years ago when you were experimenting with some bootstrap table accordion stuff. I'm not sure what you mean by the accordion view and quest detail view. Just so we're on the same page, when you click a row, it creates a `<tr class="detail-view">` element, and the content that is inserted has several list-group-item elements. 
image

It looks like the specific commit that you added this in is "89dc604: accordian table scrolling", but from what I can see it doesn't look like what you were working on (`suggestions/templates/suggestions/suggestion_list.html`) is being used anymore. And before I moved those styles, I checked with a few of the other classes you defined, and they weren't in use at all anymore. 

image

Enprogames:
> "Finally, I added a timestamp span element to the submitted_status.html snippet:" Those are going to need a code comment because there is no context for why they are there and will easily get deleted in some future code cleanup.

You're right. I meant to do that. They're there now.

src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
```diff
@@ -1,3 +1,27 @@
...
+  // Get the timestamp from the end of the string. This will break if the HTML structure changes.
+  var timestampA = a.includes('data-timestamp') ? parseFloat(a.split('data-timestamp="')[1].split('"')[0]) : 0;
+  var timestampB = b.includes('data-timestamp') ? parseFloat(b.split('data-timestamp="')[1].split('"')[0]) : 0;
+
+  // Compare the timestamp values.
+  if (timestampA < timestampB) return -1;
+  if (timestampA > timestampB) return 1;
+  return 0;
+}
+
```
tylerecouture, line 24: Uneccessary.  Just use a column with `data-visible='false'`with a numeric value for the datetime (see last_sub_datetime column in profile_list.html) and sort based on that column via `data-sort-name='name_of_other_column'`

This info was in the original issue.

src/quest_manager/templates/quest_manager/snippets/submitted_status.html
tylerecouture, line 1: Revert this file.  No changes needed here.

src/quest_manager/templates/quest_manager/tab_quests_approvals.html
```diff
@@ -23,14 +23,14 @@
         <th class="col-sm-1 col-xs-2" data-field="xp" data-sortable="true">XP</th>
         <th class="col-sm-1 hidden-xs" data-field="group_name" data-sortable="true">{% group_name %}</th>
         <th class="col-sm-2 col-xs-3 hidden-xs" data-field="user" data-sortable="true">User</th>
-        <th class="col-sm-3 hidden-xs" data-field="status" data-sortable="true">Status</th>
+        <th class="col-sm-3 hidden-xs" data-field="status" data-sortable="true" data-sorter="dateSorter">Status</th>
```
tylerecouture, line 26: See original issue for how to do this with only a couple lines of code using `data-sort-name`

src/quest_manager/templates/quest_manager/tab_quests_submission.html
```diff
@@ -21,15 +21,15 @@
         <th class="col-sm-1 text-right col-xs-2" data-field="xp" data-sortable="true">XP</th>
         <th class="col-sm-2 text-right hidden-xs" data-field="campaign" data-sortable="true">Campaign</th>
         <th class="col-sm-1 text-center hidden-xs" data-field="tags" data-sortable="true">{{ config.custom_name_for_tag }}s</th>
-        <th class="col-sm-3 hidden-xs" data-field="status" data-sortable="true">Status</th>
+        <th class="col-sm-3 hidden-xs" data-field="status" data-sortable="true" data-sorter="dateSorter">Status</th>
```
tylerecouture, line 24: per above

tylerecouture (changes requested):
Please submit the data sort changes as it's own PR using the method provided in the issue.

src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
Enprogames: This was changed as per the description in the original issue. I also moved it to a new pull request:
https://github.com/bytedeck/bytedeck/pull/1420

tylerecouture (approved):

PR #1412, opened 2023-06-24
feat: Add Simple Course Registration; Closes #1396

### LINKED ISSUE #1396: Simplified course enrollment option
# The Issue

Currently, even when a deck only has one semester, one group, and one course, the users still need to sign up for the course.  For example, the current default development environment does this:

**Step 1 after registering:**
image

**Step 2 after hitting the Join a Course:**
There are no other options here, so the from is kind of pointless in a simple deck like this:
image

# Solution

 A new SiteConfig option call "simplified course registration" that is True (???) by default.

If SiteConfig.simplified_course_registration == True

Then, fields in the "Join a Course" form that only have a single value, are hidden.  If ALL of the fields only have a single value, then the form can auto-submit itself using those default values.  

I'm not sure if the logic for this is best put in the view or in the form.

Dylan-RG-Patterson:
### What?
This is a site configuration option for simplified course registration for students, an update to the logic in courses/forms.py and courses/views.py that pulls from a siteconfig setting, simplified_course_registration, that when set to true changes the student course registration page like so:
- fields with only one option are hidden, so students accessing the view will not have to choose them manually.
- if all form fields would be hidden and have their values programmatically set, accessing the page instead automatically creates the corresponding CourseStudent object and redirects students to the quests page.

### Why?
Simplified course registration is useful for applications of bytedeck with only one course available, because it streamlines the student application process and trims unnecessary choices for decks that might not utilize multiple courses or blocks.

### How?
A new field in site config has been added, "simplified course registration". When enabled, the CourseStudentForm will check for the number of available options in each field, and if there is only 1 available choice, the field will be hidden from users and the only choice will be programmatically included in the form submission as a keyword argument.

If simplified course registration is enabled and ALL student registration fields are hidden, the get() method called when initially accessing the registration page will detect these conditions and create a Coursestudent object with values equal to the defaults set, then redirect users to the quests page, where they would be redirected normally had they submitted the form manually. 

### Testing?
Tests for logic behind hiding fields and automatically creating an object if all fields are hidden have been added.

### Anything Else?
As referenced in placeholder code comments, an additional component that has yet to be added is the implementation of the UserPassesTestMixin, as changing the core logic of the view impacts existing tests, warranting a refactor. This PR is a functional implementation of simple course registration, and the addition of UserPassesTestMixin to detect illegal student registration page accesses and display an error 403, while important for long-term stability, is currently handled by an error message upon redirect when accessing the student registration page as a student after already having registered for a course.

### Review request
@tylerecouture


Dylan-RG-Patterson:
uncaught merge error is responsible for the deletion of the 3 lines in views, will fix shortly. Sorry for the holdup

tylerecouture (changes requested):
I'm sorry, I don't understand what state this code is currently in, it seems to be a mishmash of commented out and not commented code in a way I can't understand.

A couple notes:

1. Why not just copy the method used for the already hidden field on this form: the user?  it's excluded no matter what (by not being included in the form's `Meta.fields` variable), but it's the same thing, the user needs to be entered into the model forms data at some point, and that is done in the get_form_kwargs() method.  Seems to me the logical place to "insert" the block and course data would be here.  This allows us to inject additional data into the form (via kwargs), so the validation passes (because like user, these are required fields)

2. Don't literally submit the form in the view, if everything is defaults, you don't need a form at all, you just create the studentcourse object in python and redirect to the `success_url`.  Not exactly sure the best view method to do that in, have you worked much with Class Based Views?  maybe the `dispatch` method? really, whatever is the first method in the stack, because nothing else matters.

I feel like we've wasted a week on a fairly simple task that you went off on some way more complicated direction than needed.  Regular check ins would have saved us a week.  No biggie, but please try to check in regularly with your WIPs, even daily like Ethan does, so I can advise and correct before too much time is wasted.

tylerecouture:
@Dylan-RG-Patterson this works for me. Of course it's missing the logic to check SiteConfig if the field should be hidden (and data injected) or not, but that part seems trivial:

**forms.py**
```
class CourseStudentForm(forms.ModelForm):

        ....
        # hide the block widget:
        self.fields['block'].widget = forms.HiddenInput() # <<<<<<<<<<<<<<<<<<<<<

```

and in **views.py**

```
class CourseStudentCreate(NonPublicOnlyViewMixin, SuccessMessageMixin, LoginRequiredMixin, CreateView):
....

    def get_form_kwargs(self):
        kwargs = super(CreateView, self).get_form_kwargs()
        kwargs['instance'] = CourseStudent(user=self.request.user, block = Block.objects.first()) # <<<<<<<<<<<<<<<<<
        return kwargs

```

tylerecouture:
@Dylan-RG-Patterson simplest way I can think of to redirect if all fields are defaults and siteconfig option is on:

**views.py  CourseStudentCreate()**
```
    def get(self, request, *args, **kwargs):
        # check if we need a form at all, or can just create the studentcourse object using all defaults
        if True: # << put the proper logic to check if all defaults and siteconfig option
             # create the object and check for duplicate errors in case the student already exists in this course, block, etc.
            CourseStudent.objects.create(
                user=self.request.user, 
                block = Block.objects.filter(active=True).first(),
                semester = SiteConfig.get().active_semester,
                course = Course.objects.filter(active=True).first()
            )
            # need to actually catch duplicates exception though.  Can probably just throw it out and redirect?  Shouldn't be able to get here if
            # they are already in teh course, but maybe they could enter the url manually into the browser.

            return redirect(self.success_url)

        else:
            return super().get(request, *args, **kwargs)
```

src/courses/views.py
```diff
@@ -214,15 +218,70 @@ class CourseStudentCreate(NonPublicOnlyViewMixin, SuccessMessageMixin, LoginRequ
...
+
+        if form_auto_create_condition:
+            # form_auto_create_condition == True means only one active/current semester, block, course object
+            # initialize semester, block, course objects in variables to create new CourseStudent with
+            auto_create_semester = SiteConfig.get().active_semester
+            auto_create_block = Block.objects.filter(active=True).first()
+            auto_create_course = Course.objects.filter(active=True).first()
+
+            # before object is created, check for existing identical CourseStudent object to avoid duplicate error
+            duplicate_coursestudent = CourseStudent.objects.filter(
```
tylerecouture, line 242: This might be simpler (more DRY) create the object within a try/except block, and catch the duplicate error, rather than repeating all this code here.

alternately, use [get_or_create()](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#get-or-create) method, and if created==False you know it already exists.

src/courses/views.py
```diff
@@ -214,15 +218,70 @@ class CourseStudentCreate(NonPublicOnlyViewMixin, SuccessMessageMixin, LoginRequ
...
+                user=self.request.user,
+                semester=auto_create_semester,
+                block=auto_create_block,
+                course=auto_create_course
+            )
+
+            # if duplicate CourseStudent object exists, raise duplicate error
+            if duplicate_coursestudent.exists():
+                messages.error(self.request, "An identical course entry for this user already exists.")
```
tylerecouture, line 251: Since this will happen automatically, more feedback might be helpful in this message, like:
{{user}} is already registered in {{course}} course, in {{block}} group, for {{semester}} semester (as an example)

src/siteconfig/models.py
```diff
@@ -186,6 +186,13 @@ class SiteConfig(models.Model):
             Check this if you want to cap marks at 100%. This setting is only relevent if you are using mark percentages."
     )
 
+    simplified_course_registration = models.BooleanField(
+        verbose_name="Simple course registration", default=False,
+        help_text="When students are registering for a course, if simple course registration is enabled, they won't have to select choices for \
+            fields with only one option in the course registration form. If all registration fields only have one option, the form will \
+            automatically submit itself."
```
tylerecouture, line 193: help_text="If this option is enabled, when a student registers for a course they will only have to select fields that have more than one option.  If your deck only has a single course, group, and semester, then the user will automatically enrolled in the course without having to submit a form."

tylerecouture (changes requested):
Looks great!  A couple minor tweaks, but looks ready for prime time once you've added the tests.

Tip: make sure the tests fail when they should, by commenting out a line in your code or something similar that should break the test. Adding tests that pass after you've already written the working code, means you aren't certain the test will actually fail when you think they will.

src/courses/views.py
Dylan-RG-Patterson: get_or_create() method implemented!

src/courses/views.py
Dylan-RG-Patterson: pending the implementation of previously discussed changes involving the UserPassesTestMixin view and having illegal accesses of the student registration form return an Error 403, this specific logic change has not been implemented. If a work-around to still display the error message despite the niche method of raising it is desired, please let me know!

src/siteconfig/models.py
Dylan-RG-Patterson: help text changed.

src/siteconfig/models.py
```diff
@@ -186,6 +186,13 @@ class SiteConfig(models.Model):
             Check this if you want to cap marks at 100%. This setting is only relevent if you are using mark percentages."
     )
 
+    simplified_course_registration = models.BooleanField(
+        verbose_name="Simple course registration", default=False,
+        help_text="If this option is enabled, when a student registers for a course they will only have to select fields that have more than one \
+            option. If your deck only has a single course, group, and semester, then the user will automatically enrolled in the course without \
```
tylerecouture, line 192: "will automatically be enrolled" -- typo missing the "be"

src/courses/views.py
tylerecouture: Nah, thanks for checking. Can leave it as is.

tylerecouture (changes requested):
Code is looking good!  Playing around with it:

If I access the add course page manually (just hit back on browser and hit Join a Course button again) as a student I get an error instead of 403 or 404:

image

Same error if I try to add the student to another course as a staff member from the button in their profile:

tylerecouture:
We definitely need a test to fail if a staff member can't access /courses/student/57/join/ view when they should be able to.

tylerecouture:
When a course is added automatically (all defaults), you don't get a confirmation message or the nice dingle sound confirmation.  This user feedback is important especially for new users.

image

src/siteconfig/models.py
Dylan-RG-Patterson: typo fixed!

Dylan-RG-Patterson:
missing increased test coverage for add course with student already enrolled

src/siteconfig/models.py
```diff
@@ -186,6 +186,13 @@ class SiteConfig(models.Model):
             Check this if you want to cap marks at 100%. This setting is only relevent if you are using mark percentages."
     )
 
+    simplified_course_registration = models.BooleanField(
+        verbose_name="Simple course registration", default=False,
```
tylerecouture, line 190: Form shows the wrong name for the field, please remove the verbose name so it uses the field name, which is the name we want.

tylerecouture (changes requested):
Looks great.  One minor review comment, and then one bigger issue which is that the form no-longer defaults to the empty option when there is more than one option.

Here is what the form currently does by default when there is more than one option:

image

Where as you forms appear to default to the first item.

src/courses/views.py
```diff
@@ -207,22 +212,65 @@ def get_success_url(self):
...
+            return super().get(request, *args, **kwargs)
+
     def get_form_kwargs(self):
         kwargs = super(CreateView, self).get_form_kwargs()
-        kwargs['instance'] = CourseStudent(user=self.request.user)
+        # visible fields with defaults set via kwargs have their defaults overwritten whenever a form is manually submitted, no conditional needed
+        kwargs['student_registration'] = True
+        kwargs['instance'] = CourseStudent(user=self.request.user,
+                                           semester=SiteConfig.get().active_semester,
+                                           block=Block.objects.filter(active=True).first(),
```
tylerecouture, line 264: I think this is why the form isn't defaulting to the blank option, because you are setting the defaults here with .first().

Do you have to pass more than just the user?  That's how it used to work. The rest of the logic should be taken care of in the form itself.

src/siteconfig/models.py
Dylan-RG-Patterson: Verbose name removed!

src/courses/views.py
Dylan-RG-Patterson: Those kwarg defaults are necessary to set values for hidden form fields, but on their own they do interfere with the field defaults when not hidden. Conditional added for block and course kwarg defaults so they'll appear as normal when not necessary.
image
note form shows blank option for field with multiple choices while still hiding fields as needed

tylerecouture: FYI, I appreciate the explanation. Makes review so much easier!

PR #1414, opened 2023-06-27
Delete Decks, fix #1255

### LINKED ISSUE #1255: Delete Decks
It would be great to be able to clean up our tenant list by deleting old or test decks.

bashu:
### What?

This PR is for feature requested in #1255 where site administrators can clean up list of "Tenants" by (mass) deleting old or test decks via simple admin "action" (or each object individualy as a bonus), but leaving schemas in database for security reasons. 

### Why?

This feature was requested in #1255, probably because "delete_selected" action was disabled (deleting individual objects was prohibited too) and "safe" deleting was planned, but never implemented.

### How?

A custom two-step admin action was implemented for registered `TenantAdmin` class, named [delete_selected](https://github.com/bytedeck/bytedeck/pull/1414/files#diff-47ae64eb329ad1a0843236bc24bf9196fd52e1cfe453f7a765ab27ad559dad59R41) (see screenshots).

If you look at `Tenant` change list in the admin, you’ll see this feature in action; it's called “delete selected tenants” action.

Upon execution, on a second (or intermediate) step you will see a summary table of what objects were selected (including related) and will be deleted, also confirmation input field where you need to type "confirmation phrase" before you can delete anything (see screenshot). 

**Note**: for multiple selections I decided to use `coolname` library to generate confirmation phrase in format `<word>/<another word>`, when you delete individual object from change form the `<owner.username>/<tenant.name>` format will be used.

Once you confirm deletion, selected objects (and its dependents) will be deleted, but the schema itself will stay untouched and left in database as-is.

Next, if you try to create new tenant object with name that was used before, you will be stopped with error message, telling you that schema with that name still exists in database (see screenshot)

### Testing?

Test coverage for new "delete_selected" action of the admin views has been done, also overriden "_delete_view" method was tested too.

admin action for selected objects:

Screenshot 2023-06-27 at 12 44 15 PM

deleting individual object (on change object page/form): 

Screenshot 2023-07-17 at 8 16 24 PM

Creating new tenant, but schema still exists in db:

Screenshot 2023-07-02 at 1 10 13 PM

### Anything Else?

New dependency: `coolname` library.

`coolname` is a Python script that generates random names and/or slug.

Function `generate` from `coolname` library is used to generate `<owner>/<deck>` keywords when deleting **multiple** decks at once.

### Review request
@tylerecouture


src/tenant/actions.py
```diff
@@ -0,0 +1,143 @@
...
+            # return None to display the change list page again.
+            return None
+    else:
+        # create form and pass the data which objects were selected before triggering 'post' action.
+        # '_selected_action' (ACTION_CHECKBOX_NAME) is required for the admin intermediate form to submit
+        form = DeleteSelectedConfirmationForm(
+            initial={
+                helpers.ACTION_CHECKBOX_NAME: objects.values_list("id", flat=True),
+                # generate "sekret" keyword from two random words
+                "keyword": "/".join(generate(2)),
```
tylerecouture, line 95: Instead of adding a new dependancy, can we just type the deck name for confirmation?  Similar to when you delete a github repo?  This is what is requested in issue #1255.

tylerecouture (changes requested):

src/tenant/actions.py
bashu: there multiple objects (hencse it admin actions), but which one I should use as keyword / passphrase? first one, or just random object from queryset?

tylerecouture: @Bashu, can we only allow a single deck at a time?  It can be an admin action (one selection only) or an action on the detail view of the item.

tylerecouture (changes requested):
Everything looks great, and thank you for the great documentation, however I do not want to be able to delete multiple tenants at the same time.  Each one is a serious action and must be deliberate and individual.

Also, I want to try to limit unnecessary dependencies, so please remove the coolname dependency.  It is not needed if we only delete a single tenant at a time.

You can choose: 1. You can limit the admin action to a single selection (provide error message if more than one selected) or 2. you can just remove it, and we can use the tenant detail view delete button.

bashu:
to @tylerecouture admin action "delete_selected" was removed, now the only option is to open each tenant and click on big red delete button.

deleteme

src/tenant/admin.py
```diff
@@ -130,6 +170,106 @@ def get_queryset(self, request):
                     tenant.update_cached_fields()
         return qs
 
+    def delete_queryset(self, request, queryset):
+        """
+        Given a queryset, delete (but leave the schema) it from the database.
+        """
+        # for reference: https://django-tenants.readthedocs.io/en/stable/use.html#deleting-a-tenant
+        for obj in queryset:
+            obj.delete(force_drop=False)  # delete model, but *DO NOT* drop schema
```
tylerecouture, line 179: This seems like a criticla line to be covered by tests, but codecov patch note says it is missed.  However, what is this method for?  Didn't queryset deletion (deletion from list view) get removed? Or am I confused?

src/tenant/tests/test_admin.py
```diff
@@ -255,6 +282,149 @@ def test_cant_change_existing_name(self):
...
+        response = self.client.get(delete_url)
+        self.assertContains(response, "tenant/tenant/%s/" % self.extra_tenant.pk)
+        self.assertContains(response, "Summary")
+        self.assertContains(response, "Tenants: 1")
+        post = self.client.post(delete_url, delete_dict)
+        self.assertRedirects(post, reverse("admin:index"))
+        # tenant object was removed (extra tenant is gone)
+        self.assertEqual(Tenant.objects.count(), 2)
+        # ...but schema still in database
+        self.assertTrue(schema_exists("extra"))
```
tylerecouture, line 412: Can this be put in its own test?  This is critical, so better to have own explicit test that schema still exists after tenant deletion.

src/tenant/tests/test_admin.py
```diff
@@ -255,6 +282,149 @@ def test_cant_change_existing_name(self):
...
+            )
+            self.extra_tenant.save()
+            domain = self.extra_tenant.get_primary_domain()
+            domain.domain = "extra.localhost"
+
+        self.client = TenantClient(self.public_tenant)
+
+    @override_settings(ROOT_URLCONF=__name__)
+    def test_delete_view(self):
+        """Delete view should restrict access and actually delete items, but leaves schemas in database."""
```
tylerecouture, line 357: "Delete view should restrict access..."<-- can you elaborate on this? What restrictions are actually tested in this test?

"... but leaves schemas in database." This is critical and should be in its own test

src/tenant/tests/test_admin.py
```diff
@@ -255,6 +282,149 @@ def test_cant_change_existing_name(self):
         form.instance = Tenant.get()  # test tenant with schema 'test'
         self.assertFalse(form.is_valid())
 
+    def test_cant_create_if_schema_still_exists(self):
```
tylerecouture, line 285: docstring please

tylerecouture (changes requested):
This are looking great. A couple comments on testing only, which means we're almost done this!

src/tenant/tests/test_admin.py
bashu: Django comes with a built-in permissions system. It provides a way to assign permissions to specific users and groups of users, and it’s used by the Django admin site.
       
The Django admin site uses permissions as follows:

- Access to view objects is limited to users with the “view” or “change” permission for that type of object.
- Access to view the “add” form and add an object is limited to users with the “add” permission for that type of object.
- Access to view the change list, view the “change” form and change an object is limited to users with the “change” permission for that type of object.
- Access to delete an object is limited to users with the “delete” permission for that type of object.

bashu: relevant commit https://github.com/bytedeck/bytedeck/pull/1414/commits/c6b3fe11a1dcb90c0e3d5f0ed5cbd8f5f10e03a4

src/tenant/tests/test_admin.py
bashu: yes, I've split the tests, see https://github.com/bytedeck/bytedeck/pull/1414/commits/c6b3fe11a1dcb90c0e3d5f0ed5cbd8f5f10e03a4

PR #1444, opened 2023-07-19
fix: slug validation on tag creation and tag deletion info; Closes #1282

### LINKED ISSUE #1282: Tags creation
When creating tags from the admin access, if you don't use a hyphen instead of a space, it will not use the tag when "tagged" to a quest, it will create a new one with hyphens. So when I selected "Solve Problems" as a tag on a quest, it created a new one instead: "solve-problems". When I renamed the tags from admin, and added the hyphen to the name, it worked just fine. 
This could be very confusing for folks. We either need some help instruction on the Tag page or eliminate the recreation with hyphens.

Dylan-RG-Patterson:
### What?
This is a fix for the tag create/delete views that remedies two issues:
- The tag create view would not validate tag names properly, causing duplicate tags to be created when accessed through any other view/form
- The tag delete view would not display related quest/badge information on the confirmation page

### Why?
both of these are unintended bugs that create problems when encountered that are severe enough to warrant remediation:
- duplicate tags cause bloat and unintentional inconsistencies in tag organization
- incomplete information regarding the deletion of tags could lead to tags being deleted by mistake

### How?
a form field validator was added in a newly created __init__ method for the tag form.
additionally, a missing context variable was added to the tag delete view that was preventing information from being displayed.

### Testing?
a test for name validation during tag creation has been implemented.

new tag create view with implemented help text and visible validation failure
image

delete view confirmation prior to fix (missing one related quest)
image

delete view confirmation after fix (no assignations changed)
image

### Review request
@tylerecouture

### DIFF
--- a/src/tags/forms.py
+++ b/src/tags/forms.py
@@ -1,5 +1,6 @@
 from django import forms
 from django.utils.text import slugify
+from django.core.validators import validate_slug
 
 from taggit.models import Tag
 
@@ -22,9 +23,15 @@ class TagForm(forms.ModelForm):
         model = Tag
         fields = ['name']
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['name'].validators = [validate_slug]
+        self.fields['name'].help_text = 'Tags cannot contain spaces, use hyphens to create multi-word-tags-like-this'
+
     def save(self, *args):
         # will be saved on super func, assuming commit=True
-        # necessary since slug is unique=True, so we use unique name to generate a new slug to prevent possible duplicates
+        # save method override necessary since slug is unique=True, so we use unique name to generate a new slug to prevent possible duplicates
+        # without it, updating a tag and assigning it to an object will actually re-create the OLD tag and assign that instead...
         self.instance.slug = slugify(self.cleaned_data['name'])
 
         return super().save(*args)
--- a/src/tags/tests/test_views.py
+++ b/src/tags/tests/test_views.py
@@ -207,6 +207,21 @@ class TagCRUDViewTests(ViewTestUtilsMixin, TenantTestCase):
 
         self.assertTrue(Tag.objects.filter(name=form_data['name']).exists())
 
+    def test_CreateView__slug_validation(self):
+        """
+        Invalid slug names provided to Tag CreateView should be rejected
+        validator that must be passed (validate-slug):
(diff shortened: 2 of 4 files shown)


src/tags/forms.py
```diff
@@ -22,6 +23,11 @@ class Meta:
         model = Tag
         fields = ['name']
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.fields['name'].validators = [validate_slug]
+        self.fields['name'].help_text = 'Tags cannot contain spaces, use hyphens to create multi-word-tags-like-this'
+
     def save(self, *args):
```
tylerecouture, line 31: I think this method can be deleted now.

src/tags/tests/test_views.py
```diff
@@ -207,6 +207,17 @@ def test_CreateView(self):
 
         self.assertTrue(Tag.objects.filter(name=form_data['name']).exists())
 
+    def test_CreateView__slug_validation(self):
+        """Invalid names provided to Tag CreateView should be rejected"""
```
tylerecouture, line 211: Better to be more explicit, here or inline code comment WHY it's invalid (must meet slug_validator and link to it)

src/tags/views.py
```diff
@@ -198,6 +198,7 @@ class TagDelete(NonPublicOnlyViewMixin, DeleteView):
     success_url = reverse_lazy('tags:list')
 
     def get_context_data(self, **kwargs):
+        kwargs['view'] = 'staff'
```
tylerecouture, line 201: Nice catch. Thank you!

tylerecouture (changes requested):
A couple micro tweaks. Thank you!

src/tags/forms.py
Dylan-RG-Patterson: This save method override is actually necessary to properly update tags - it has no bearing on creation, but updating a tag's name value doesn't actually change its slug... Deleting this override seems to have no bearing on function until you assign the updated tag to an object, which will actually create a NEW tag using the OLD tag's name - demonstrated here:
image
image
image

src/tags/tests/test_views.py
Dylan-RG-Patterson: code comment fix added!

tylerecouture (changes requested):
Excellent work sir!

tylerecouture (approved):

Issue #1304, opened 2023-03-18
Additional and multiple submission types


tylerecouture:
### Issue
Currently, when students submit a quest, there are only two fields in the submission form: a rich text field and an attach files/upload field  
![image](https://user-images.githubusercontent.com/10604391/226144651-4ad8db4b-9146-418f-9280-2fd7307a6672.png)

This is limiting in several ways.  For example, take a quest that requires the student to answer a few short answer questions and upload a piece of evidence.  Here's an example (#1-5 are short answer questions, and #6 is a task that requires evidence to be uploaded):

![image](https://user-images.githubusercontent.com/10604391/226144368-78b428b4-4e23-479e-ba2f-d637961d8d3f.png)

Here is what a submission looks like:
![image](https://user-images.githubusercontent.com/10604391/226144398-6f4fe9dd-7e76-4c7a-8ae4-d862229e7a65.png)

A couple things to point out:
- The student was able to submit the quest without adding the required attachment.  Question 6 just asks for an attachment, but the quest is "dumb" and the quest can be completed without an attachment, so the teacher has to ask for it.
- Marking the short answer questions is annoying. Although the answers are provided in the Instructor Notes section, it requires the teacher to scroll up and down to check the answers.
- Often what exactly is required of students isn't clear to many students who struggle with reading, or struggle to follow sequential steps.  This is especially true where there are mixed requirements (e.g. answer these questions, AND upload something).

### Possible Solution

Re-design the quests and submissions to use a quiz-like mechanic, where the quest creator can require multiple types of question/response submission criteria (to start: rich text/long answer, short answer written response, or file upload/attachment).   A quest should be able to have an arbitrary number of of these, in any combination.  e.g. 5 short answer and 1 file upload (per the example above).

From @fjbagley:

> There would be 3 types of questions/responses plus the current system: written responses, embedded responses and attached responses . These question types could be selected by the quest creator and multiple questions could be selected.
> 
> Here would be a potential workflow for setup for a written response: 
> 
>    *Creator selects written response.
>    *Creator types question
>    *Creator types answer 
> 
> Workflow for student:
> 
>    *Student qoes through quest details and watches videos etc
>    *Student has questions to answer (or images to embed or files to attach) that teacher selected in questions
>    *Responses would be input directly below each of the questions 
> 
> Workflow for marker:
> 
>    *Question, sample answer and student answers are visible
>    *Teacher approves or returns quest
> 
[...]

PR #1451, opened 2023-07-29
Added Additional and Multiple Submission Types for Quest Submissions

Enprogames:
## What?
I created the models that will allow us to support quests having multiple questions. I also created CRUDL views to allow teachers to create/view/edit/delete questions for quests. For any quest, they can access these views by clicking the new "Edit Questions" button beside the "Edit Prerequisites" button. Finally, I added the functionality for students to submit their answers to questions. 

The new models implement the following ER diagram:

question_uml

The dashed lines indicate weak relationships, meaning, for example, that a Question cannot exist without a Quest.

## Why?
With these changes, teachers can add multiple submissions to quests. This will make it easier for them to enforce that students answer certain questions, and will make multi-question quests easier for students to understand.

### Question and QuestionSubmission Models

Everything related to questions, except for how they are used in the quest_manager app, has been moved to a new app called "questions".

❗ **Edit**: I've made a few crucial design decisions since the last review, and these decisions should be evaluated:
#### on_delete behavior for Questions and QuestionSubmissions
- When a Question's Quest is deleted, so is the Question.
- When a QuestionSubmission's Comment is deleted, so is the QuestionSubmission.
- When a QuestionSubmission's Question is deleted, the QuestionSubmission is *not* deleted. This reflects the same relationship that Quests have to QuestSubmissions.

#### `__str__` String representations of Questions and QuestionSubmissions
This was tricky to come up with. The string represenations needed to be different from other string representations. I thought it would make sense for a Question to be represented as `<Quest Name> (<Question Ordinal in Quest>)`, but this format is already used by QuestSubmissions, so I had to come up with something new. Below is what I went with for each object:

```py
class Question(models.Model):
...
    def __str__(self):
        """Return string representation in format of: [Quest Name] ([Question Type] - [Question Ordinal])
        e.g. "Quest 1 (short_answer - 1)" for the first question in the quest with name "Quest 1"
        """
        return f"{self.quest.name} ({self.type} - {self.ordinal})"
...
class QuestionSubmission(models.Model):
...
    def __str__(self):
        """Return string representation in format of: [Comment datetime string] - ([question ordinal])
        e.g. "2022-3-22 1:00PM - (1)" for the first question in the student's quest submission
        If the question is None, return: [Comment datetime string] - (?)
        """
        comment_time = self.comment.timestamp.strftime("%Y-%m-%d %-I:%M%p")
        if self.question:
            return f"{comment_time} - ({self.question.ordinal})"
        else:
            return f"{comment_time} - (?)"
```

#### `next_ordinal(cls, quest)` Question Class Method
Deciding how to handle incrementing ordinal values was tricky. We want to ensure that Questions can be added, deleted, and reordered by teachers, and we want to avoid the problems that occurred with the Quest Submission ordinals. I tried overriding the `save()` method, overriding the `delete()` method, and creating a `reorder(self, new_position)` method, but each of these would have been out of the scope of this task. So instead, we opted for creating a simple `new_ordinal(cls, quest)` class method for the Question model. It is implemented as follows:
```py
class Question(models.Model):
...
    @classmethod
    def next_ordinal(cls, quest):
        """Return the next ordinal value for a question in this quest"""
        max_ordinal = Question.objects.filter(quest=quest).aggregate(Max('ordinal'))['ordinal__max']
        return (max_ordinal or 0) + 1
```

This can be called like so:
```py
from quest_manager.models import Quest, Question

quest1 = Quest.objects.get(...)
next_ordinal_in_quest = Question.next_ordinal(quest1)

next_question = Question.objects.create(quest=quest1, ordinal=next_ordinal_in_quest, instructions='')
```

Go to `src/questions/models.py` for the full implementation of the question models.

### Question CRUDL Views

These views can accessed through the following URLS:
```py
path('<int:quest_id>/questions/', views.QuestionListView.as_view(), name='question_list'),
path('<int:quest_id>/questions/create/<str:question_type>', views.QuestionCreateView.as_view(),
     name='question_create'),
path('<int:quest_id>/questions/<int:pk>/', views.QuestionUpdateView.as_view(), name='question_update'),
path('<int:quest_id>/questions/<int:pk>/delete/', views.QuestionDeleteView.as_view(), name='question_delete'),
```

The full implementations for these views can be seen in `src/questions/views.py`.

### QuestionSubmission Views
With these changes, when a student fills in their responses for a multi-question quest, they will be shown a form for each question. And teachers can view their responses in a dropdown menu underneath their submission comment.

Please see the screenshots below.

## Testing?
- Tests for Question and QuestionSubmission models in `questions/tests/test_models.py`.
    - 23 new tests
- Tests for Question CRUDL views in `questions/tests/test_views.py`
    - 20 new tests
- Tests for QuestionSubmissions in `quest_manager/tests/test_views.py`
    - SubmissionViewTests
    - `test_ajax_save_draft_has_changes`
    - `test_ajax_save_draft_question_has_changes`
    - `test_ajax_save_draft_question_file_upload`
    - `test_ajax_save_draft_question_file_wrong_type_rejected`
    - `test_ajax_save_draft_questions_no_changes`
    - `test_question_submission_form_fields_for_types`
    - `test_students_can_view_past_question_submissions`
    - `test_teachers_can_view_past_question_submissions`
    - `test_student_cannot_see_solutions_or_notes`
    - `test_teachers_can_see_solutions_and_notes`
    - `test_question_submissions_created_for_new_submission`
    - `test_question_submissions_and_comment_not_created_with_existing_submissions`
    - `test_comment_and_questions_deleted_on_drop`
    - SubmissionCompleteViewTest
    - `test_submit_questions`
    - `test_submit_file_question`
    - `test_students_must_complete_required_questions`
    - `test_questions_allowed_file_type`
- Revised tests for QuestSubmissions in `quest_manager/tests/test_models.py` (due to change of draft_text to draft_comment)
    - `test_submission_mark_completed`

image

image

image

Instructors can see a table of questions, and can easily click to edit or delete them.
image

Students can see a list of questions for a quest in their version of the quest detailed view.
image

Above the original quest submission field, a formset of question response fields is shown. This is where the student responds to the quest questions.
image

Question responses can be viewed in a table by clicking the "Question Submissions" dropdown.
image

Teachers are shown a different view of question submissions that includes solutions and marker notes:
image

## Anything Else?
None of the frontend code is setup to look good on mobile devices. I suggest that we create an issue for it, and work on doing that in a brand-new PR, after this one is merged.

I also wanted to mention that a lot of the question content is truncated, since it is difficult to show a lot of content in a table. For this reason, I added tooltips to the question table shown below. I'm not sure how to show this in a screenshot, but the full content of the instructions and solutions can be seen in a bootstrap tooltip by hovering over the content.
image

## Review request
@tylerecouture


src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
 
         minutes = (self.first_time_completed - self.timestamp).total_seconds() / 60
         return minutes
+
+
+class QuestionType(models.Model):
+    id = models.CharField(
```
tylerecouture, line 994: Don't override id, `id` is a default field all model.Models come with that is equivalent to a primary key (.pk).    `.id` is used all over the place and will be assumed it is a unique identifier for the object.

We don't need to generate primary keys for our models, the default id field works and is what everyone will expect.

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+        max_length=100,
+        primary_key=True,
+        help_text="Short, memorable, unique identifier for a question type e.g. 'short_answer', 'file_upload', 'multiple_choice'.",
+        serialize=False
+    )
+    name = models.CharField(
```
tylerecouture, line 1000: should be unique=True

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+        max_length=255,
+        help_text="The name of the question type e.g. 'Short Answer', 'File Upload', 'Multiple Choice'."
+    )
+    description = models.TextField(blank=True, null=True)
```
tylerecouture, line 1004: help_text="This may be used as help text for the question type."

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+
+    def __str__(self):
```
tylerecouture, line 1006: None of your string methods are covered by tests.

src/quest_manager/models.py
```diff
@@ -439,6 +440,18 @@ def all_drafts(self, user):
 
 
 class Quest(IsAPrereqMixin, HasPrereqsMixin, TagsModelMixin, XPItem):
+    """
```
tylerecouture, line 443: Thanks for the docstring!

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+    solution_file = RestrictedFileField(blank=True, null=True, upload_to='quest_questions/')
+    marker_notes = models.TextField(null=True, blank=True)
+
+    required = models.BooleanField(default=True)
+    datetime_created = models.DateTimeField(auto_now_add=True)
+    datetime_last_edit = models.DateTimeField(auto_now=True)
+    ordinal = models.PositiveIntegerField(default=1)
+
+    quest = models.ForeignKey(Quest, on_delete=models.CASCADE)
+    type = models.ForeignKey(QuestionType, on_delete=models.CASCADE, null=True, blank=True)
```
tylerecouture, line 1022: Why null=true blank=true?  What would it mean for a question to not have a type?

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+    )
+    description = models.TextField(blank=True, null=True)
+
+    def __str__(self):
+        return self.name
+
+
+class Question(models.Model):
+    instructions = models.TextField()
+    solution_text = models.TextField(null=True, blank=True)
```
tylerecouture, line 1012: What's the difference between solution text and marker notes?

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+        help_text="The name of the question type e.g. 'Short Answer', 'File Upload', 'Multiple Choice'."
+    )
+    description = models.TextField(blank=True, null=True)
+
+    def __str__(self):
+        return self.name
+
+
+class Question(models.Model):
+    instructions = models.TextField()
```
tylerecouture, line 1011: All of these fields should have help_text.  Since these will be user facing models, we want good help text (which by default is the help text that appears in forms) to make use of this feature as easy as possible.

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+    required = models.BooleanField(default=True)
+    datetime_created = models.DateTimeField(auto_now_add=True)
+    datetime_last_edit = models.DateTimeField(auto_now=True)
+    ordinal = models.PositiveIntegerField(default=1)
+
+    quest = models.ForeignKey(Quest, on_delete=models.CASCADE)
+    type = models.ForeignKey(QuestionType, on_delete=models.CASCADE, null=True, blank=True)
+
+    class Meta:
+        ordering = ['ordinal']
```
tylerecouture, line 1025: Make sure this sorts the correct direction (lower number first) and ensure there is a test for that.

src/quest_manager/models.py
```diff
@@ -975,3 +988,55 @@ def get_minutes_to_complete(self):
...
+        max_length=255,
+        help_text="The name of the question type e.g. 'Short Answer', 'File Upload', 'Multiple Choice'."
+    )
+    description = models.TextField(blank=True, null=True)
+
+    def __str__(self):
+        return self.name
+
+
+class Question(models.Model):
```
tylerecouture, line 1010: You added a nice docstring to Quest model, how about nice docstring to Question and QuestionSubmission models? =)

tylerecouture (changes requested):
First look over the models.  Nice work! A few notes for you to look at.

src/quest_manager/models.py
Enprogames: Quite a while ago we had a brief conversation about this: https://github.com/bytedeck/bytedeck/issues/1304#issuecomment-1585349776.

I suppose the idea is that teachers could insert the actual answer to a text-based answer to the question in `solution_text`, or a file answer in `solution_file`. And then the teacher would be able to add additional notes in `marker_notes`. 

For example, the instructions would be "what is the best video game?", the solution_text would be the explicit solution e.g. "Tetris", and the marker_notes might elaborate e.g. "The quest hints towards tetris being the best video game".

If this doesn't make sense, maybe either marker_notes or solution_text could be scrapped?

src/quest_manager/models.py
Enprogames: I added help_text for all user-facing fields. I didn't add it for some fields, like the foreignkeys, because I wasn't sure if it would be helpful. Let me know if I should add help_text for them too.
```python
class Question(models.Model):
...
    quest = models.ForeignKey(Quest, on_delete=models.CASCADE)
    type = models.CharField(
        max_length=20, choices=QuestionType.choices, default=QuestionType.SHORT_ANSWER
    )
...
```

src/quest_manager/models.py
Enprogames: My mistake. I was getting Django migration errors, and it was telling me that I had to either provide a default value or allow null values. I changed it so that now the default value is 'short_answer'.

Enprogames:
We decided to add a new field allowed_file_type to Question, so I will need some time to implement it. Please don't merge the PR yet 😄.
```py
# Each tuple contains the file type as both the stored value and the display value
    # e.g. ("media", "Media") or ("image", "Image")
    allowed_file_type = models.CharField(
        max_length=50,
        choices=[(key, key.capitalize()) for key in FILE_MIME_TYPES.keys()],
        default="media",
        help_text=("The types of files that can be uploaded by a student for this question.")
   ) 
```

Enprogames:
I've added and tested the `allowed_file_type` field. The PR is ready for review.

src/quest_manager/models.py
```diff
@@ -12,6 +12,8 @@
 
 from siteconfig.models import SiteConfig
 
+from utilities.models import RestrictedFileField
```
tylerecouture, line 15: This seems weird that the RestrictedFileField is in the `models` module, when we have a `fields` module.  That's not your doing but we should move that at some point.

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+    """Identifies what types of questions can be created.
+
+    This is used in the views and forms to determine what type of question is being created, and to show/hide specific
+    fields. For example, with a short answer question, the solution_file field is hidden, but with a file upload question,
+    the solution_file field is shown.
+    """
+
+    SHORT_ANSWER = "short_answer", "Short Answer"
+    LONG_ANSWER = "long_answer", "Long Answer"
+    FILE_UPLOAD = "file_upload", "File Upload"
```
tylerecouture, line 1004: Your human readable labels are identical to what would be inferred by, for example `FILE_UPLOAD.label`, so seems redundant and unnecessary.  See docs: https://docs.djangoproject.com/en/3.2/ref/models/fields/#enumeration-types

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+
+
+class Question(models.Model):
+    """Used for adding additional submission fields to Quests.
+
+    This model is used for all types of questions, including multiple choice, short answer, file upload, etc.
```
tylerecouture, line 1010: Good for now, but once you get this working, please add a description how it handles different questions types.  Maybe a list of relevant fields for each type or similar info.

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+            "Appears to the student when they are completing the quest. Should provide "
+            "specific guidance, a prompt, or a problem statement for the student to address."
+        )
+    )
+    solution_text = models.TextField(
+        null=True,
+        blank=True,
+        help_text=(
+            "This text will appear to the marker, allowing for an easy comparison of "
+            "the student's submission to an ideal solution."
```
tylerecouture, line 1024: ..."ideal or example solution."

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+        ),
+    )
+    solution_file = RestrictedFileField(
+        blank=True,
+        null=True,
+        upload_to="quest/question/solution/%Y/%m/%d",
+        help_text=(
+            "This file will appear to the marker, allowing them to easily compare "
+            "the student's submission to the ideal solution for file-based questions."
```
tylerecouture, line 1033: ...or example

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+        choices=[(key, key.capitalize()) for key in FILE_MIME_TYPES.keys()],
+        default="media",
+        help_text=("The types of files that can be uploaded by a student for this question.")
+    )
+
+    marker_notes = models.TextField(
+        null=True,
+        blank=True,
+        help_text=(
+            "This text will appear to the marker, providing additional details about "
```
tylerecouture, line 1050: "This text will appear to the marker, providing additional context and info, if needed.

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+            "what a student's submission should be or how the question should be marked."
+        ),
+    )
+    required = models.BooleanField(
+        default=True,
+        help_text=(
+            "If this is checked, the student must complete this "
```
tylerecouture, line 1057: Since the default is True, help text should describe what happens if UNchecked:
"If unchecked, the student can complete the quest without completing this question." or something like that.

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+            "question to complete the quest."
+        ),
+    )
+    datetime_created = models.DateTimeField(auto_now_add=True)
+    datetime_last_edit = models.DateTimeField(auto_now=True)
+    ordinal = models.PositiveIntegerField(
+        default=1,
+        help_text=(
+            "Indicates the order of the question in the quest so that "
```
tylerecouture, line 1066: "Questions will be displayed to students in order from lowest ordinal to highest ordinal."

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+
+
+class QuestionSubmission(models.Model):
+    """Used for storing a student's submission for a question.
+
+    This model is used for all types of questions, including multiple choice, short answer, file upload, etc.
+    When a student submits a quest, each Question will have a QuestionSubmission created for it.
+    """
+
+    response_text = models.TextField(
```
tylerecouture, line 1095: Is this intended for short answer questions?  Short answer should be a CharField with 128 characters or so.  A TextField would be displayed with a Summernote widget for long-answer (with formatting and embedding etc)

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+        null=True,
+        upload_to="quest/question/submission/%Y/%m/%d",
+        help_text=(
+            "A file response to a question. This is what the teacher will "
+            "look at when marking the question."
+        ),
+    )
+
+    # Similar to a quest's submission being deleted, a question's submission should not be deleted if the question is
+    question = models.ForeignKey(Question, on_delete=models.SET_NULL, null=True)
```
tylerecouture, line 1114: If we are not deleting the QuestionSubmission (which I think is a good idea), we'll need to make sure we have good tests covering deleted Questions and ensure those situation are handled.

src/quest_manager/models.py
```diff
@@ -975,3 +989,142 @@ def get_minutes_to_complete(self):
...
+
+    class Meta:
+        ordering = ["question__ordinal"]
+
+    def __str__(self):
+        """Return string representation in format of: [Quest Name] ([Question Type] - [Question Ordinal] - [Quest Submission Ordinal])
+        e.g. "Question 1 (short_answer - 1 - 1)" for the first question in the student's first quest submission
+        """
+        if self.question:
+            return f"{self.question.quest.name} ({self.question.type} - {self.question.ordinal} - {self.submission.ordinal})"
```
tylerecouture, line 1128: This could just use self.submission (which would cover the quest.name and the submission ordinal) and require less chaining.

tylerecouture (changes requested):
Nice work. I left some comments for you!

src/quest_manager/models.py
Enprogames: I was under the impression that we'd use the same for both, and that for short answer, we'd limit the length in the form field. How does that sound? I think it would be simpler than adding  additional `short_solution_text` and `short_response_text` fields.

src/quest_manager/models.py
Enprogames: I think this has already been done. Not sure what else could be tested. Maybe I could add a test ensuring that the files also aren't deleted or something.

image

src/quest_manager/models.py
tylerecouture: Yep, that makes sense. Short answer can use a charfield widget on the form, and Long answer can use a SafeSummernote. Might need to also look at their processing and display (safe filter or not).

Good idea!

src/quest_manager/models.py
tylerecouture: Looks good!

src/quest_manager/models.py
Enprogames: Sure! I'll write down that I have to do that.

src/quest_manager/models.py
Enprogames: You're right. You can do `question.get_type_display()` and it does the same thing.

src/quest_manager/models.py
tylerecouture: Is this because your dev DB has objects with no type, from earlier migrations?  We shouldn't need to provide a type (just as you don't provide a quest).

Might just be a hazard of re-migrating the same thing multiple timeswith different parameters.  Your DB doesn't get reset, so might be good to refresh your db, i.e start a new one on the dev branch, then switch over to this branch and run migrations, so it more closely matches the state of when these migrations would actually happen.

src/quest_manager/models.py
tylerecouture: Makes sense!

src/quest_manager/models.py
```diff
@@ -975,3 +989,150 @@ def get_minutes_to_complete(self):
...
+    class Meta:
+        ordering = ["ordinal"]
+        # ensure two questions in the same quest cannot have the same ordinal
+        unique_together = ["quest", "ordinal"]
+
+    def __str__(self):
+        """Return string representation in format of: [Quest Name] ([Question Type] - [Question Ordinal])
+        e.g. "Quest 1 (short_answer - 1)" for the first question in the quest with name "Quest 1"
+        """
+        return f"{self.quest.name} ({self.type} - {self.ordinal})"
```
tylerecouture, line 1087: I don't think the quest name here is necessary.  When would a question be viewed independent of it's quest? The quest context will always be there, (i.e. the question will always be listed with the quest or with a submission so I think it just needs "ordinal - type".  In the detail view we can add the quest name explicitly.

If you think the quest name is still necessary, it should be the part in brackets, not the ordinal and type.

src/quest_manager/models.py
```diff
@@ -975,3 +989,150 @@ def get_minutes_to_complete(self):
...
+
+    @classmethod
+    def next_ordinal(cls, quest):
+        """Return the next ordinal value for a question in this quest"""
+        max_ordinal = Question.objects.filter(quest=quest).aggregate(Max('ordinal'))['ordinal__max']
```
tylerecouture, line 1092: This feels overly complicated.  Why not just `Question.objects.filter(quest=quest).last()`? Since we are already setting the order via ordinal property?

tylerecouture (commented):
A couple small code notes.  I will fetch the PR and check out your views/forms.

src/quest_manager/models.py
Enprogames:
```py
def __str__(self):
    """Return string representation in format of: [Quest Submission __str__] - ()
    e.g. "Difficult Quest (1) - (1)" for the first question in the student's first quest submission
    """
    if self.question:
        return f"{self.submission} - ({self.question.ordinal})"
    else:
        return f"{self.submission} - (?)"
```

src/quest_manager/models.py
Enprogames:
```py
    datetime_created = models.DateTimeField(auto_now_add=True)
    datetime_last_edit = models.DateTimeField(auto_now=True)
    ordinal = models.PositiveIntegerField(
        default=1,
        help_text=(
            "Questions will be displayed to the student in order from lowest ordinal to "
            "highest ordinal."
        ),
    )
```

src/quest_manager/models.py
Enprogames:
```py
    required = models.BooleanField(
        default=True,
        help_text=(
            "If this field is unchecked, the student can complete the quest without "
            "answering this question."
        ),
    )
```

src/quest_manager/models.py
Enprogames:
```py
    marker_notes = models.TextField(
        null=True,
        blank=True,
        help_text=(
            "This text will appear to the marker, providing additional "
            "context and info, if needed."
        ),
    )
```

src/quest_manager/models.py
Enprogames: done

src/quest_manager/models.py
Enprogames: done

src/quest_manager/models.py
Enprogames: As we discussed, we will save this for later.

src/quest_manager/models.py
Enprogames: I'm not sure. I guess my rational was that doing  `Question.objects.filter(quest=quest).aggregate(Max('ordinal'))['ordinal__max']` is getting just a single number from the database. But if  we did `Question.objects.filter(quest=quest).last().ordinal`, I think it's getting the whole object and then the value afterwards?

Performance gains for this are probably negligible though, so maybe `Question.objects.filter(quest=quest).last().ordinal` is better since it is more readable?

src/quest_manager/models.py
Enprogames:
```py
    def __str__(self):
        """Return string representation in format of: [Question Ordinal] - [Question Type]
        e.g. "1 - short_answer" for the first question in the quest with type short_answer
        """
        return f"{self.ordinal} - {self.type}"
```

Enprogames:
I added a more detailed docstring to the Question class:

```py
...

class Question(models.Model):
    """Used for adding additional submission fields to Quests.

    This model is used for all types of questions, including short answer, long answer, file upload, and hopefully more
    in the future. Different types of questions are distinguished by the question_type field. This field is used in
    the views and forms to determine what type of question is being created, and to show/hide specific fields.

    For example, with a short_answer question, the solution_file/response_file fields are hidden and the solution_text/
    response_text fields are shown, but with a file_upload question, the solution_file/response_file fields are shown etc.

    For file_upload questions, teachers can specify what type of file is expected. For example, if they specify "video",
    the student will only be able to upload a video file. This is enforced in the form validation, allowing only specific
    MIME types. The allowed MIME types are specified in the FILE_MIME_TYPES dictionary.
    """

    instructions = models.TextField(
        help_text=(
            "Appears to the student when they are completing the quest. Should provide "
            "specific guidance, a prompt, or a problem statement for the student to address."
        )
    )
...
```

Enprogames:
> A couple small code notes. I will fetch the PR and check out your views/forms.

What do you mean by "fetching my PR"? Do you want me to push my work that I've done on views/forms for the Questions and QuestionSubmissions in this PR? or a new one after this one is accepted?

src/quest_manager/models.py
tylerecouture: I doubt there is any performance difference at all.  1. Django will convert the SQL behind the scenes to make it efficient, and both only hit the database once.  2 .You can't get the Max value of a queryset without looking at all objects in the queryset anyway!  

But you got me curious so I asked ChatGPT.  It thinks first() is probably faster for the reasons it gives (grain of salt of course..)
https://chat.openai.com/share/8a774ccf-da5e-45b0-a4d6-8553fbeea62e

tylerecouture:
> > A couple small code notes. I will fetch the PR and check out your views/forms.
> 
> What do you mean by "fetching my PR"? Do you want me to push my work that I've done on views/forms for the Questions and QuestionSubmissions in this PR? or a new one after this one is accepted?

I can fetch PRs as a new branch and run them locally.  Yes, please push to the PR.

`git fetch origin pull/ID/head:BRANCH_NAME`   
where: ID =  the github PR #
BRANCH_NAME is whatever I what the local branch to be called that holds the PR

src/quest_manager/templates/quest_manager/submission.html
tylerecouture, line 1: Huge text area doesn't make sense if you can only have limited characters.  This should be the default widget for a CharField?
image

tylerecouture, line 1: Needs an inidication for students what the file-type limitations are:
image
You can inject this into the help text (via form.py) or however you like.

src/comments/templates/comments/comments.html
tylerecouture, line 1: Table appears to be messed up, missing a column:
image

Also, the Marker Notes should just be an (i) icon similar to the question list view (see other comment) and appear when you hover.

tylerecouture (changes requested):
My first walk through of the feature. It is looking awesome.  Seems like you have the base structure working well, now just a matter of making it look good and tweaking.  I left a lot of comments for you to look at =)

If I view a quest, it doesn't show the questions anywhere?  Here it shows some questions:
image

But no indication of the questions anywhere in the Quest, only the old submission instructions.  Questions should be included for staff, similar to how Instructor Notes are.

image

src/comments/templates/comments/comments.html
Enprogames: For the hidden column issue, I fixed this in bootstrap-table-accordion.css. I think you have to do ctrl+f5 to refresh your browser's cached css?

tylerecouture: If you are changing CSS files, make sure to also do the cache-busting by incrementing the version number where it is imported into the template, you'll see a ?v9 at the end of the url as a dummy GET parameter (this doesn't change the file name, but most browsers will reload the file anyway because it's a different URL)

src/quest_manager/templates/quest_manager/submission.html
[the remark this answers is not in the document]
Enprogames: You're right. That was a mistake.

Enprogames: allowed_file_types_indicator

src/questions/forms.py
```diff
@@ -0,0 +1,165 @@
...
+                  'type',
+                  'instructions',
+                  'solution_text',
+                  'solution_file',
+                  'allowed_file_type',
+                  'marker_notes',
+                  'required')
+
+        # hide question_type
+        widgets = {
```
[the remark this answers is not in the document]
Enprogames: image
image

src/questions/templates/questions/question_form.html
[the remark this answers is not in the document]
Enprogames: image
image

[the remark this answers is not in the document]
Enprogames: image

src/quest_manager/templates/quest_manager/submission.html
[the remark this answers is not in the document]
Enprogames: image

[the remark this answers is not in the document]
Enprogames: image

Enprogames: image

src/questions/forms.py
```diff
@@ -0,0 +1,165 @@
...
+    class Meta:
+        model = Question
+        fields = (
+                  'type',
+                  'instructions',
+                  'solution_text',
+                  'solution_file',
+                  'allowed_file_type',
+                  'marker_notes',
+                  'required')
```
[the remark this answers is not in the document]
Enprogames: image
image

src/questions/templates/questions/question_form.html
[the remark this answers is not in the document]
Enprogames: image

src/quest_manager/forms.py
```diff
@@ -117,11 +117,18 @@ def __init__(self, *args, **kwargs):
             'before you can add prerequisites.">Edit Prerequisites</button>' \
             '{% endif %}'
 
+        questions_btn = '{% if object.id %}<a role="button" class="btn btn-default" href="{% url \'questions:list\' object.id %}">' \
+            'Edit Questions</a>' \
```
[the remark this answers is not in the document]
Enprogames: image

src/comments/templates/comments/comments.html
```diff
@@ -58,7 +58,77 @@ <h3 class="comment-heading">
         </li>
         {% endfor %}
         </ul>
-
+        {% if comment.questionsubmission_set %}
```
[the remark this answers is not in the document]
Enprogames: I created a `comments_with_questions.html` file in the questions app. And now that's what the quest app uses for comments.

src/questions/forms.py
```diff
@@ -0,0 +1,165 @@
...
+from crispy_forms.layout import Layout, Div, ButtonHolder, Submit, HTML
+from bytedeck_summernote.widgets import ByteDeckSummernoteSafeInplaceWidget
+from comments.models import Comment
+
+from utilities.fields import FILE_MIME_TYPES, RestrictedFileFormField
+
+from .models import Question, QuestionSubmission
+
+
+class QuestionForm(forms.ModelForm):
```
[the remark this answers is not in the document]
Enprogames: image

src/questions/templates/questions/question_list.html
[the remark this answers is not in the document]
Enprogames: image
image

src/questions/templates/questions/question_form.html
```diff
@@ -0,0 +1,17 @@
+{% extends 'quest_manager/base.html' %}
+{% load crispy_forms_tags %}
+
+{% block title %}{{ page_title }}{% endblock %}
+
+{% block content %}
+
+{{ form.dir }}
+    <h1>{{ heading }}</h1>
```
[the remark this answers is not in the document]
Enprogames: image
```html
{% extends 'questions/base.html' %}
{% load crispy_forms_tags %}
{% load static %}

{% block heading_inner %}Question List for <a href="{% url 'quests:quest_detail' quest.id %}">{{ quest.name }}</a>
<div class="btn-toolbar pt-4">
   {% for value, verbose_name in type_choices %}
      <a class="btn btn-primary"
            href="{% url 'questions:create' quest.id value %}"
            title="Create {{ verbose_name }} Question"><i class="fa fa-plus-circle"></i> {{ verbose_name }}</a>
   {% endfor %}
</div>
{% endif %}
{% endblock %}

{% block content %}
...
```

image
image

Enprogames:
Before taking a look at this, please look at #1476. These changes require that the draft field for quests be changed to a foreignkey to `Comment`, and any changes I have to make to that PR will result in more GIT issues for this PR.

tylerecouture:
@coderabitai review

tylerecouture:
Continued with #1639

PR #1454, opened 2023-08-04
feat: added list view for profiles in blocks; Closes #1293

### LINKED ISSUE #1293: Groups/Blocks list: add # active users detail page
In the Groups/Blocks page, add a column that lists the current number of active users in the group.  Also, hyperlink each block naem to a detail page that lists all the active users in the group.
image

Dylan-RG-Patterson:
### What?
This is a new view for a profile list that's tied to block objects, so every active student that's a part of a given block, specified from the block list view, can be displayed and managed.

### Why?
Displaying members of a given block can aid in administrative tasks involving the oversight of a specific block of students, such as attendance or performance monitoring.

### How?
A new view that inherits from the pre-existing ProfileListView has been created, ProfileListBlockView, which uses a relatedmanager to construct a queryset containing only profiles that are in a given block, as well as active.

### Testing?
a test verifying the new view's object queryset is constructed properly has been created, and all status code tests involving profile views have been extended to contain the new view.

list view
image

block detail/profile list view
image

### Review request
@tylerecouture


src/courses/templates/courses/block_list.html
```diff
@@ -26,11 +27,15 @@
   <tbody>
     {% for object in object_list %}
     <tr {% if not object.active %} class="text-muted" {% endif %}>
-      <td>{{ object.name }}</td>
+      <td>{{ object.name }}</a></td>
```
tylerecouture, line 30: Random closing `a` tag?

src/profile_manager/templates/profile_manager/profile_list.html
```diff
@@ -70,7 +87,7 @@
             </th>
             <th data-sortable='true' data-field='blocks'
                 {% if not request.user.is_staff %}data-visible="false"{% endif %}>
-                {% group_name %}
+                {% group_name %}(s)
```
tylerecouture, line 90: Let's just pluralize this outright

src/profile_manager/templates/profile_manager/profile_list.html
```diff
@@ -9,7 +9,19 @@
...
+    <a class="btn btn-warning" href="{% url 'courses:block_update' block_object.id %}" role="button" title="Edit this {% group_name %}">
+        <i class="fa fa-edit"></i>
+    </a>
+    <a class="btn btn-danger" href="{% url 'courses:block_delete' block_object.id %}" role="button" title="Delete this {% group_name %}">
+        <i class="fa fa-trash-o"></i>
+    </a>
+    {% endif %}
+    <!-- end of custom block user list header -->
+
+    {% if request.user.is_staff and view_type != VIEW_TYPES.BLOCK and block_object %}
```
tylerecouture, line 24: Why `and block_object`?  Seems irrelevant

src/profile_manager/tests/test_views.py
```diff
@@ -396,6 +401,34 @@ def test_profile_list_staff(self):
...
         self.assertEqual(qs.count(), 3)  # self.test_teacher, deck_owner, admin
 
+    def test_profile_list_block__get_queryset(self):
+        """ProfileListBlock view's get queryset method should return a queryset containing only students in the desired block"""
+        self.client.force_login(self.test_teacher)
+
+        # make a new empty block object and get semester + user model to test with
+        testblock = baker.make(Block)
+        current_semester = SiteConfig.get().active_semester
+        user = get_user_model()
```
tylerecouture, line 411: put this at the top of the test file, under the imports, and since it is a model (class) you are getting use capitalized `User`.  See other tests modules for consistency,

src/profile_manager/tests/test_views.py
```diff
@@ -396,6 +401,34 @@ def test_profile_list_staff(self):
         qs = response.context['object_list']
         self.assertEqual(qs.count(), 3)  # self.test_teacher, deck_owner, admin
 
+    def test_profile_list_block__get_queryset(self):
+        """ProfileListBlock view's get queryset method should return a queryset containing only students in the desired block"""
```
tylerecouture, line 405: DOcstring doesn't mention testing inactive students, but it appears below you are checking that too?

src/profile_manager/views.py
```diff
@@ -103,6 +104,27 @@ def get_queryset(self):
...
 
+@method_decorator(staff_member_required, name='dispatch')
+class ProfileListBlock(ProfileList):
+    """lists all students in a given block, is accessed through the block list view and acts as a hybrid profile list and block detail view"""
+    view_type = ViewTypes.BLOCK
+
+    def get_queryset(self):
+        """block object is queried via pk kwarg in request from block list view, then a queryset of profiles is generated via relatedmanager"""
+        block_pk = self.kwargs['pk']
+        block = Block.objects.filter(pk=block_pk).first()
```
tylerecouture, line 115: Don't filter.first() unless you are expecting the possibility of the block not existing (which would result in an empty queryset).  Instead you should use `.get(pk=block_pk)`

To handle possibility of manually entered url with a block pk that doesn't exist, django has a convenience function for that which is probably what you should use here: `get_object_or_404`.  Just make sure the 404 is properly raised (since this is in the get_queryset method I don't know if it works here):

https://docs.djangoproject.com/en/3.2/topics/http/shortcuts/#get-object-or-404

whether you want a test for the bad pk raising a 404 is up to you (what does it currently do if you manually enter 999 into the url?)

src/profile_manager/views.py
```diff
@@ -103,6 +104,27 @@ def get_queryset(self):
...
+        # queryset specifications: profile objects that are: part of active semester, a part of a coursestudent object that's in the desired block
+        profiles_qs = Profile.objects.all_for_active_semester().filter(user__coursestudent__block=block)
+        return self.queryset_append(profiles_qs)
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+
+        # block object is queried again to pull name + description data from
+        context['block_object'] = Block.objects.filter(pk=self.kwargs['pk']).first()
```
tylerecouture, line 124: Don't add another query.  It's unnecessary (and unnecessary queries are our biggest source of lag and processor drain).

Instead, save this into a variable when you first get it in get_queryset() (add a new variable at the top of the class where you set the view_type) and then just use that variable to inject the block into the context.

tylerecouture (changes requested):
Nicely done!  A few suggestions to tighten up the code is all.

src/courses/templates/courses/block_list.html
```diff
@@ -26,11 +27,15 @@
...
       <td>{{ object.description|truncatechars:40 }}</td>
+      <td>{{ object.coursestudent_set.count }}</td>
       <td>{{ object.current_teacher }}</td>
       <td>{{ object.active }}</td>
       <td>
+        <a class="btn btn-info" href="{% url 'profile_manager:profile_list_block' object.id %}" role="button" title="View member list">
```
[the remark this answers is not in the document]
Dylan-RG-Patterson: Button title changed!

src/courses/templates/courses/block_list.html
Dylan-RG-Patterson: Random closing </a> tag removed. That was an uncaught remnant of when the list view was accessed through a hyperlinked name instead of a button

src/profile_manager/templates/profile_manager/profile_list.html
Dylan-RG-Patterson: Pluralized outright!

src/profile_manager/templates/profile_manager/profile_list.html
Dylan-RG-Patterson: the thought was to catch requests that were made illegally via url editing by adding a conditional to accessing object data, but with the implementation of get_object_or_404 this becomes irrelevant, if it was ever relevant before, aha

src/profile_manager/tests/test_views.py
Dylan-RG-Patterson: Test Docstring updated to better reflect inactive testing

src/profile_manager/views.py
Dylan-RG-Patterson: second query eschewed via implementation of block_object as a self variable.

src/profile_manager/views.py
Dylan-RG-Patterson: get_object_or_404 function implemented, added an additional line in the page status codes test to catch new 404 conditional.

src/profile_manager/tests/test_views.py
Dylan-RG-Patterson: User capitalized, the siteconfig query was actually redundant, another test already used the active semester variable (shortened to active_sem so it dodged my ctrl + f), I was able to just call that instance instead of getting a new one.

PR #1455, opened 2023-08-07
feature: custom middleware to handle RequestDataTooBig exception

### LINKED ISSUE #1350: Request body exceeded settings.DATA_UPLOAD_MAX_MEMORY_SIZE.
We need to handle this error by providing user feedback instead of just an error page, since it's not clear to the user what the problem is.

This can be reproduced by editing a quest description too contain too much data (waaaay to much, so will need to paste in a LOT of text)

bashu:
### What?

This is fix for a bug reported in https://github.com/bytedeck/bytedeck/issues/1350 when request body exceeded `settings.DATA_UPLOAD_MAX_MEMORY_SIZE` and exception (server side error) raises, we need to handle this error by providing user feedback instead of just an error page.

### Why?

This bug was reported in https://github.com/bytedeck/bytedeck/issues/1350

### How?

Solved by introducing new middleware class `RequestDataTooBigMiddleware` (see [hackerspace_online/middleware.py](https://github.com/bytedeck/bytedeck/pull/1455/files#diff-0b2510ac5d32351730fabc8ba7d04f062a98292a3e6919b770f7f703697d0ed3)). This custom middleware can handle requests that exceed the `DATA_UPLOAD_MAX_MEMORY_SIZE` setting, by catching `RequestDataTooBig` exception, it generates a more user friendly error message (not just server error as it was before), see screenshot bellow.

`RequestDataTooBigMiddleware` depends on `django.contrib.messages` module, and must be installed after `MessageMiddleware` middleware:
```python
MIDDLEWARE = [
    ...
    'django.contrib.messages.middleware.MessageMiddleware',
    'hackerspace_online.middleware.RequestDataTooBigMiddleware',  # after MessageMiddleware
]

```

### Testing?
Yes, and they are self explanatory.

Screenshot 2023-08-07 at 1 05 14 PM

### Anything Else?

Closes #1350 

### Review request
@tylerecouture

### DIFF
--- a/src/hackerspace_online/middleware.py
+++ b/src/hackerspace_online/middleware.py
@@ -1,4 +1,7 @@
 from django.db import connections
+from django.contrib import messages
+from django.http import HttpResponseRedirect
+from django.core.exceptions import RequestDataTooBig
 
 
 class ForceDebugCursorMiddleware:
@@ -10,3 +13,29 @@ class ForceDebugCursorMiddleware:
         connections['default'].force_debug_cursor = True
         response = self.get_response(request)
         return response
+
+
+class RequestDataTooBigMiddleware:
+    """
+    Custom middleware to handle requests that exceed the settings.DATA_UPLOAD_MAX_MEMORY_SIZE
+    """
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        """
+        If the request body is too big then request.body will throw a
+        RequestDataTooBig exception. Check here explicitly if that will happen.
+        """
+        try:
+            _ = request.body
+        except RequestDataTooBig:
+            # if the size of the request (excluding any file uploads) exceeded settings.DATA_UPLOAD_MAX_MEMORY_SIZE.
+
+            # generate a more sensible response and redirect back
+            messages.add_message(
+                request, messages.ERROR, "The size of the request is too large. Please reduce its size and try again.")
+            return HttpResponseRedirect(request.build_absolute_uri())
+
+        # ...as if nothing has happened
+        return self.get_response(request)
--- a/src/hackerspace_online/middleware.py
+++ b/src/hackerspace_online/middleware.py
@@ -1,7 +1,9 @@
+from django.conf import settings
 from django.db import connections
 from django.contrib import messages
 from django.http import HttpResponseRedirect
 from django.core.exceptions import RequestDataTooBig
+from django.template.defaultfilters import filesizeformat
 
 
 class ForceDebugCursorMiddleware:
@@ -19,6 +21,7 @@ class RequestDataTooBigMiddleware:
     """
     Custom middleware to handle requests that exceed the settings.DATA_UPLOAD_MAX_MEMORY_SIZE
     """
+
     def __init__(self, get_response):
         self.get_response = get_response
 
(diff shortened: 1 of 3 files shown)


bashu:
cc @tylerecouture

src/hackerspace_online/tests/test_middleware.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+
+        # should receive an empty response as if nothing has happened
+        response = mw(request)
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.content, b"")
+
+    @override_settings(DATA_UPLOAD_MAX_MEMORY_SIZE=2, ALLOWED_HOSTS="*")
+    @patch("hackerspace_online.middleware.messages.add_message")
+    def test_request_data_too_big_exception(self, mock_add_message):
```
tylerecouture, line 44: This is probably just my lack of understanding, but it's not clear to me that this test ensures 1. That the specific exception is thrown and 2. that the specific exception is caught?  Maybe this isn't necessary, but from the code it's not obvious the test is specific enough.  Maybe it just needs to check the assert_called for the the correct message?  What are your thoughts?

src/hackerspace_online/middleware.py
```diff
@@ -10,3 +13,29 @@ def __call__(self, request):
...
+        RequestDataTooBig exception. Check here explicitly if that will happen.
+        """
+        try:
+            _ = request.body
+        except RequestDataTooBig:
+            # if the size of the request (excluding any file uploads) exceeded settings.DATA_UPLOAD_MAX_MEMORY_SIZE.
+
+            # generate a more sensible response and redirect back
+            messages.add_message(
+                request, messages.ERROR, "The size of the request is too large. Please reduce its size and try again.")
```
tylerecouture, line 37: Let's use this error (Do template filters work in messages?)

"This requests exceeds the maximum size of {{settings.DATA_UPLOAD_MAX_MEMORY_SIZE|filesizeformat}}.  This is likely caused by one of the text fields contains too much data, either from text (including whitespace) or a large data URI (an image or other data encoded as text)."

tylerecouture (changes requested):
Thank you for the detailed PR.  A couple notes for you.

src/hackerspace_online/middleware.py
bashu: done, more verbose error message added in https://github.com/bytedeck/bytedeck/pull/1455/commits/345b484ecf4c1911f2923b7b08466ad2598f2415

src/hackerspace_online/tests/test_middleware.py
bashu: please, check https://github.com/bytedeck/bytedeck/pull/1455/commits/ade3bac3a1ae640980aa9e4fab2ff1088a343e6e

1) first case aka default behaviour, if nothing happened (request didn't exceeded the limit) act normal and do nothing

2) second case when request exceeded the size limit, but there are no custom middlewares that can handle `RequestDataTooBig` exception, response will be error page (400 Bad Request).

3) third case when request exceeded the size, but this time with installed custom middleware that can handle exception gracefully, redirect back to requested page and show **nice** error message via `django.contrib.messages` framework.

Do you need further explanation?

tylerecouture: Thank you! My comment was specifically about the test.  Is there a reason the test doesn't check that the RequestDataTooBig exception is thrown? That was my concern here.  From your explanation, seems like it should be checking that the specific exception is thrown before it tests that it correctly redirects?

tylerecouture: You have a test that check's without the middleware.  That works for me!  Seems more complicated than just checking for the exception, haha, but does the job!

bashu: to @tylerecouture the whole point of middleware is to catch and handle `RequestDataTooBig` exception, if you want to test for that exception you need:

1. Make sure middleware is disabled
2. Make sure you run tests with DEBUG set to **True**

For some reason you test suite runs with DEBUG set to False, possible solution:

1. Set DEBUG=True in settings, line

PR #1456, opened 2023-08-07
fix: handle "menuitem" links, closes #1397

### LINKED ISSUE #1397: /utilities/menuitem/ external non-relative urls always open in new tab, regardless of setting.
If you set a non-relative link to NOT "open in a new tab" it will still open in a new tab.  The option works for relative urls such as `/courses/ranks/` but not for full urls `https://hackerspace.bytedeck.com/courses/ranks/

I believe there is some javascipt/jquery in the base template that add `target="_blank"` to any non relative links after a page loads.

image

bashu:
### What?
This is fix for a bug reported in https://github.com/bytedeck/bytedeck/issues/1397 where if you set a non-relative link to NOT "open in a new tab" it will still open in a new tab. The option works for relative urls such as `/courses/ranks/` but not for full urls `https://hackerspace.bytedeck.com/courses/ranks/`.

### Why?

This bug was reported in https://github.com/bytedeck/bytedeck/issues/1397

### How?

Quick and dirty fix, where a dummy CSS class was added to distinguish menu links from other links (see [utilities/models.py](https://github.com/bytedeck/bytedeck/pull/1456/commits/a8107d4bc461571f83fed4d9944d63bc8f65d452#diff-e54977f4110713a908cd1bba953065e78bacc6a0692de962b8af772b7a52f7fcR84)), and then these "special" links are excluded from altering as it currently done (see [templates/javascript.html](https://github.com/bytedeck/bytedeck/pull/1456/commits/a8107d4bc461571f83fed4d9944d63bc8f65d452#diff-1812463e491cd4e367a9e3a74f3d0866f999bec3e2cc9c1301cfd806d462d9a4R106))

### Testing?
Manually :)
Nope
### Anything Else?

Dummy CSS class: `menuitem`

Closes #1397 

### Review request
@tylerecouture

### DIFF
--- a/src/templates/javascript.html
+++ b/src/templates/javascript.html
@@ -103,7 +103,8 @@
       // _blank added for comments automatically in python when they are created, so this probabyl only still necessary
       // for in quest/announcement links
       //
-      // exclude dummy (.menuitem) CSS class, fix #1397
+      // exclude dummy (.menuitem) CSS class because these urls will set their own "target" attribute and
+      // should be left untouched
       $("a[href^='http://']:not('.menuitem')").attr("target","_blank");
       $("a[href^='https://']:not('.menuitem')").attr("target","_blank");
       $("a[href^='/media/']").attr("target","_blank");
--- a/src/templates/javascript.html
+++ b/src/templates/javascript.html
@@ -102,8 +102,10 @@
       // Ensure external links and media open in new tabs by adding target="_blank".
       // _blank added for comments automatically in python when they are created, so this probabyl only still necessary
       // for in quest/announcement links
-      $("a[href^='http://']").attr("target","_blank");
-      $("a[href^='https://']").attr("target","_blank");
+      //
+      // exclude dummy (.menuitem) CSS class, fix #1397
+      $("a[href^='http://']:not('.menuitem')").attr("target","_blank");
+      $("a[href^='https://']:not('.menuitem')").attr("target","_blank");
       $("a[href^='/media/']").attr("target","_blank");
 
       // Enable allowfullscreen on responsive iframes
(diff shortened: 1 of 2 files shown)


bashu:
cc @tylerecouture

src/templates/javascript.html
```diff
@@ -102,8 +102,10 @@
       // Ensure external links and media open in new tabs by adding target="_blank".
       // _blank added for comments automatically in python when they are created, so this probabyl only still necessary
       // for in quest/announcement links
-      $("a[href^='http://']").attr("target","_blank");
-      $("a[href^='https://']").attr("target","_blank");
+      //
+      // exclude dummy (.menuitem) CSS class, fix #1397
```
tylerecouture, line 106: A more useful code comment here would explain WHY, e.g "exclude .menuitem class because these urls will set their own target attribute"

src/utilities/models.py
```diff
@@ -80,6 +80,7 @@ class Meta:
 
     def __str__(self):
         target = 'target="_blank"' if self.open_link_in_new_tab else ''
-        return '<a href="{}" {}>' \
+        # append dummy (.menuitem) CSS class, fix #1397
```
tylerecouture, line 83: Can remove this.  Issue #'s don't link and can be ambiguous, so if you want to include issues for contenct, paste the entire url not just a #.  In this case, comment is unnecessary.  It is more of a PR comment that is helpful for review, but doesn't add much info to the code base for future devs.

tylerecouture (changes requested):

src/templates/javascript.html
bashu: corrected, see latest commit https://github.com/bytedeck/bytedeck/pull/1456/commits/b7a28dae9617ba7b29d20c5b5ac9ef58ead8d249

src/utilities/models.py
bashu: corrected, see latest commit https://github.com/bytedeck/bytedeck/pull/1456/commits/b7a28dae9617ba7b29d20c5b5ac9ef58ead8d249

PR #1458, opened 2023-08-09
Fix: return httpresponseredirect instead of raise; Closes #1446

### LINKED ISSUE #1446: [Bytedeck]ERROR (EXTERNAL IP): Internal Server Error: /notifications/read/465/
`raise HttpResponseRedirect(reverse('notifications:list'))` <-- should be return not raise?

TypeError at /notifications/read/465/
exceptions must derive from BaseException

```
  File "/app/src/notifications/views.py", line 75, in read
    raise Http404

During handling of the above exception (), another exception occurred:
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/app/src/tenant/views.py", line 37, in wrapper
    return f(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/app/src/notifications/views.py", line 78, in read
    raise HttpResponseRedirect(reverse('notifications:list'))

Exception Type: TypeError at /notifications/read/465/
Exception Value: exceptions must derive from BaseException
```

Dylan-RG-Patterson:
### What?
small fix to change the way an httpresponseredirect is accessed upon exception in notification views, all redirects should be returned, not raised.
 
### Why?
raising a redirect such as httpresponseredirect will result in an error, make sure to return redirects.

### How?
changed "raise" to "return", plus the exception condition was specified to be "notification.doesnotexist", so it won't catch other errors we would want to have raised incase of any other site-breaking conditions.

### Testing?
a test to assert accessing the notification "read" view with a non-existent notification id  redirects to the notification list has been created

N/A

### Anything Else?
Is there anything specific about this issue I should be addressing? The redirect happens upon exception of a try/except block, if there's a specific, re-creatable issue at play here, would there be any error message that should be displayed?

### Review request
@tylerecouture

### DIFF
--- a/src/notifications/tests/test_views.py
+++ b/src/notifications/tests/test_views.py
@@ -64,6 +64,12 @@ class NotificationViewTests(ViewTestUtilsMixin, TenantTestCase):
         self.assertEqual(self.client.get(reverse('notifications:list')).status_code, 200)
         self.assertEqual(self.client.get(reverse('notifications:list_unread')).status_code, 200)
 
+        # Bad id notification read request should redirect to list view
+        self.assertRedirects(
+            response=self.client.get(reverse('notifications:read', args=[999])),
+            expected_url=reverse('notifications:list'),
+        )
+
         self.assertRedirects(
             response=self.client.get(reverse('notifications:read_all')),
             expected_url=reverse('notifications:list'),
--- a/src/notifications/views.py
+++ b/src/notifications/views.py
@@ -1,5 +1,6 @@
 import json
 
+from django.contrib import messages
 from django.contrib.auth.decorators import login_required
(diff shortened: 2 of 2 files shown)


Dylan-RG-Patterson:
> Can you catch this in a test? This should have never happened but that line of code was never tested. Also typo in old comment "bare" = "rare" can you fix that while in here?

Caught in a test, plus redirect exception correctly specified and error message implemented!

PR #1460, opened 2023-08-13
Modular Side Menu

### LINKED ISSUE #1437: Modular main sidebar menu
Add an option that allows for a list of menu items that can be added or removed from the main menu on the left sidebar:
image

This should be set up similar to the menu list under the hamburger menu, where you can "Edit this list"

image

Actually, we might even be able to use the same model, just add a field that indicates which menu the items are for and filter appropriately for the topbar menu vs sidebar menu

The main sidebar menu currently has several items, the rendering should not change, only which items are listed.

- [x] Covert left sidebar menu to use the MenuItem model
- [x] Add an "edit this menu" option at the bottom for staff users
- [x] Two items should be hardcoded and NOT part of the refactor: Quests and Badges.  These both have submenus and should remain as they are now.
- [x] Users and Quest Approvals (staff only) should also be always visible to staff, hardcoded, and remain as they currently are.
- [x] **Maps, Announcements, Profile, Portfolio** can all be converted to MenuItem model
- [x] Above 4 items should be included in a new deck, so they'll need to be created along with other model data in tenant/initialization.py
- [ ] Above 4 items should  NOT be deletable in the MenuItem list.  However they can be set to visible = False.
- [x] Additional items can be added by provided an icon, url, and label (see MenuItem model), and the left sidebar should render it as a menu item.
- [x] Items should order according to the sort order (but Quests and Badges hardcoded on top, and Students and Quest Approvals hard coded on the bottom, i.e. not part of MenuItem)

yujinyuz:
### What?

Modularize side menu items.

Closes #1437

### Why?

These changes will let the users customize their own decks

### How?

This includes a migration for adding new fields to the MenuItem model, changes to the menu item delete view. A method under the model manager is also added for retrieving or creating the default menu items.

### Testing?

I've added tests for creating default side menu items, and also prevents users from deleting a default menu item.

https://github.com/bytedeck/bytedeck/assets/10972027/fefbb8ed-4628-4b8b-a459-c1a69351a99e

### Anything Else?

- I also modified the update menu item to include a preview for the fa icon
- I added a new profile url `/profiles/own/` because the current one requires a primary key and I don't think you can include that within the `url` of the MenuItem
- 

### Review request
@tylerecouture


tylerecouture (changes requested):
I haven't looked at code, but I fetched the PR to test it out live.  This is awesome!  A few tweaks requested.  We want to minimize the effect on existing users (they should see no change, other than an option to edit the list)

1. The top group doesn't seem to have the correct bootstrap styles (missing bottom border)
2. swap Students and Quest Approvals buttons so Quest Approvals is on the bottom (as before)
image

When I did the migration, I lost some of the menu items that used to be there.  Those items should be the defaults on a new deck and existing deck.  See before on the right and after on the left:
image

Please remove the margin above the "Edit this list" link so it is more associated with the middle button group that can be customized and change text to "Customize this menu".
image


To differentiate the list views, please title the new one "Side Menu Items"

yujinyuz:
@tylerecouture Can you try opening it in an incognito window? I already changed the css and experienced the same thing where it still had no borders. Not sure how to properly fix the cached css issue on existing decks though.

Updated PR to reflect with the requested changes. Can you test again? Thanks!

tylerecouture:
> @tylerecouture Can you try opening it in an incognito window? I already changed the css and experienced the same thing where it still had no borders. Not sure how to properly fix the cached css issue on existing decks though.
> 
> Updated PR to reflect with the requested changes. Can you test again? Thanks!

Try cache busting by incrementing CSS file version in the url after the ? In the (header?) template.

yujinyuz:
@tylerecouture thanks! didn't know about this. updated the version and it should work now

src/profile_manager/urls.py
```diff
@@ -23,6 +23,7 @@
     url(r'^(?P<profile_id>[0-9]+)/comment_ban_toggle/$', views.comment_ban_toggle, name='comment_ban_toggle'),
     url(r'^(?P<profile_id>[0-9]+)/comment_ban/$', views.comment_ban, name='comment_ban'),
     url(r'^(?P<pk>[0-9]+)/edit/$', views.ProfileUpdate.as_view(), name='profile_update'),
+    url(r'^own/$', views.ProfileDetailOwn.as_view(), name='profile_detail_own'),
```
tylerecouture, line 26: Can you add basic access tests for this new view? For anonymous, random authenticated (verifying it is their own), and staff?

tylerecouture (changes requested):

src/profile_manager/urls.py
yujinyuz: added

src/profile_manager/views.py
```diff
@@ -163,16 +163,28 @@ class ProfileDetail(NonPublicOnlyViewMixin, DetailView):
...
             return super().dispatch(*args, **kwargs)
 
         return redirect('quests:quests')
 
+    def get_object(self):
+        pk = self.kwargs.get('pk')
+        if pk is not None:
+            profile = get_object_or_404(Profile, pk=self.kwargs.get('pk'))
+        else:
+            profile = self.request.user.profile
```
tylerecouture, line 180: L180 Is not covered by a test: `profile = self.request.user.profile`

tylerecouture (changes requested):
@yujinyuz Hopefully last series of tweaks required!

src/utilities/templates/utilities/menuitem_form.html
```diff
@@ -12,9 +12,22 @@
...
             <input type="submit" value="{{ submit_btn_value }}" class="btn btn-success"/>
         </form>
 
     </div>
-{% endblock %}
\ No newline at end of file
+{% endblock %}
+
+{% block js %}
+<script>
```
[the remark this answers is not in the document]
yujinyuz: This is just a nice little preview added when typing a font awesome class

src/utilities/templates/utilities/side_menu_items.html
```diff
@@ -0,0 +1,12 @@
...
+<div class="nav list-group modular-side-menu hidden-xs sidebar-menu" id="lg-menu-secondary">
+{% for side_menu_item in side_menu_items %}
+  <a id="{{side_menu_item.label}}-menu" class="list-group-item
+      {% if side_menu_item.url in request.path_info %}active{% endif %}"
+        href="{{ side_menu_item.url }}">
+    <i class="fa fa-{{ side_menu_item.fa_icon }} fa-fw"></i>&nbsp; {{ side_menu_item.label }}
+  </a>
+{% endfor %}
+</div>
+<a href="{% url 'utilities:menu_items' %}?is_side_menu=true"><i class="fa fa-fw fa-pencil"></i>&nbsp;&nbsp;Customize this menu</a>
```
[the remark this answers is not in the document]
yujinyuz: Can't deal with this atm. Not sure if I can apply an immediate fix for this. Will deal with this on the weekend

src/utilities/templates/utilities/menuitem_form.html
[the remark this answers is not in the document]
tylerecouture: oh!  Very cool!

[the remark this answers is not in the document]
tylerecouture: Very cool!

src/profile_manager/tests/test_views.py
```diff
@@ -176,6 +179,23 @@ def test_profile_detail__student_marks_button(self):
         response = self.client.get(reverse('profiles:profile_detail', args=[s_pk]))
         self.assertContains(response, 'View your Mark Calculations')
 
+    def test_profile_own__correct_displayed_text(self):
```
tylerecouture, line 182: Now that this view is serving both profile_own and other profile_detail, there should also be a test to confirm profile_detail actually returns the correct profile (and not own).

Currently we are only checking for 200 response.

src/profile_manager/views.py
```diff
@@ -163,16 +163,28 @@ class ProfileDetail(NonPublicOnlyViewMixin, DetailView):
...
+
+        profile = self.get_object()
+        if profile.user == self.request.user:
             return super().dispatch(*args, **kwargs)
 
         return redirect('quests:quests')
 
+    def get_object(self):
+        pk = self.kwargs.get('pk')
+        profile = self.request.user.profile
```
tylerecouture, line 177: Why is `profile = self.request.user.profile` added here if ProfileDetailOwn has it's own view?

src/utilities/templates/utilities/side_menu_items.html
```diff
@@ -0,0 +1,11 @@
...
+<div class="nav list-group modular-side-menu hidden-xs sidebar-menu" id="lg-menu-secondary">
+{% for side_menu_item in side_menu_items %}
+  <a id="{{side_menu_item.label}}-menu" class="list-group-item
+      {% if side_menu_item.url in request.path_info %}active{% endif %}"
+        href="{{ side_menu_item.url }}">
+    <i class="fa fa-{{ side_menu_item.fa_icon }} fa-fw"></i>&nbsp; {{ side_menu_item.label }}
+  </a>
+{% endfor %}
+</div>
+<a href="{% url 'utilities:menu_items' %}?is_side_menu=true" class="hidden-xs" style="text-decoration: none; color: #777;"><i class="fa fa-fw fa-pencil"></i>&nbsp;&nbsp;Customize this menu</a>
```
tylerecouture, line 11: Why did you opt for using get parameters?  This feels clunky to me and I think makes it difficult for future developers, since we don't use get parameters anywhere else in this project.  Even here for example, it's a hard coded addition that is prone to easy breakage and goes against the whole django url design philosophy.  This should be a view with it's own name.

src/utilities/templates/utilities/side_menu_items_navbar.html
```diff
@@ -0,0 +1,72 @@
...
+        <li class="visible-xs-block">
+            <a title="Approvals" href="{% url 'quests:approvals' %}">
+                <div class="navbar-mobile-dropdown-item visible-xs-block">
+                    <i class="fa fa-shield fa-rotate-180 fa-fw"></i>
+                    <span>Approvals</span>
+                </div>
+            </a>
+        </li>
+        {% endif %}
+        {% endif %}
```
tylerecouture, line 68: Can you tab this code so the if blocks aren't aligned?

src/utilities/templatetags/utility_tags.py
```diff
@@ -58,10 +58,22 @@ def group_name():
...
+@register.inclusion_tag('utilities/side_menu_items.html', takes_context=True)
+def side_menu_list(context):
+    side_menu_items = MenuItem.objects.get_side_menu_items().get_visible_items()
+    return {'side_menu_items': side_menu_items, 'request': context.request}
+
+
+@register.inclusion_tag('utilities/side_menu_items_navbar.html', takes_context=True)
+def side_menu_list_navbar(context):
+    side_menu_items = MenuItem.objects.get_side_menu_items().get_visible_items()
+    return {'side_menu_items': side_menu_items, 'request': context.request if 'request' in context else None}
```
tylerecouture, line 74: These methods don't have any doc strings. What's the purpose behind this check for `request`?

tylerecouture: Docstrings should make it more clear what the difference is between these two tags.

tylerecouture:
@yujinyuz can you reply to each of my comments and mark resolved if they are?

src/profile_manager/views.py
yujinyuz: I was going through the code again and I kinda remember now that the sole reason I added this was just so profile doesn't get set to `None`.

I removed this and replaced it with raising an Http404 instead.

PR #1466, opened 2023-08-27
Fixed duplicate quest list IDs across quests and submissions

Enprogames:
## What?
This issue occurred by some changes I mistakenly made in the bootstrap-table quest list. In the quest list, I changed some IDs. For example, I changed `heading-{{ [s.id](http://s.id) }}` to `heading-quest-{{ s.id }}`. I did this to match the other selectors in the other templates e.g. `heading-quest-{{ [q.id](http://q.id) }}`. But without realizing it, I created another bug! Since quests and submissions have different numberings for their IDs, their ID values can conflict. So due to my changes, the HTML ID values were conflicting across the tabs of the quest list. I noticed this after some strange HTML formatting issues were occurring, and it was a huge headache to discover the cause.

I’m going to give a quick refresher on how the quest list works. Each tab has an HTML table loaded into it, and each row in the table is a quest. It is important to note that each row in the “Available” tab represents a `Quest` object, while each row in the “In Progress”/”Completed” tabs are `QuestSubmission` objects. After the page loads, AJAX is used to insert all of the preview content into the quests. This content is inserted into a hidden DOM location, and whenever a quest is clicked, the preview content is inserted underneath the quest row. And when the preview content is collapsed, the content is sent back to the hidden location, and the quest goes back to its original look.

An example might make this issue easier to understand. Right now, while doing the Bytedeck orientation campaign, an HTML formatting bug can be triggered by this. If after completing the “Welcome to Bytedeck!” quest the student immediately submits the “Screenshots” quest, HTML formatting issues will be present if they collapse the submission as follows:

duplicate_IDs_problem_bad_formatting

This is due to the duplicate IDs that appear in both the “Available” and “Completed” tabs. Since the “Screenshots” submission has an ID of “2” at this point, its elements will be labelled, for example, with the ID `collapse-quest-2`. But the “Bytedeck Class Contract” quest has an ID of 2, so it will label its elements using the same IDs. These IDs will then collide, since every quest list tab is rendered in the DOM simultaneously. This is problematic since the preview content for the submission will be inserted into the wrong spot. And although I’m not entirely certain about the specifics of the issues with the HTML formatting in this case, it has something to do with the fact that the preview content for the “Screenshots” quest is incorrectly inserted into where the “Create an Avatar” quest’s preview content is. Everything still mostly works, but the improper formatting causes the quick reply form to be hidden when the “Screenshots” quest submission is expanded.
The problematic IDs are as follows:

- `collapse-quest-{{s.id}}` and `collapse-quest-{{q.id}}`
- `heading-quest-{{s.id}}` and `heading-quest-{{q.id}}`
- `preview-content-{{s.id}}` and `preview-content-{{q.id}}`

## Why?
To ensure proper functioning of the quest list.

## How?
Instead of changing the IDs for submissions back to `collapse-{{s.id}}`, `heading-{{s.id}}`, etc, I changed them to `collapse-submission-{{s.id}}`, `heading-submission-{{s.id}}`, etc. This makes it easier to identify the elements as submissions for future reference.

## Testing?
Since this is a frontend problem, and since we currently have no way of testing our frontend JavaScript code, I did not write any tests for this.

Although I believe I have fully fixed this issue, and I haven't noticed any other issues, I wouldn't be surprised if the reviewer notices other issues after taking a look! Please let me know about any other potential issues you notice.

The same screenshot as from earlier:

duplicate_IDs_problem_bad_formatting

## Anything Else?
While working on this issue, I also discovered #1467. It is another highly related, yet separate issue. Please also take a look at it.

## Review request
@tylerecouture

### DIFF
--- a/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
+++ b/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
@@ -24,8 +24,14 @@ function collapseRow($expandedRow, $table) {
   $expandedRow.next().find('td').addClass('collapsing');
   const $detailViewRow = $expandedRow.next();
   const expanded_row_html_id = $expandedRow.attr("id");
-  const expanded_quest_id = parseInt(expanded_row_html_id.match(/\d+$/)[0], 10);
-  const $expanded_hiddenDIV = $detailViewRow.find("#collapse-quest-" + expanded_quest_id);
+  const expanded_object_id = parseInt(expanded_row_html_id.match(/\d+$/)[0], 10);
+  let $expanded_hiddenDIV;
+  // find rows matching "collapse-quest-<id>" or "collapse-submission-<id>"
+  if (expanded_row_html_id.includes("quest")) {
+    $expanded_hiddenDIV = $detailViewRow.find(`#collapse-quest-${expanded_object_id}`);
+  } else {
+    $expanded_hiddenDIV = $detailViewRow.find(`#collapse-submission-${expanded_object_id}`);
+  }
 
   $expanded_hiddenDIV.slideUp(function() {
     // After animation complete, hide the div and append it back to its original parent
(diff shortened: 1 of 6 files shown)


tylerecouture (approved):
Nice catch. This would have been a headache if not caught before the end of summer.

Can you add your description of the tabs (in the PR description) to the top of the relevant views?  Future devs will be thankful for the explanation.

Enprogames:
Sure! I'll create a new PR with the description. I think it would make the most sense to have it at the top of the `quests.html` template.

PR #1468, opened 2023-08-27
Added Docstring and Renamed 'ajax' quest_manager View

Enprogames:
### What?
While trying to understand the various views in the `quest_manager` app, I kept coming across the `ajax` function and wondering what it was for. Eventually I decided to take the time to understand what it did, as it could have been important, and I discovered that it allows the teacher to see how many submissions are currently in their marking inbox. The only location I could find that it was being used was in `src/templates/javascript.html`. Since I was very confused about the naming, I ended up spending a few minutes writing a docstring for it and giving it a new name.

### Why?
As mentioned above, I found this function very confusing! I think that my changes will improve the readability of this project's code.

### How?
Here is what the function looks like now:

```py
@non_public_only_view
@login_required
def ajax_submission_count(request):
    """Returns the number of submissions awaiting approval for the current user
    This is used to update the number beside the "Approvals" button in the navbar"""
    if request.is_ajax() and request.method == "POST":
        submission_count = QuestSubmission.objects.all_awaiting_approval(
            teacher=request.user
        ).count()

        return JsonResponse(data={"count": submission_count})
    else:
        raise Http404
```

### Testing?
Yes. It looks like this function didn't have tests previously, so this should increase coverage. I mostly followed the format of other tests for AJAX functions. I ensure get and non-ajax requests are rejected, and that it returns the expected result when accessed in the proper way.

### Anything Else?
Unlike some other views that are used by staff, this view has the `@login_required` decorator, rather than the `@staff_member_required` decorator. I wasn't sure if this was intentional, so I didn't change it, but let me know if I should! I'm sure this function doesn't pose any sort of security risk if students can access it, other than possibly telling them that they have zero submissions awaiting their approval?

### Review request
@tylerecouture

### DIFF
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -14,6 +14,7 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.contenttypes.models import ContentType
 from django.urls import reverse
+from django.http import JsonResponse
 
 from django_tenants.test.cases import TenantTestCase
 from django_tenants.test.client import TenantClient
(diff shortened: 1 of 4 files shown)


tylerecouture (approved):
Thank you!!!

PR #1481, opened 2023-09-02
tests: filterable QuerySetSequence in GFK Select Field/Widget

### LINKED ISSUE #1472: Filterable QuerysetSequence in GFK Select Field/Widget
Currently there is no way to filter the querysets included in a QuerysetSequence for our GFKChoiceField.

We need this to be able to solve issue #1370 so we can filter out/exclude a specific object from the queryset.

bashu:
### What?
This is fix for a bug reported in #1472 that there is no way to filter the querysets included in a `QuerySetSequence` for our `GFKChoiceField`.

### Why?
This bug was reported in #1472
### How?
Quick and dirty fix, see [utilities/widgets.py](https://github.com/bytedeck/bytedeck/pull/1481/commits/5268056aada0d402c0e47426ef77359fc3887c41#diff-c4caff1c56921182fde27c25c76ab9ee7845fe41739f9cb4a13349953f7d7acfR100), speaking shortly I found and fixed bug in `filter_queryset` method that didn't used existing queryset (with all previously applied filtering and custom queries), but instead initialized completely new one.

### Testing?
Yes
None
### Anything Else?
Closes #1472 
### Review request
@tylerecouture

### DIFF
--- a/src/utilities/tests/test_widgets.py
+++ b/src/utilities/tests/test_widgets.py
@@ -4,6 +4,7 @@ import string
 import collections
 
 from django import forms
+from django.db.models import Q
 from django.core import signing
 from django.contrib.auth.models import Group
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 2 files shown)


bashu:
cc @tylerecouture

PR #1485, opened 2023-09-07
Fix error creating maps with already "in use" initial objects

### LINKED ISSUE #1370: Creating maps with duplicate initial objects results in error
When creating a new map object through the create view, selecting a Quest, Rank or Badge as an initial object that is already an initial object for another map throws an error.
image
Results in this error:
image

Exception Type: IntegrityError at /maps/generate/
Exception Value: duplicate key value violates unique constraint "djcytoscape_cytoscape_initial_content_type_id__6fc35a3b_uniq"
DETAIL:  Key (initial_content_type_id, initial_object_id)=(39, 1) already exists.

bashu:
### What?
This is fix for a bug reported in #1370, when creating a new map object through the create view, selecting a `Quest`, `Rank` or `Badge` as an initial object that is already an initial object for another map throws an error.

### Why?
When the bug #1370 was reported, there where no way to filter the querysets included in a `QuerysetSequence` for our `GFKChoiceField` (and all its subclasses). It was impossible to filter out (exclude) already used initial objects, so they were available in autocomplete list of our select2 widget even after being used in creating of other maps.

### How?

Simple, yet viable solution:

* modify `AllowedGFKChoiceField` class, to make it possible to run any actions on `QuerySetSequence` instance without overridding `.__init__()` instance method of subclasses, see [src/utilities/fields.py](https://github.com/bytedeck/bytedeck/pull/1485/commits/30668c797d8d40ef94cb451f157266f1a03bfc12#diff-8d3fdc182c6a97b6c31fb8f2e8ab62a22d94c4dc16b8482520373d3ba5e1f193R197)
* modify `CytoscapeGFKChoiceField` class by *overriding* `overridden_querysetsequence` method, applying custom filtering on `QuerySetSequence` instance to exclude `Quest`, `Rank` or `Badge` from querysets, if it is already an intitial object for another map, see [src/djcytoscape/forms.py](https://github.com/bytedeck/bytedeck/pull/1485/commits/30668c797d8d40ef94cb451f157266f1a03bfc12#diff-2d6c4aeb937381eeb696ddb646e4a16ff79ca0c739d0321fdb0097f2a0ee2030R23)

The [patch](https://github.com/bytedeck/bytedeck/pull/1481/commits/5268056aada0d402c0e47426ef77359fc3887c41) made it possible to use (preserve) custom querysets, not just return *all* existing objects as it was discovered by reporter. This PR goes further and fix `IntegrityError` reported in #1370, by excluding individual objects from querysets "on-a-fly" based on defined rules, next time you load the "generate new map" page and try to type in autocompletion field, you will see that all previously used "initial objects" are gone from the list.

### Testing?
Yes, PR includes test to check that the `overridden_querysetsence` method does custom filtering as described above.

exclude

### Anything Else?
Closes #1370 
### Review request
@tylerecouture

### DIFF
--- a/src/djcytoscape/forms.py
+++ b/src/djcytoscape/forms.py
@@ -32,7 +32,8 @@ class CytoscapeGFKChoiceField(AllowedGFKChoiceField):
             # do not use Quest, Rank or Badge as an initial object,
             # if it is already an intitial object for another map
             queryset_models.append(
-                model.objects.exclude(
+                # use existing queryset, with all previously applied filtering and custom queries
+                qs.exclude(
                     pk__in=CytoScape.objects.filter(
                         initial_content_type=ContentType.objects.get_for_model(model),
                     ).values_list('initial_object_id', flat=True)
--- a/src/djcytoscape/forms.py
+++ b/src/djcytoscape/forms.py
@@ -1,6 +1,8 @@
 from django import forms
 from django.contrib.contenttypes.models import ContentType
 
+from queryset_sequence import QuerySetSequence
+
 from utilities.forms import FutureModelForm
 from utilities.fields import AllowedGFKChoiceField
 
@@ -17,6 +19,23 @@ class CytoscapeGFKChoiceField(AllowedGFKChoiceField):
         So instead provide a hard coded list which is checked during testing to ensure it matches
         what the dynamically loaded list would have produced
     """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        queryset_models = []
+        for qs in self.queryset.get_querysets():
+            model = qs.model
+            # Do not use Quest, Rank or Badge as an initial object, if it is already an intitial object for another map
+            queryset_models.append(
+                model.objects.exclude(
+                    pk__in=CytoScape.objects.filter(
+                        initial_content_type=ContentType.objects.get_for_model(model),
+                    ).values_list('initial_object_id', flat=True)
+                )
+            )
+        # Aggregate querysets
+        self.queryset = QuerySetSequence(*queryset_models)
 
     def get_allowed_model_classes(self):
         model_classes = [
--- a/src/djcytoscape/forms.py
+++ b/src/djcytoscape/forms.py
@@ -21,12 +21,16 @@ class CytoscapeGFKChoiceField(AllowedGFKChoiceField):
     """
 
     def __init__(self, *args, **kwargs):
+        # Initialize the new instance of CytoscapeGFKChoiceField,
+        # inherits features and methods from AllowedGFKChoiceField.
         super().__init__(*args, **kwargs)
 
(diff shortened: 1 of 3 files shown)


src/djcytoscape/tests/test_fields.py
```diff
@@ -8,12 +8,26 @@
...
-        """ Quick test to see if the hardcoded model list is equal to CytoScape.ALLOWED_INITIAL_CONTENT_TYPES """
         dynamically_loaded_models = [
             ct.model_class() for ct in ContentType.objects.filter(CytoScape.ALLOWED_INITIAL_CONTENT_TYPES)]
 
         f = CytoscapeGFKChoiceField()
 
         self.assertEqual(dynamically_loaded_models, [qs.model for qs in f.queryset.get_querysets()])
+
+    def test_queryset_is_filterable(self):
+        from djcytoscape.forms import CytoscapeGFKChoiceField
```
tylerecouture, line 22: It's not at all clear to me what this test is actually testing.  There's no docstring, the code is condensed and has no comments either.

src/djcytoscape/forms.py
```diff
@@ -18,6 +20,23 @@ class CytoscapeGFKChoiceField(AllowedGFKChoiceField):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        queryset_models = []
+        for qs in self.queryset.get_querysets():
+            model = qs.model
+            # Do not use Quest, Rank or Badge as an initial object, if it is already an intitial object for another map
+            queryset_models.append(
+                model.objects.exclude(
+                    pk__in=CytoScape.objects.filter(
```
tylerecouture, line 32: Can you provide a test that these objects are excluded? This should be in its own method, fully tested for the correct return values, and from __init__().

tylerecouture (changes requested):
Needs documentation.  Also please do not submit "dirty" code.  especially "quick and dirty" code.  I'm only interested in well written, clean, and well documented code.  There's enough old dirty code in this project that's hard to maintain.  I don't want to add more of it!

bashu:
> Also please do not submit "dirty" code. especially "quick and dirty" code

replaced with "Simple, yet viable solution..."

src/djcytoscape/forms.py
bashu: Calling the `CytoscapeGFKChoiceField()` class constructor creates, and returns a new instance of the class.

First, Python calls `.__new__()` and then `.__init__()`, resulting in a new and fully initialized intance of `CytoscapeGFKChoiceField`.

src/djcytoscape/tests/test_fields.py
bashu: It's a test to see if objects "in use" are being excluded from widget's queryset or not.

tylerecouture (approved):
Thank you!

PR #1492, opened 2023-09-18
Fix email verification link when you create new deck

### LINKED ISSUE #1490: New deck form email verification doesn't work
Was supposed to be implemented in #1217 but it doesn't work.

The link that is sent uses bytedeck.com/ as a base instead of newdeckname.bytedeck.com/ so the link is bad.

If you manually add the deck name to the verification url it does work.

bashu:
### What?
This is fix for a bug reported in #1490 where when you create new deck the link that is sent  uses bytedeck.com/ as a base instead of newdeckname.bytedeck.com/ so the link is bad.

### Why?

This bug was reported in #1490 

### How?

Simple, but viable solution: 

* modify `CustomAccountAdapter` class by overriding `get_email_confirmation_url` method to use current `Site` object to build absolute url (protocol + domain + path) instead of `request` object, this is can be done by calling `allauth.utils.build_absoulute_uri` function with `request` set to `None`.

### Testing?
Yes, PR includes test to check that the verification link contains correct domain name.

url

### Anything Else?

Closes #1490 

### Review request
@tylerecouture

### DIFF
--- a/src/tenant/tests/test_views.py
+++ b/src/tenant/tests/test_views.py
@@ -136,6 +136,8 @@ class TenantCreateViewTest(ViewTestUtilsMixin, TenantTestCase):
         self.assertIn("Please Confirm Your E-mail Address", mail.outbox[0].subject)
         # expecting to see john.doe@example.com as recipient
         self.assertEqual(mail.outbox[0].to, ['john.doe@example.com'])
+        # expecting to see correct domain name in confirmation link
+        self.assertIn('http://default.localhost/accounts/confirm-email/', mail.outbox[0].body)
 
         owner = SiteConfig.get().deck_owner or None
         self.assertEqual(owner.get_full_name(), "John Doe")  # should be equal and prove the case
--- a/src/tenant/tests/test_views.py
+++ b/src/tenant/tests/test_views.py
@@ -7,6 +7,7 @@ from django.contrib.auth import get_user_model
 from django.shortcuts import reverse
 from django.test import RequestFactory
 
+from allauth.account.models import EmailAddress, EmailConfirmationHMAC
 from django_tenants.test.cases import TenantTestCase
 from django_tenants.test.client import TenantClient
 from django_tenants.utils import get_public_schema_name
@@ -136,8 +137,11 @@ class TenantCreateViewTest(ViewTestUtilsMixin, TenantTestCase):
         self.assertIn("Please Confirm Your E-mail Address", mail.outbox[0].subject)
         # expecting to see john.doe@example.com as recipient
         self.assertEqual(mail.outbox[0].to, ['john.doe@example.com'])
-        # expecting to see correct domain name in confirmation link
-        self.assertIn('http://default.localhost/accounts/confirm-email/', mail.outbox[0].body)
+        # expecting to see correct domain name in confirmation link and make sure link is correct
+        email_address = EmailAddress.objects.get(email="john.doe@example.com")
+        key = EmailConfirmationHMAC(email_address).key
+        confirmation_link = "".join(["http://default.localhost:8000", reverse("account_confirm_email", args=[key])])
+        self.assertIn(confirmation_link, mail.outbox[0].body)
 
         owner = SiteConfig.get().deck_owner or None
         self.assertEqual(owner.get_full_name(), "John Doe")  # should be equal and prove the case
--- a/src/tenant/views.py
+++ b/src/tenant/views.py
@@ -3,13 +3,15 @@ import functools
 from django.contrib.auth.mixins import LoginRequiredMixin
 from django.contrib.sites.models import Site
 from django.db import connection
-from django.http import Http404
+from django.http import Http404, HttpResponseRedirect
 from django.utils.decorators import method_decorator
 from django.views.generic.edit import CreateView
 from django.urls import reverse_lazy
 
 from django_tenants.utils import get_public_schema_name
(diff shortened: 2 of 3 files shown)


bashu:
to @tylerecouture this is how you write tests for admin actions:

https://github.com/bytedeck/bytedeck/blob/develop/src/tenant/tests/test_admin.py#L612

this thing break my tests and disabled until re-written: https://github.com/bytedeck/bytedeck/pull/1492/commits/b3baf2884cf2aa50f5aecb2fa62209ec2812e30f

bashu:
to @tylerecouture I am aware about failing test, but if you run it with command `python src/manage.py test src/tenant` they will pass as expected and I am working on solution...

tylerecouture:
> to @tylerecouture I am aware about failing test, but if you run it with command `python src/manage.py test src/tenant` they will pass as expected and I am working on solution...

@bashu did you have any luck finding a solution?

tylerecouture:
@bashu this is still failing

bashu:
to @tylerecouture possible fix is to force cache clearing for Site objects, see latest commit from me

src/tenant/views.py
```diff
@@ -57,14 +59,12 @@ class TenantCreate(PublicOnlyViewMixin, LoginRequiredMixin, CreateView):
...
         """ Copy the tenant name to the schema_name and the domain_url fields."""
-        from siteconfig.models import SiteConfig
-
         # TODO: this is duplication of code in admin.py.  Move this into the Tenant model?  Perhaps as a pre-save hook?
         form.instance.schema_name = form.instance.name.replace('-', '_')
         form.instance.domain_url = f'{form.instance.name}.{Site.objects.get(id=1).domain}'
 
         # save the form and get the response (HttpResponseRedirect)
-        response = super().form_valid(form)
+        self.object = form.save()
```
tylerecouture, line 67: Is there a reason to skip the super().form_valid() here?  That seems dangerous, as we're only checking some extra things here, not EVERYthing?  I'm not sure, but please clarify why it's not needed if you think it is not needed.

It may be fine as you have it. Please just explain as I'm not sure. THanks!

tylerecouture (commented):
This looks great!  Can you please take a look at my one comment and let me know? Thanks!

src/tenant/views.py
bashu: I think I "unpacked" form_valid method, when I tried to find the source of the bug and left it as-is. Note that `CreateView` class inheritted `form_valid` method from `ModelFormMixin` class:

Screenshot 2023-11-26 at 11 50 01


which inherit it from `FormMixin`:

Screenshot 2023-11-26 at 11 50 08


so not a big deal, but I reverted this unnecessarry changes, that may (but I doubt it) be a source of problems in a future.

tylerecouture: Thanks for the explanation and update!

tylerecouture (changes requested):
Did you check that this works?  Because it doesn't quite work yet, the link is broken if you try to actually go to the url provided.

If I create a new deck from here: http://localhost:8000/decks/new/

Here's the email verification link before this PR (as found in `_sent_mail` directory)
http://localhost:8000/accounts/confirm-email/MQ:1r7pyW:Uq4zQjDHgFLE-7ctbWwxWCvw7TexdQRFhliD8Vb0MKo/

Here's the link after the PR.  Note it added the tenant subdomain, but removed the :8000 required for the development server, and thus results in a broken link.
http://test2.localhost/accounts/confirm-email/MQ:1r7q2D:G---on11aRz1f85lP7y7UtrKEXxwmVGJdEJD6wJ7btc/

I assume it would work with a proper URL in staging/production?  But it needs to work in development too.

The Tenant.get_root_url() is what we use in other places.

bashu:
to @tylerecouture I will fix that too

src/tenant/tests/test_views.py
```diff
@@ -136,6 +136,8 @@ def test_default(self):
         self.assertIn("Please Confirm Your E-mail Address", mail.outbox[0].subject)
         # expecting to see john.doe@example.com as recipient
         self.assertEqual(mail.outbox[0].to, ['john.doe@example.com'])
+        # expecting to see correct domain name in confirmation link
+        self.assertIn('http://default.localhost/accounts/confirm-email/', mail.outbox[0].body)
```
tylerecouture, line 140: This is a 404 page.  The test should verify the url is accessible and return 200 or redirect or something accessible.  That's been the problem and still is with this url.

tylerecouture (changes requested):

src/tenant/tests/test_views.py
bashu: do you know how to run webserver from unittest to test if "http://default.localhost:8000" is running ?

bashu: I've modified `get_email_confirmation_url` method (see latest commit) to use current (active) tenant object to build absolute uri for confirmation link, tests were updated too.

bashu:
> Did you check that this works? Because it doesn't quite work yet, the link is broken if you try to actually go to the url provided.

yes, you can see it on screenshot above

bashu:
> I assume it would work with a proper URL in staging/production? But it needs to work in development too.
> 
> The `Tenant.get_root_url()` is what we use in other places.

Yes, you're right, it should work in development and be accessible on localhost domain, see my latest commit. I've resolved this issue by using `get_root_url` method of `Tenant` object.

PR #1511, opened 2023-10-29
Refactor create_notification_tasks

### LINKED ISSUE #1510: Refactor periodic tasks creation
We currently have a function that creates periodic task for all tenants for sending out emails.

https://github.com/bytedeck/bytedeck/blob/06943bddc318dcf444c1a36b3fec7211dc56a8b4/src/notifications/apps.py#L12-L16

This is not a good approach since that function only gets run if we re-deploy an the application or by re-running the `manage.py runserver` command.

- [x] Convert any periodic task to use the dispatcher task pattern as recommended by `tenant-schemas-celery`: https://github.com/maciej-gol/tenant-schemas-celery#dispatcher-task-pattern
- [ ] Make sure to manually delete the tasks within any deployed instance in the database so that celery won't pick it up
- [x] Update django-celery-beat and tenant-schemas-celery if possible

yujinyuz:
### What?

Refactor creating notification tasks

Resolves #1510 

### Why?

The current implementation has a problem since the `create_notification_tasks` gets run if the app is redeployed. Meaning, newly created tenants won't have that daily task in their periodic task queue.

### How?

A dispatcher task inside `src/notifications/tasks.py::email_notification_to_users_on_all_schemas` is added to celery's periodic task via `celery.py::app.conf.beat_schedule` which gets called at 5:00 daily. This task then calls `src/notifications/tasks.py::email_notification_to_users_on_schema` that contains the code that sends the email notifications to users. 

### Review request
@tylerecouture

### DIFF
--- a/src/hackerspace_online/celery.py
+++ b/src/hackerspace_online/celery.py
@@ -29,10 +29,14 @@ app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
 # always attempt to run a schedule that is listed in the database even if
 # it is not listed here or the name of the task is changed.
 app.conf.beat_schedule = {
+    "Send daily email of noticiations to all schemas": {
+        "task": "notifications.tasks.email_notification_to_users_on_all_schemas",
+        "schedule": crontab(minute=0, hour=5),
+    },
     "Invalidate Profile Cache for all schemas daily task": {
         "task": "profile_manager.tasks.invalidate_profile_cache_in_all_schemas",
         "schedule": crontab(minute=0, hour=0),
-    }
+    },
 }
 
 
--- a/src/hackerspace_online/celery.py
+++ b/src/hackerspace_online/celery.py
@@ -29,12 +29,12 @@ app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
 # always attempt to run a schedule that is listed in the database even if
 # it is not listed here or the name of the task is changed.
 app.conf.beat_schedule = {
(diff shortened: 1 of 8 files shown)


src/hackerspace_online/celery.py
```diff
@@ -29,10 +29,14 @@
 # always attempt to run a schedule that is listed in the database even if
 # it is not listed here or the name of the task is changed.
 app.conf.beat_schedule = {
+    "Send daily email of noticiations to all schemas": {
```
tylerecouture, line 32: Typo in notifications

src/hackerspace_online/celery.py
```diff
@@ -29,10 +29,14 @@
...
+        "task": "notifications.tasks.email_notification_to_users_on_all_schemas",
+        "schedule": crontab(minute=0, hour=5),
+    },
     "Invalidate Profile Cache for all schemas daily task": {
         "task": "profile_manager.tasks.invalidate_profile_cache_in_all_schemas",
```
tylerecouture, line 37: I know this was a previous PR, but this needs a better name.  We do not cache profiles, so we are not invalidating profile caches.  What does this method actually do? it should be named to match what it's doing.

tylerecouture (changes requested):
Typo and method name improvement.

tylerecouture:
@yujinyuz have the comments been resolved? If so, please reply to each and mark as resolved.

yujinyuz:
@tylerecouture done.

Does the new description/method name look better?

PR #1520, opened 2023-11-04
Fix badge sort order

### LINKED ISSUE #1484: Badges types in wrong order on user profile page
They don't appear according to the sort order of the types.
`/badges/types/`
image

Student profile:
image

yujinyuz:
### What?

Fix incorrect badge type sort order

Fixes #1484 

### Why?

To be consistent with the sort order of the badge type displayed

### How?

Inside `src/badges/models.py::BadgeAssertionManager::all_for_user_distinct`, we call the `sorted` function. But the parameters for sorting only considers the `badge.sort_order` which leaves out the badge type order.

To fix this, we simply need to pass a list or tuple that contains the badge type's sort order and the badge's sort order 

### Testing?

Added a test that includes modifying the sort_order of the badge_type so we know that it would consider sorting the queryset by badge_type.sort_order and then badge.sort_order

### Anything Else?

Reduced the number of queries when performing the sort by using `prefetch_related` when listing a student's badge

### Review request
@tylerecouture

### DIFF
--- a/src/badges/models.py
+++ b/src/badges/models.py
@@ -270,8 +270,9 @@ class BadgeAssertionManager(models.Manager):
         This only works in a postgresql database, but the app is designed around postgres
         https://docs.djangoproject.com/en/1.10/ref/models/querysets/#distinct
         """
-        qs = self.get_queryset(False).get_user(user).order_by('badge_id').distinct('badge_id')
-        sorted_qs = sorted(qs, key=lambda x: x.badge.sort_order or 0)  # sort_order defaults to 0 if not set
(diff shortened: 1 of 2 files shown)


tylerecouture:
@yujinyuz Why is testing NA?

tylerecouture:
Do I need to fetch these changes and manually test them myself?  Does the test you modified cover the changes?  How do we know this works?

yujinyuz:
> Does the test you modified cover the changes?

Nope. It was just there because it was much easier to read the failed tests since I would easily know the order they got.

> Why is testing NA?

I'll add another test case that includes different orders of the badge type and will update the description.

I just manually tested it as of the moment and seems to work properly.

tylerecouture:
Ok no prob. If you manually test it please note that, and preferably past in some screenshots showing the manual test works!

PR #1530, opened 2023-11-16
Owner user list tweaks

### LINKED ISSUE #1503: Owner user list tweaks
/admin/tenant/tenant/

(not referring to the deprecated)

- [ ] The list of owner emails should show all addresses, verified and unverified.
- [ ] Add a new "verified" column beside the email address list that shows a green tick if the email address is verified and a red x if it is not verified (similar to Google Sign On enabled)
- [ ] owner name and owner email should be sortable

bashu:
### What?
This PR is for feature requested in https://github.com/bytedeck/bytedeck/issues/1503, request was to made some tweaks at /admin/tenant/tenant/

### Why?
It was requested in https://github.com/bytedeck/bytedeck/issues/1503

### How?

> The list of owner emails should show all addresses, verified and unverified.

Two new non-editable DB fields were added: `owner_full_name_cached` and `owner_email_cached` to store calculated values, that was previously done "on-a-fly" by `TenantAdmin.owner_full_name_text` and `TenantAdmin.owner_email_text` methods. 

`TenantAdmin.owner_full_name_text` and `TenantAdmin.owner_email_text` methods were modified to match this changes. 

`Tenant.get_search_results` method was removed, instead cached fields was included in a list of searchable fields.

> Add a new "verified" column beside the email address list that shows a green tick if the email address is verified and a red x if it is not verified (similar to Google Sign On enabled)

`TenantAdmin` class was extended with new custom field (method) `owner_email_verified_boolean`  that retrieves (from SiteConfig object) owner's primary email address, and returns boolean `True` if the email was verified, otherwise `False` (see screenshot below).

> owner name and owner email should be sortable

New non-editable DB fields are used to make the columns sortable.

### Testing?

Updated to match new requirements, see above.

Screenshot 2023-11-16 at 20 24 37

### Anything Else?

Closes #1503 

### Review request
@tylerecouture


bashu:
cc @tylerecouture

tylerecouture:
@bashu how/when are the new cached fields updated?

bashu:
> @bashu how/when are the new cached fields updated?

in `TenantAdmin.get_queryset` method, it gets called on every page load/refresh (/admin/tetant/ tenant/ list), see my comment inside the patch. Why it there? I don't know, twas made by other contributors, they can speak for themselves.

I can suggest solution similar to https://github.com/mpasternak/django-denorm-iplweb

src/tenant/tests/test_admin.py
```diff
@@ -235,7 +250,7 @@ def test_search_on_custom_fields(self):
             reverse("admin:{}_{}_changelist".format("tenant", "tenant")) + "?q=doe.com"
         )
         # confirm the search returned one object (by email address)
-        self.assertContains(response, "1 result")
+        self.assertContains(response, "2 result")
```
tylerecouture, line 253: Code comment says return one object but you are returning two... 

`# confirm the search returned one object (by email address)`

tylerecouture (changes requested):
Thank you! small typo in code comment I think.

src/tenant/tests/test_admin.py
bashu: correct, it was my mistake. it should say "two objects", since after this patch "change list" view shows both verified and unverified email

bashu: should be corrected in a latest commit

bashu (commented):
resolved

PR #1532, opened 2023-11-19
Fix new quest not appearing for newly registered students

### LINKED ISSUE #957: quest available outside course and no prereqs doesn't appear for an existing student without a course
Related to #936 

Currently the 'update_quest_conditions_all_users()` task only runs on current students currently enrolled in a course.

**prerequisites.tasks.update_quest_conditions_all_users**
```
# only cycle through users currently in a course
    users = CourseStudent.objects.all_users_for_active_semester()
```

But this leaves an edge case hole, because a new user that isn't in a course won't  be able to see any new quests that are created (new quest that has no prereqs and is also available outside of course).

Steps to Reproduce:

1. Create a new student, DON'T join a course
2. As Admin, create a new quest with no prereqs, and also "Available outside course" = True
3. Confirm celery task is received:
```
celery_1       | [2022-03-15 20:01:10,036: INFO/MainProcess] Task prerequisites.tasks.update_quest_conditions_all_users[fc3966b0-a272-45c1-923b-1a70ba75c13e] received
```
4. Wait for celery task is complete
5. Refresh Available Quests on the new user... **quest doesn't show up**
6. Create another new user, that quest appears for them (under the "join a course" button.) because they do a calculation when the account is created.

yujinyuz:
### What?

Fixes #957 

### Why?

Newly registered students without a course could not see a newly created quest event though it's available outside a course.

### How?

By adding a new handler `update_cache_triggered_by_quests_available_outside_course`, this calls the `update_conditions_for_quest` which then makes the student view the course.

### Testing?

* Tested new receiver
* Manually tested (see screen recording)

https://github.com/bytedeck/bytedeck/assets/10972027/3b269787-f40f-47eb-8df3-f7e39580893f

### Review request
@tylerecouture


tylerecouture (commented):
Taking a look through the code, this seems like a more complicated fix than is required.

Why not just add a little logic to the existing Quest post_save receiver method,

In one of the other tasks I see this:

```
    if quest.available_outside_course:
        users = User.objects.all()
    else:
        users = CourseStudent.objects.all_users_for_active_semester()
```

Adding this to the other tasks should solve the problem in the cleanest way?

src/quest_manager/tests/test_managers.py
```diff
@@ -462,6 +464,39 @@ def test_get_available(self):
                     ordered=False
                 )
 
+    @patch('prerequisites.signals.update_conditions_for_quest.apply_async')
+    def test_get_available__student_outside_course(self, mock_task):
```
tylerecouture, line 468: This is not a Quest Manager method, why is this test here?

tylerecouture (changes requested):

tylerecouture (approved):
I take it all back! Ignore my previous comments.  I understand now what you did and why THANK YOU!

PR #1533, opened 2023-11-19
Fix emails sent to unverified users

### LINKED ISSUE #1374: Emails being sent to unverified email
I am getting email rejected reports. However when I check the students profile, it shows the email is unverified.

Why is it sending Unread Notifications email to an unverified address?

Need to check:

1. That unverified email addresses are filtered out of Notification emails (emails go out for Announcements and Notifications, we might have accidentally only filtered one of them)

Or

2. Why else would this happen? Maybe they have a verified login email (Gmail) but changed their notifications email in their profile and the filter logic is buggy (check for a verified email addy, but sending to the unverified one?)

yujinyuz:
### What?

Fixes #1374 

### Why?

We don't want to send emails to unverified users.
### How?

Currently, we only filter the list of users to send emails to by `profile__get_notifications_by_email=True`. We have an existing method for announcements that has a filter for students who are enrolled in a semester and has a verified email address.

### Testing?

Updated tests

### Anything Else?

Renamed the method within announcements `get_users_to_email` -> `get_users_to_email_for_announcements` to avoid confusion

### Review request
@tylerecouture


src/notifications/tasks.py
```diff
@@ -28,8 +30,29 @@ def email_notifications_to_users(root_url):
     # send_email_notification_tenant.delay(root_url)
 
 
+def get_users_to_email_for_notifications():
+    students_to_email = list(
```
tylerecouture, line 34: These to lists of emails seem almost the same as the lists in the other method.  These should come from a single common method rather than repeating ourself, and then apply any extra filters and flattening afterwards.

Instead of duplicating code, can you put these into a method, perhaps with a couple parameters to indicate students, staff, etc?  That would prevent the quadruplication of this code which we are repeating 4 times:

```
.filter(emailaddress__verified=True, emailaddress__primary=True)
                    .exclude(email='')
                    .exclude(email__isnull=True)
```
Actually, why do the two teacher lists exclude email__is_null=True but the two student lists do not?

tylerecouture (changes requested):
Looks great!  But can we DRY this code a bit?  See my suggestion.

src/notifications/tasks.py
yujinyuz: Refactored the code and made it to the `Profile.objects.get_mailing_list`.. that method wasn't used anywhere so I thought that was a good place to put it

tylerecouture (approved):
Nicely done!

PR #1534, opened 2023-11-19
Remove date_type from ExcludeDate

### LINKED ISSUE #1241: Can't add an excluded day to semester through Semester/Update Semester
I tried to add a new excluded day to our semester calendar for the currently active semester running from Sep 7 to Nov 10.  I was not able to do this through the front end byte deck site.  When I entered the date Sept 19, 2022 (Memorial for QEII Holiday) from the calendar on the Update Semester page and hit update I got a "500 Server Error."
image

The new excluded date was successfully added by another staff member with "back end" access.

yujinyuz:
### What?

Fixes #1241 

### Why?

For some reason, one of the tenants could not exclude a date for a Semester using the frontend because the `date_type` is set to null.

The migration scripts might've messed up at some point.

### How?

By removing the `date_type` field from `ExcludeDate`, hopefully it will resolve the issue since the field is not present anymore.

### Testing?

Could not reproduce but testing needs to be done within the tenant where this bug occurs.

### Review request
@tylerecouture


tylerecouture (changes requested):
This is the only place that the DateType model is used, so we might as well remove the model and any other references to it (e.g. Admin) to prevent dangling unused code.

PR #1535, opened 2023-11-20
Automatically remove quest as a prerequisite when archived

### LINKED ISSUE #946: Archived Quests Remain Prerequisites
When a quest is archived it remains a prerequisite for badges. This could lead to users not having future quests become available. 

I'm not sure if we would want it to automatically remove an archived prerequisite from badges, as if it was an accidental archive, that link would be gone. Maybe that's fine though.

Does the badge checking take into account that a quest is already archived?

yujinyuz:
### What?

Closes #946 

### Why?

Archived quests could not be accessed by students so if it is a prerequisite for other quests, you could not actually get that quest because can't complete the archived quest.

### How?

I created a method under `Quest` model called `remove_as_prereq`.

The `self.prereq_prereq.all()`,  returns all objects that use the current object as a prerequisite.

### Testing?

Added test where archiving a quest removes it as a prerequisite

### Review request
@tylerecouture

## Summary by CodeRabbit

- New Features
  • Enhanced prerequisite management for key elements, ensuring flexible and accurate association updates.  
  • Automatically clears prerequisite associations when items are archived for improved consistency.  
  • Adjusts form handling based on user roles, delivering a more tailored and intuitive experience.

- Tests
  • Added tests to verify that archived items update related prerequisites correctly.


src/quest_manager/models.py
```diff
@@ -565,6 +565,13 @@ def expired(self):
         # utilize existing code in QuestQuerySet method not_expired()
         return not Quest.objects.filter(id=self.id).not_expired().exists()
 
+    def remove_as_prereq(self):
+        """
+        When a quest is archived, remove it from all prerequisite lists
+        """
+        if self.archived:
+            self.prereq_prereq.all().delete()
```
tylerecouture, line 573: How will this handle prerequisites that have an alternate OR prereq object?

i.e. Prereq objects that have "quest1 OR quest2", how does it handle this?

tylerecouture: Also, I feel like this should be a method in the IsAPrereq mixin, and it doesn't seem necessary to check archived here, that's already being checked before the method is called, and in the future we may want to call this method for other reasons (and on other objects that implement IsAPrereq)

src/quest_manager/views.py
```diff
@@ -160,6 +160,10 @@ def form_valid(self, form):
         # Save the object so we can add prereqs to it
         super_response = super().form_valid(form)
         self.set_new_prereqs(form)
+
+        if form.cleaned_data.get("archived") is True:
```
tylerecouture, line 164: Perhaps a post-save signal is a cleaner way to do this?  Having this in a view feels clunky to me. Also, this will miss instances where archived is set to true via django admin.

We need to ensure this works for bulk archiving too, there is a Django admin action for this.

tylerecouture (changes requested):

src/quest_manager/models.py
yujinyuz: > Prereq objects that have "quest1 OR quest2", how does it handle this?

Can you give me another example? I'm not sure I understand what you mean here..

tylerecouture: Here's an example of a quest with an alternate prerequisite.  In the example its a Rank OR a Badge, but they could both be quests.

image

Which can be added in the prereq form as an "Alternate"

image


THe docstring for the class I think describes it fairly well?

```
class Prereq(IsAPrereqMixin, models.Model):
    """
    A Prereq object indicates some conditions (prerequisites) that must be met before gaining access to something else.

    parent_object: The parent object is the thing with restricted access (for example, a quest, or a badge).
     Access to it is granted once the conditions are met.

    prereq_object: This is the main condition that has to be met for a user to gain access to the parent object.
     How that condition is met is determined by the Class's implementation of `condition_met_as_prerequisite()` from the
     IsAPrereqMixin.  This object could be any model that implements the mixin (quest, badge, grade, course, etc)

    or_prereq_object: This is an ALTERNATE condition that could be met instead of the condition laid out by prereq_
     object.

    Simple example:  Imagine a quest that a student only gains access to if they are in a grade 10 course:

        Prereq <
            parent_object = quest_manager.models.Quest object <"Some Quest">
            prereq_object = courses.models.Grade <"Grade 10">
        >

        Given a specific user, when the querying of available quests gets to this quest, it will find this Prereq and
          call condition_met() on it. If condition_met() returns True (supposedly, because the student has a Grade 10
          course) then this Prereq (Quest object <"Some Quest"> will not be filtered out of the user's available quests

          See condition_met() for further details.


    Links two (or three) objects (of any registered content type) in a prerequisite relationship.
    Note that a Prereq can itself be a prereq (in combination with AND OR NOT, this allows for arbitrary complexity)
    parent: The owner of the prerequisite. Generally, for the parent to become 'active' the prereq_object conditions
        must be met.  'active' can mean whatever the user of this app wants it to mean.  It's a generic trigger.
    prereq: The object whose conditions must be met in order for the parent to become 'active', which optional
        additional conditions NOT and a number to indicate how many times the condition much be met.
    or_prereq: An optional alternate object who's condition can me met instead of prereq: prereq OR or_prereq
    """
```

src/quest_manager/views.py
yujinyuz: done

tylerecouture:
@coderabbitai review

tylerecouture:
@coderabbitai review

tylerecouture:
@9921622 Can you review the issue connected to this, and this Pull Request, and test it all out and make sure it works?

PR #1536, opened 2023-11-23
Modular Side Menu

### LINKED ISSUE #1437: Modular main sidebar menu

yujinyuz:
### What?

Modularize side menu items.

Closes #1437

### Why?

These changes will let the users customize their own decks

### How?

This includes a migration for adding new fields to the MenuItem model, changes to the menu item delete view. A method under the model manager is also added for retrieving or creating the default menu items.

### Testing?

I've added tests for creating default side menu items, and also prevents users from deleting a default menu item.

https://github.com/bytedeck/bytedeck/assets/10972027/fefbb8ed-4628-4b8b-a459-c1a69351a99e

** Updated **

https://github.com/bytedeck/bytedeck/assets/10972027/a51082b3-8519-4e45-8abc-92fadb9a000d

### Anything Else?

- I also modified the update menu item to include a preview for the fa icon
- I added a new profile url `/profiles/own/` because the current one requires a primary key and I don't think you can include that within the `url` of the MenuItem
- 

### Review request
@tylerecouture


tylerecouture:
@coderabbitai review

PR #1568, opened 2024-02-05
added id to line 55

Happyashbunny:
I was asked to add a confirmation step to prevent mistakenly ending a semester 

Upon looking into the file i noticed that there was already a js script tag at the very bottom with the onclick property on a tag 
id-ed "endsemester" however there was no tag that was id-ed that particular name. So I just id-ed the respective anchor tag mentioned as "endsemester".  

Another approach to this would be to use the "onclick()" function directly on the responsible anchor tag and use the "confirm" function within it. I am also ready to try out this approach. 

I apologize for any oversight that I may have made. I had no intention of wasting anybody's time.

### DIFF
--- a/src/courses/templates/courses/semester_list.html
+++ b/src/courses/templates/courses/semester_list.html
@@ -77,7 +77,7 @@
 
 <script>
   $('body').on('click', '#end-semester', function(e) {
-      if(!confirm("Are you sure?")) {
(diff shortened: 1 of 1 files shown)


tylerecouture (approved):
Thank you for finding this!  Your contribution is hugely appreciated!

PR #1585, opened 2024-04-13
Refactor docker-compose files to fix #940.

### LINKED ISSUE #940: Migration timing in containers can cause errors
When running `docker-compose up` the celery container will be ready before the web container.

However, migrations are run by the web container:

```
  web:
    stdin_open: true
    tty: true   # Enabled together with stdin_open so it won't raise an error when we are in pdb mode. Attach using `docker attach $(docker-compose ps -q web)`
    environment:
      POSTGRES_HOST: db  # takes precendent over .env file
      REDIS_HOST: redis  # takes precendent over .env file
    command: bash -c "cd /app/src/ &&
                      python manage.py migrate_schemas --shared &&
                      python manage.py migrate_schemas --executor=multiprocessing &&
                      python manage.py runserver 0.0.0.0:8000"   # no need for uwsgi in development, this is easier
```
which means celery might be trying to access columns or tables that don't exist yet. For example:
```
celery container started before the web container when running docker-compose up but the migrations are only run in the web container
```
Probably same thing with the celery-beat container.

What's the solution?

joshwizzy:
### What?
Refactor docker-compose files to start celery and celery-beat after django migrations 
### Why?
When running `docker-compose up` the celery container will be ready before the web container.
However, migrations are run by the web container:
which means celery might be trying to access columns or tables that don't exist yet. 

### How?
* Introduce separate Dockerfiles for development and production builds.
* Update service configurations to use custom pass through entrypoints.
* Entry points for celery and celery-beat wait for django app to be reachable

## Summary by CodeRabbit

- **New Features**
  - Enhanced Docker setup to include specific entrypoints and commands for services like web, celery, and celery-beat.
  - Added installations and scripts for improved service management in Docker environments.
  
- **Bug Fixes**
  - Corrected Docker compose configurations and indentation issues for better reliability and clarity.

- **Documentation**
  - Updated comments and documentation within Docker configuration files for clearer understanding and maintenance.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,4 +1,4 @@
-version: '2.2'
+version: "2.2"
 
 ###############################################################################
 #
@@ -25,7 +25,7 @@ services:
       timeout: 5s
       retries: 5
     ports:
-    - 5432:5432  # Expose port so app can connect to db when developing locally with `python src/manage.py [runserver|test src]
+      - 5432:5432 # Expose port so app can connect to db when developing locally with `python src/manage.py [runserver|test src]
 
   redis:
     image: "redis:5.0-alpine"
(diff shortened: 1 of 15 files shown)


tylerecouture:
This is a large change with lots of duplication and little explanation in the PR what is change and why it's needed.  Might be great, but unfortunately is too difficult to review due to lack of documentation.

PR #1593, opened 2024-05-29
Fix: Tag detail view shows duplicate ordinal for repeatable quests; Closes #1210

### LINKED ISSUE #1210: Tag detail view shows duplicate ordinal for repeatable quests
image

9921622:
### What?
Fixed an issue where ordinality appears twice in student TagDetailView but only for repeat quests.

### Why?
This issue is a formatting error. Ordinal values showing up twice is unexpected behaviour
See #1210

### How?
The issue was that ordinal value is added to the quest name if its a repeat quest. While the DetailView does the same thing in the html via template tags.

The solution is to use the ```QuestSubmission.__str__``` value by itself
```
Original code
+ Quest (1)
+ Quest (2) (2)  # Quest (2) is the __str__ value of the quest submission
Solution using __str__ by itself
+ Quest  # ordinality does not show up but is expected behaviour
+ Quest (2)
```

I fixed it by removing the html template tag that adds the extra ordinal values. Hence removing the extra "(2)", "(3)", etc.
image

### Testing?
Added test case for checking correct ordinality.
image

screen shot of what the test case looks like
image

### Anything Else?
None

### Review request
@tylerecouture

### DIFF
--- a/src/tags/templates/tags/snippets/related-xp-item.html
+++ b/src/tags/templates/tags/snippets/related-xp-item.html
@@ -17,7 +17,6 @@
                         <li>
                             <a href="{{ submission.quest.get_absolute_url }}">
                                 {{ submission }}
-                                {% if submission.is_multiple %}({{submission.ordinal}}){% endif %}
                                 ({% firstof submission.xp_requested submission.quest.xp %} XP)
                             </a>
                         </li>
--- a/src/tags/tests/test_views.py
+++ b/src/tags/tests/test_views.py
@@ -200,6 +200,47 @@ class TagCRUDViewTests(ViewTestUtilsMixin, TenantTestCase):
         self.assertContains(response, self.tag.name)
         self.assertContains(response, self.test_student.username)
 
+    def test_DetailView__student_view_ordinal(self):
+        ''' test if detail view list the ordinal quests correctly '''
+        num_subs = 5
+        unrelated_quest = baker.make('quest_manager.quest', xp=1)
+        repeatable_quest = baker.make('quest_manager.quest', name='Repeatable Quest', max_repeats=-1, xp=1)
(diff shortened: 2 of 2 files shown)


9921622:
@tylerecouture review

tylerecouture:
Description WHY doesn't include link to issue.  Please link.

tylerecouture (changes requested):
1. Please link to issue in WHY section of PR description
2. Please expand on this, It's not clear to me what you are saying.  Also are you pretty sure?  if you are only pretty sure, please verify.

9921622:
alright. updated the wording and linked the issue in the "WHY" lmk if it needs more clarity

tylerecouture:
@9921622 Can you clarify what this means? "...and its the first one."  The first what?

tylerecouture:
@9921622 also, are you not able to track down the issue?  Why are you not certain that this is the problem?  If you aren't certain it's the problem, then we can't be certain this will fix it... haha

9921622:
i know the issue is the ordinal naming scheme. 
Here's how it's implemented:
```QuestSubmission```
image
```QuestSubmissionManager ```
image

But it doesn't look like a bug. just a naming convention for multiple submissions of the same quest.

9921622:
also updated the wording for the "...and its the first one."

9921622:
@tylerecouture

9921622:
updated the pr so DetailView follows the same ordinal naming conventions as Quests Tab in ProfileView
image

9921622:
@tylerecouture

9921622:
wait i now have 2 commits

9921622:
@tylerecouture since you added a merge branch can i assume its good to go?

PR #1595, opened 2024-05-29
Feat: Intro tag to starter quests and badges; Closes #1207

9921622:
Please ensure you are familiar with our Pull Request Description expectations described here: https://www.pullrequest.com/blog/writing-a-great-pull-request-description/
### What?
Added new tag "Intro" to"Welcome to ByteDeck!" quest + all quests in the orientation campaign, and "Bytedeck Proficiency" badge

### Why?
see #1207

### How?
Manually added the tag to the quests and badges.

image
image

### Testing?
Added 2 new test cases where it tests to see if intro tag is created on initialization of a new tenant.
image

### Screenshots (if front end is affected)
None
### Anything Else?
None
### Review request
@tylerecouture

PR #1598, opened 2024-05-30
Fix: Failing build because of tenant test case

9921622:
What?
Small issue with the test case causing builds to fail

Why?
Causing builds to fail because of https://github.com/bytedeck/bytedeck/pull/1595

How?
was filtering for "Intro" instead of "intro"
image

Testing?
Screenshots (if front end is affected)
Anything Else?
Review request
@tylerecouture


9921622:
@tylerecouture

PR #1602, opened 2024-05-31
Feat: Added Semester name; Closes #1565

### LINKED ISSUE #1565: Semester naming is bad
Currently it defaults to "mmm-YYYY"  But if you create two semesters in the same month, you can't tell them apart.

We need a name field that defaults to mmm-YYYY (or returns mmm-YYYY through a get_name method). Can be blank.

Unique = False so it can be left blank.

9921622:
### What?
Added a new char field to Semester called "name". Name field now shows up in forms, admin, and listview

### Why?
See #1565

### How?
Added name field to model. Also modified the __str__ to use name but if its empty, use first date.
Name cant really be empty in the front view, but can in tests. Without the "or" tests will fail
image

Added the field to form fields and helper layout.
Also changed the save function so it defaults to first day if name is empty
image
(did not add anything to admin as it was already using __str__ and i modified that in Semester model)

### Testing?
edited existing tests to check for semester name
```test_SemesterCreate__without_ExcludedDates__view``` and ```test_SemesterList_view```
image
There was a conflict in ```test_semester_creation``` but i removed it since it was testing for something that just got removed
image

ListView
image
UpdateView fullscreen (but applicable to anything that uses Semester form)
image
squished
image
delete view
image

Admin View
image

### Review request
@tylerecouture

### CODE REVIEW

File: src/courses/migrations/0028_semester_name.py [hunk 1]
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.25 on 2024-05-31 17:12
+
+import courses.models
```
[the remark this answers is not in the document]
9921622: this is the migration thing. unnecessary

9921622:
@tylerecouture review

src/courses/forms.py
```diff
@@ -123,6 +123,7 @@ def __init__(self, *args, **kwargs):
         self.helper.form_tag = False
 
         self.helper.layout = Layout(
+            Div('name', css_class='col-xs-6 col-sm-4',),
```
tylerecouture, line 126: In Bootstrap, there are 12 columns, so if you add a new column and want it to be equal, all three should be xs-4.  I don't know what the -sm-4 is doing there, I don't think it's doing anything an you can remove from all three.  However please check responsive behaviour by making browser wider (full screen) and narrower (phone-like) to confirm doesn't break the layout.  Reference: https://getbootstrap.com/docs/3.4/css/#grid

src/courses/models.py
```diff
@@ -200,11 +200,17 @@ def default_end_date():
     return date.today() + timedelta(days=135)
 
 
+def formatted_date():
+    return date.today().strftime('%B-%Y')
```
tylerecouture, line 204: This should not be today(), it should be the first_day (as it currently is).  It might need to be set in a post save (creation) signal

src/courses/models.py
```diff
@@ -248,6 +254,9 @@ def num_days(self, upto_today=False):
         # We want to return an int because a numpty.int64 is not JSON serializable as of the moment
         return count.item() if hasattr(count, 'item') else count
 
+    def get_name(self):
```
tylerecouture, line 257: What the purpose of a getter for name?  Just access it directly.

tylerecouture (changes requested):
See comments

src/courses/models.py
9921622: removed

src/courses/forms.py
9921622: got it

src/courses/models.py
tylerecouture: Correction.  The field can be left blank, and if it's blank, then __str__ returns MMM-YY, if it's not blank, then __str__ returns name field.

tylerecouture: Where MMM-YY is form the start date, as it currently does.

tylerecouture: We don't need to set a default on the name field.

9921622: ah ok

9921622: right now i have the listview showing its name. im guessing I also show first day as its name if left emtpy?

tylerecouture: Good point. Yes.  So just put `__str__` in that column.  i.e. the semester object itself (which will return it's __str__ when cast to a string by the template)

9921622: ill just commit + push for now

9921622: oof just commited. ill change it

src/courses/forms.py
```diff
@@ -117,16 +117,31 @@ class Meta:
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
+        self.fields['name'].help_text = 'If empty, will instead use First day as it\'s name'
```
tylerecouture, line 120: This should be in the model field's help_text, then the form will use it by default.  Is there a specific reason to put it hear instead?

src/courses/forms.py
```diff
@@ -117,16 +117,31 @@ class Meta:
...
+        self.fields['name'].required = False
```
tylerecouture, line 121: This should also just get picked up from the model field's `required=False` parameter

src/courses/forms.py
```diff
@@ -117,16 +117,31 @@ class Meta:
...
 
         self.helper.layout = Layout(
-            Div('first_day', css_class='col-xs-6 col-sm-4',),
-            Div('last_day', css_class='col-xs-6 col-sm-4',),
+            Div('name', css_class='col-xs-4',),
+            Div('first_day', css_class='col-xs-4',),
+            Div('last_day', css_class='col-xs-4',),
         )
 
+    def save(self, commit=True):
```
tylerecouture, line 134: Not clear to me why this is required...

src/courses/models.py
```diff
@@ -205,6 +205,8 @@ class Semester(models.Model):
     QUEST_AWAITING_APPROVAL = -2
     STUDENTS_WITH_NEGATIVE_XP = -3
 
+    name = models.CharField(null=True, unique=False, max_length=50)
```
tylerecouture, line 208: null=True, blank=False <--- that might be why you needed that extra stuff in the form?

tylerecouture: null <--- database can be emtpy
blank <--- form can be empty

tylerecouture (changes requested):
I think you're missing `blank=True` in your model field, causing you to have all the uneccessary extra code in your form.

src/courses/forms.py
9921622: changed

src/courses/forms.py
9921622: changed

src/courses/models.py
9921622: changed

src/courses/forms.py
9921622: mixed up blank=True and null=True. so the form was giving me an error everytime I updated/created with an empty name field

src/courses/models.py
```diff
@@ -205,6 +205,11 @@ class Semester(models.Model):
     QUEST_AWAITING_APPROVAL = -2
     STUDENTS_WITH_NEGATIVE_XP = -3
 
+    name = models.CharField(
+        blank=True, unique=False, max_length=50,
+        help_text="If empty, will instead use First day as it's name"
```
tylerecouture, line 210: "If left blank, the semester will display it's First Day as a default name, in the form: mmm-YYYY."

tylerecouture (changes requested):
One last nitpick.

PR #1613, opened 2024-06-06
Fix: Crash at 11 PM; Closes #1327

### LINKED ISSUE #1327: Date field unexpected behaviour at midnight in app default TZ
This bug can be observed in tests, if run at midnight in default TZ (settings.py: `TIME_ZONE = 'America/Vancouver'`)

See PR #1326.

I suspect one of these tests fails 1 hour before midnight to midnight, and the other fails from midnight to one hour after:

**QuestTestModel.test_active()**

```
        # create and test a quest that won't be available until one hour later in the same day
        baker.make(
            Quest, name="future-time-quest", date_available=timezone.localtime(),
            time_available=(timezone.localtime() + timezone.timedelta(hours=1)))
        self.assertEqual(Quest.objects.get(name="future-time-quest").active, False)
```

and

```
        # create and test a quest that's expired one hour ago
        baker.make(Quest, name="past-time-quest", time_expired=(timezone.localtime() - timezone.timedelta(hours=1)))
```

So...we need a test that specifically covers midnight for these two (freezetime?) and fails on existing code, and then a fix.These should be placed in their own test method.

9921622:
### What?
Fixed a test that would crash at 11:00 PM.

### Why?
See #1327

### How?
Constrained the test that crashes with ```freeze_time```.

only occuring at 11:00 PM because...
(using 24 hour clock for now)

At 23:00 when creating a ```Quest``` object.
```
q = baker.make(
	Quest,
	date_available=timezone.localdate(),
	time_available=(timezone.localtime() + timezone.timedelta(hours=1)).time()
)
self.assertFalse(q.active)
```

```time_available=(timezone.localtime() + timezone.timedelta(hours=1)).time()``` is equivalent to ```23:00 + 1 hour``` or ```0:00```

Then causes this specific check in ```Quest.active``` to fail (if statement false)
```
# an XPItem object is inactive on the day it's made available if its availability time is in the future
	if self.date_available == date_now and self.time_available > time_now:
		return False
```

inciting variables
```
time_available = 0:00  # current time + 1 hour
time_now = 23:00  # current time

(time_available > time_now) is false 
(0:00 > 23:00) is false
```

So this is more of a test_case error than something with the models. Hence why I only used ```freeze_time``` 

### Testing?
Only manually tested this using ```freeze_time ``` and changing my system clock
image

Before fix at 11:00 PM (using develop branch):
image
After fix:
image

### Review request
@tylerecouture

## Summary by CodeRabbit

- **Tests**
  - Improved test reliability for quest availability by using `freeze_time` to handle time calculations accurately.

### DIFF
--- a/src/quest_manager/tests/test_models.py
+++ b/src/quest_manager/tests/test_models.py
@@ -170,12 +170,15 @@ class QuestTestModel(TenantTestCase):
         self.assertFalse(q.active)
 
         # create and test a quest that won't be available until one hour later in the same day
-        q = baker.make(
-            Quest,
-            date_available=timezone.localdate(),
-            time_available=(timezone.localtime() + timezone.timedelta(hours=1)).time()
-        )
-        self.assertFalse(q.active)
+        # using freeze_time to prevent time_available from looping from 23:00 to 0:00 when adding 1 hour
+        dt = datetime.datetime(2024, 1, 1, 6, tzinfo=timezone.get_current_timezone())
+        with freeze_time(dt):
+            q = baker.make(
+                Quest,
+                date_available=timezone.localdate(),
+                time_available=(timezone.localtime() + timezone.timedelta(hours=1)).time()
+            )
+            self.assertFalse(q.active)
 
         dt = timezone.get_current_timezone().localize(timezone.datetime(2023, 6, 13, 12, 0, 0))
         with freeze_time(dt):


9921622:
@tylerecouture review

tylerecouture (commented):
Amazing.  Thank you!!

tylerecouture (approved):

tylerecouture:
@9921622 That explains why fails occur 11pm to midnight.  Lets rerun tests tonight after 11pm for a sanity check!

9921622:
yep ill do that and post a screenshot on the issue itself

PR #1625, opened 2024-06-17
WIP Feat: flake8 for git action, pre-commit, flake8 uses the same config; Closes #1623

### LINKED ISSUE #1623: Synchronize Flake8 checks
We currently have 3 different possible Flake8 checks that occur.  All three may have different configurations, which is confusing, because a locally run `flake8 src` might not pick up on an issue that the pre-commit hook does, or the GitHub Action does.

Our three Flake8 checks are:

* run in local development with: `flake8 src` -- config is at `.flake8` I THINK?! Does it actually use that?
* run when you make a local commit (via pre-commit hook if they were installed per readme)  -- config is in `.pre-commit-config.yml` does it also use `.flake8` ??
* run by GitHub Actions on a push or pull request. -- config is at `.github/workflows/lint.yml`

I don't know if any of these are using the same configurations, which mean some runs are flagging issues that others aren't.

Need to get all these processes synchronized to use the same config.  if we can't get them all to recognize the `.flake8` file then we may just need to hardcode the same config settings into three different places!

9921622:
### What?
Made sure all 3 flake8 versions (flake8, precommit flake8, git action flake8) use the same config file

### Why?
See https://github.com/bytedeck/bytedeck/issues/1623

### How?
Since none of the flake8 versions (except the base one) use the `.flake8` file as a configuration. i've added `--config=.flake8`
for precommit file and git actions file.

### Testing?

temporarily added `count = True` to the `.flake8` file

#### flake8
manually tested using `flake8` command and seeing if the count was shown
image

#### precommit
added a minor change and used `git commit -am "f"` and checked if precommit shown the number of errors
image

#### git actions
since I cant push a pr without triggering precommit. i've commited one without errors and checked if the output has 0. (for 0 errors)
image

### Review request
@tylerecouture

## Summary by CodeRabbit

- **Chores**
  - Updated linting configuration to use a custom `.flake8` file.
  - Enhanced pre-commit setup to include a custom linting configuration.


9921622:
@tylerecouture review

tylerecouture:
Thank you! That seems pretty easy, lol. I have no idea why we didn't already have this.

Excellent demo in the PR description showing that it's working for all of them!

PR #1631, opened 2024-06-24
Feat: List of related maps in Quest and Badge view; Closes #273

### LINKED ISSUE #273: Link quests to maps they they appear on
Student suggestion:
> being able to find your current quest in the quest maps instead of having to search through it."

This should not be super hard, but it might be difficult?
Get a queryset of Map objects filtered for any maps that contain... what?  What is the fast easiest thing to search on the Map model (djcytoscape)?  If we can figure out how to get this queryset in a single step (so as not to slow down the page load), then we can easily add a list of those maps with links in the Quest header info where the tags and prereqs are!

9921622:
### What?
Added a list of related maps in Quest and Badge views

### Why?
See #273

### How?

Added a new method in `CytoScapeManager`, called `get_related_map`. Gets all related maps to "ContentType"/selector id.
Added the query to the context of Quest and Badge detail views.
Formatted the html similar to tags list

### Testing?
added testing for `get_related_maps` which tests for correct query when quest is:
+ part of no maps
+ part of exactly one map
+ part of three maps

Quest
image
Badge
image

### Anything Else?
Noticed the  `help_text` for `CytoElement.selector_id` was slighly incorrect. Saying Selector id is `model-#` instead of `model: #`

I'll probably make an extra pr to fix that. But it might be too inconsequential to make a migration for

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Introduced a "Maps" section in badge and quest detail views to display related maps.
  - Added conditional rendering of maps based on their count.

- **Tests**
  - Added tests to ensure maps are correctly associated with quests.

- **Bug Fixes**
  - None

- **Documentation**
  - None

### DIFF
--- a/src/badges/templates/badges/detail.html
+++ b/src/badges/templates/badges/detail.html
@@ -34,6 +34,7 @@
         </div>
         <div class="col-sm-6">
           <ul class="list-unstyled">
+            <li>Maps: {% include 'djcytoscape/snippets/map-list.html' %}</li>
             {% if not request.user.is_staff %}
             <li>{% tag_name %}s: {% include 'tags/snippets/tag-list.html' with object=badge user_obj=request.user %}<br></li>
             {% else %}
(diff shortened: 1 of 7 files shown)


tylerecouture (commented):
One question: what's it look like if it is not on a map?  Does it say "None"?

9921622:
@tylerecouture yeah, its using the same format as prereq and tags.
image

PR #1632, opened 2024-06-26
1476 Submission quest feature merge resolve

9921622:
Please ensure you are familiar with our Pull Request Description expectations described here: https://www.pullrequest.com/blog/writing-a-great-pull-request-description/
### What?
There were some issues with #1476 and a couple of bugs.

### Why?
See #1627

### How?
Rebased with develop branch and resolved conflicts. 

There was an issue with quick reply form where after the user has submitted, and makes a comment via quick reply. They get a 404.

Fixed another issue where admin was creating a draft comment for a user when going to submission view (and making the owner of the draft comment admin)

#### manual tests
+ user comment submission 
+ user comment after submission (quick reply)
+ admin comment before approving (quick reply)
+ admin approve comment (quick reply)
+ user comment after approve (quick reply)
+ user comment using submission view
+ user comment using submission view (2nd time)
+ test if draft was deleted when submission is deleted
+ manually test the migration from draft_text to draft_comment in submission view 

#### automated tests
+ checks the functionality of multiple comments using standard and quick reply view

### Anything Else?
added a pre delete signal to QuestSubmission deleting any comments linked to it when QuestSubmission is deleted.
Also included a test for it

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Added the `draft_comment` field to `QuestSubmission` to store submission comments.

- **Bug Fixes**
  - Updated `mark_completed` method to manage the new `draft_comment` field properly.

- **Tests**
  - Added tests to check the association of comments with `QuestSubmission`.
  - Introduced tests to validate the `pre_delete` signal functionality for `QuestSubmission`.

- **Chores**
  - Implemented a signal handler to delete comments linked to a `QuestSubmission` upon its deletion.

Issue #1630, opened 2024-06-23
Implement full_clean() method before using save() on models outside of a ModelForm


tylerecouture:
There appear to be issues around the save() and full_clean() methods that I only recently discovered thanks to @rabbitai and @9921622.

The problem is described here:
https://jamescooke.info/djangos-model-save-vs-full_clean.html

As of 3.2 (this project's django version as of creation of this issue), the model save() method does not automatically call full_clean() when manually saving objects outside of a ModelForm:
https://docs.djangoproject.com/en/3.2/ref/models/instances/#django.db.models.Model.full_clean

> Note that full_clean() will not be called automatically when you call your model’s [save()](https://docs.djangoproject.com/en/3.2/ref/models/instances/#django.db.models.Model.save) method. You’ll need to call it manually when you want to run one-step model validation for your own manually created models. For example:

We create a lot of objects this way, in code.  Each of these manually created objects should call full_clean before hand.

Here's a suggestion to update our codebase to include `full_clean()`

1. Do a little more research first! Get a decent grip on the issue!
2. Pick an app (e.g. announcements) within the project, and do a search for "save()".  We may also need to consider .save(commit=False)? I can remember, that might just be for ModelForms
3.  anytime you find a model.save() call, add model_object.full_clean() before it.
4. Run the app's tests, and make sure nothing fails.
5. If something fails, it could mean we have some bad code that is trying to save incomplete/unvalidated model instance.  REPORT the failing part back here for discussion, and we can determine what to do on a case by case.
6. Once the app is fixed PR the changes, and move on to the next app!

I'm hoping this might solve some of our rare/hard to reproduce data integrity issues that crop up from time to time.

Note that because our production data may have some integrity issues because of this, when fixing, we need to be aware of possible bad data in production.  This means we may need to log or email exceptions.  Alternately, we could run a project wide scan, looping through all the tenants and models and data and full_clean() all objects within.  That will find integrity errors so we can manually fix them - @9921622 please create a new issue for this: create an interactive management method to loop through each tenant, and then each model and all it's data, full_clean() and save() in a try block and have some way of reporting errors as found so they can be fixed.

9921622:
- [ ] Announcements
- [x] Badges
- [ ]

PR #1637, opened 2024-06-28
Chore: full clean comments app

9921622:
### What?
Added full_clean() before any save in comments app

### Why?
See #1630

### How?
Added full_clean() before any model save()

Because of how https://github.com/bytedeck/bytedeck/pull/1632 changed how draft comments work, draft_comments initially need text to be none. I've added `blank=True` to comment model's text field. To not break full_clean with a validation error.

### Testing?
Ran the automated testing for comments app.

Ran full tests on comments app
image

### Anything Else?
Noticed `CommentManager.create_comment` has `comment.save` twice. It looks unnecessary, and comment does not have a custom save function. But removing it causes a couple of tests to fail. So i added a full clean before that instead of removing it

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Enhanced comment submission flexibility by allowing optional text input.
  
- **Bug Fixes**
  - Improved data validation when creating, flagging, or unflagging comments to enhance data integrity.
  
- **Tests**
  - Updated test cases to reflect changes in comment behavior regarding text representation and initialization.


src/comments/models.py
```diff
@@ -60,10 +60,12 @@ def create_comment(self, user=None, text=None, path=None, target=None, parent=No
         if parent is not None:
             comment.parent = parent
 
+        comment.full_clean()
         comment.save(using=self._db)
 
         # add anchor target to Comment path now that id assigned when saved
         comment.path += "#comment-" + str(comment.id)
+        comment.full_clean()
```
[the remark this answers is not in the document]
9921622: make this a committable suggestion

9921622:
@tylerecouture review

src/comments/models.py
```diff
@@ -60,10 +60,12 @@ def create_comment(self, user=None, text=None, path=None, target=None, parent=No
         if parent is not None:
             comment.parent = parent
 
+        comment.full_clean()
         comment.save(using=self._db)
```
tylerecouture, line 64: Why are we saving twice?  Can this one be removed?

src/comments/models.py
```diff
@@ -130,7 +132,7 @@ class Comment(models.Model):
     user = models.ForeignKey(settings.AUTH_USER_MODEL, on_delete=models.CASCADE)
     parent = models.ForeignKey("self", null=True, blank=True, on_delete=models.SET_NULL)
     path = models.CharField(max_length=350)
-    text = models.TextField()
+    text = models.TextField(blank=True)
```
tylerecouture, line 135: why blank=True but not null=True?

tylerecouture (changes requested):

src/comments/models.py
9921622: nope. the first save is to initialize the object in the db (in this case its to get id.) Then the next line of code is to calculate the path which uses id.

ill remove the first `full_clean()` since its redundant

src/comments/models.py
9921622: added `blank=True, null=True`

tylerecouture:
Thanks for this, @9921622 — the underlying work (adding `full_clean()` to the comments app for #1630) is still valuable, so I've reimplemented it fresh against current `develop` in #2006 and am closing this one.

The new PR keeps your core approach (full_clean before the saves in `create_comment`/`flag`/`unflag`, and allowing blank comment text so empty draft comments validate), with one difference: it makes `text` `blank=True` only rather than `null=True`, since the code stores `""` and never `None` — that avoids a nullable text column and leaves `__str__`/the existing tests unchanged. Credit to you for the original diagnosis and fix. Closing in favour of #2006.

— Claude Code (on behalf of @tylerecouture)

---

tylerecouture:
Refactored in #2006

PR #1654, opened 2024-07-18
Tweak: refactored approvals and quest_list to use enums for tabs; Closes #1150

### LINKED ISSUE #1150: Find and refactor templates that look at the url for setting active buttons
use an enum instead of looking at ``request.PATH_INFO``/URL to determine what page youre on

Example is from ``profile_manager.views.py``
make an enum and class variabe for said enum and pass it to the context
```
class ViewTypes:
    """ enum for ProfileList and its descendants """
    LIST = 0
    CURRENT = 1
    STAFF = 2
    INACTIVE = 3

class ProfileList(NonPublicOnlyViewMixin, UserPassesTestMixin, ListView):
    ...
    template_name = 'profile_manager/profile_list.html'
    view_type = ViewTypes.LIST  
    ...
    ...
        context = super().get_context_data(**kwargs)
        context['VIEW_TYPES'] = ViewTypes
        context['view_type'] = self.view_type
        return context

class ProfileListCurrent(ProfileList):
    ...
    view_type = ViewTypes.LIST  
```
in ``profile_list.html``
in the associated template use the class enum variable and enum itself to determine which enum the view has and apply your logic
```
class="btn {% if view_type == VIEW_TYPES.CURRENT %}btn-primary{% else %}btn-default{% endif %}">Current Semester</a>
```
this template logic checks if the current view's enum is equal to ``VIEW_TYPES.CURRENT`` and sets the active button accordingly

9921622:
### What?
Ive made some views use enums for its tabs, instead of defining a separate boolean variable for each tab in context. 

### Why?
See #1150

### How?
Found only two views that use multiple variables instead of an enum: `approvals` and `quest_list`.
Changed the view to use enums instead.

### Testing?
added two tests that check if the expected enum is used when a tab is active.

### Anything Else?
Unsure if  `approvals` needs the enums as it uses a list of dicts that store tab information 

Fixed an bug with library tab where it had accordion class attached to it. triggering `ajax_accordion` which caused this error.
image
An alternative solution if we want to preserve accordion functionality for library quests is to make an ajax view to call for library quests, like this
image

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Enhanced quest display logic by transitioning to a dynamic `view_type` system in templates, improving user interaction.
  - Introduced a structured enumeration for quest list view tabs, streamlining navigation.

- **Refactor**
  - Renamed `ViewTypes` to `ProfileViewTypes` for clearer documentation and usage in profile management views.
  - Simplified tab display logic in quest templates by using `view_type` constants.

- **Tests**
  - Updated test cases to use the new `view_type` logic for more accurate context validation.
  - Added new test methods to ensure each tab type can be activated correctly.

### DIFF
--- a/src/library/templates/library/library_quests.html
+++ b/src/library/templates/library/library_quests.html
@@ -1,1 +1,1 @@
-{% extends "quest_manager/quests.html" %}
+{% include "quest_manager/quests.html" with view_type=-1%}
--- a/src/library/templates/library/tab_library_quests.html
+++ b/src/library/templates/library/tab_library_quests.html
@@ -22,7 +22,7 @@
     </thead>
     <tbody class="panel-group panel-group-packed" role="tablist" aria-multiselectable="true">
       {% for q in library_quests %}
-        <tr class="accordian-trigger panel-title
+        <tr class="panel-title
           {% if request.user.is_staff %}
             {% if not q.visible_to_students %}danger
             {% elif q.is_expired %}warning
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -387,6 +387,17 @@ def get_minutes_to_complete_list(quest):
     return minutes_list
 
 
+class QuestListViewTabTypes:
+    """ enum for quest_list's tabs.
+    Note: using enum.auto() will not work as django template tags cant properly define its value.
+    """
+    AVAILABLE = 0
+    IN_PROGRESS = 1
+    COMPLETED = 2
+    PAST = 3
+    DRAFT = 4
+
+
 @non_public_only_view
 @login_required
 def quest_list(request, quest_id=None, template="quest_manager/quests.html"):
@@ -396,11 +407,7 @@ def quest_list(request, quest_id=None, template="quest_manager/quests.html"):
     past_submissions = []
     draft_quests = []
 
-    available_tab_active = False
-    in_progress_tab_active = False
-    completed_tab_active = False
-    past_tab_active = False
-    drafts_tab_active = False
+    view_type = QuestListViewTabTypes.AVAILABLE
     remove_hidden = True
 
     active_quest_id = 0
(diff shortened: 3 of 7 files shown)


9921622:
noticed a bug with library going to be WIP for now

src/library/templates/library/library_quests.html
```diff
@@ -1 +1 @@
-{% extends "quest_manager/quests.html" %}
+{% include "quest_manager/quests.html" with view_type=-1%}
```
tylerecouture, line 1: Why is this switched to `include` instead of `extend`? That seems significant

src/library/templates/library/tab_library_quests.html
```diff
@@ -22,7 +22,7 @@
     </thead>
     <tbody class="panel-group panel-group-packed" role="tablist" aria-multiselectable="true">
       {% for q in library_quests %}
-        <tr class="accordian-trigger panel-title
+        <tr class="panel-title
```
tylerecouture, line 25: thank you!

src/quest_manager/views.py
```diff
@@ -387,6 +387,17 @@ def get_minutes_to_complete_list(quest):
     return minutes_list
 
 
+class QuestListViewTabTypes:
```
tylerecouture, line 390: We have:
ProfileViewTypes
QuestListViewTabTypes
ApprovalsTabTypes

To be more consistent in the naming, let's make it `ApprovalsViewTabTypes` <-- add the View, that adds some good context when you are seeing that code in the model.

tylerecouture (changes requested):

src/library/templates/library/library_quests.html
9921622: you cant use the "with" template tag with `extends`. If `view_type` and `VIEW_TYPES` are empty it causes all tabs to be active since `None == None`.

An alternative approach could be adding `view_type` to library view context

src/quest_manager/views.py
9921622: done

src/library/templates/library/library_quests.html
tylerecouture: include is a completely different construct than extends.  And I'm sure you can use "with" inside an extends, every template in the app extends from base_sidebar.html with extends base.html, which extends from.  Something else is wrong with your code that is causing you problems. For example check out this one (but there are many extends + with in templates):
badges/templates/badges/detail.html

tylerecouture (changes requested):
Concerned about the change to include from extends. Please see my comment.

tylerecouture (approved):

PR #1666, opened 2024-07-26
Feat: maps auto-update when an item within the map is updated; Closes #1660

### LINKED ISSUE #1660: Auto-update maps when an item within the map is updated
Now that we have a method that allows us to easily check what maps a quest or badge is connected to (thanks to #273), We should auto-regenerate maps when items should be added or removed (currently you have to manually regenerate the map everytime you make a change)

1. This should be a siteconfig option that is ON / TRUE by default: `Automatically regenerate maps when contents changed`
2. This should be a celery task to regenerate the connected maps (concurrent so as not to block page loading)
3. There should be a message that "map x, y, and z will be update"
4. Need a smart way to trigger this, don't want to overload the server with regenerations every little quest edit.
    * When prerequisites change, for sure
    * When a quest or badge is marked visible/invisible.  Is there a consistant way this is done?  perhaps we need a "is_visible" method on the IsAPrereq and HasPrereq mixins that can be overridden by each model, then we can just use those methods to check?  But we only want to check if it's CHANGED, right?  Need ideas.

9921622:
### What?
Quests, badges, or ranks that exist within a map will update the map when they are updated or deleted.

### Why?
See #1660

### How?
Added 

+ Mixin that adds a success message when object is updated/deleted
+ new field `map_auto_update ` in SiteConfig
+ created new queryset `CytoScapeQueryset`
+ new method for `CytoScapeQueryset` called `get_maps_as_formatted_string`
+ celery task called `regenerate_map` similar to `regenerate_all_maps` but without notifications 
+ post save/delete signals to `Quest`, `Badge`, `Rank`, and `Prerequisites`

### Testing?

manually tested `get_maps_as_formatted_string`, regeneration when updating `Quest`, `Badge`, `Rank`, and `Prerequisites`

automated tests for
+ `get_maps_as_formatted_string`
+ `regenerate_map` task
+ `Quest` signal
+ `Badge` signal 
+ `Rank` Signal
+ `Prerequisites` signal

siteconfig form
image

map
image
updating badge
image
image
updating quest
image
image
updating prerequistes (using badge as a prereq for quest)
image
image

### Anything Else?
unlike with other signal tests that i made, I cant check if `regenerate_map` made new CytoElements with the `Prerequisites` signal test

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Introduced badge and rank signal handling to automatically regenerate related maps upon creation or deletion of badges and ranks.
  - Added success messages for users when badges and ranks are updated or deleted, informing them about related map changes.
  - Implemented a new task for regenerating Cytoscape maps based on specific IDs.
  - Enhanced `SiteConfig` with a new parameter for managing automatic map updates.
  - Added a new field in the `SiteConfig` model to control automatic map regeneration.
  - Introduced new mixins to improve user feedback during updates and deletions of quests, badges, and ranks.

- **Bug Fixes**
  - Improved test coverage for signal handling and view interactions related to quest, badge, and rank operations.

- **Refactor**
  - Updated view classes to utilize mixins for success message handling, enhancing user feedback.

- **Tests**
  - Added comprehensive test cases for new features and signal handling to validate functionality and ensure robustness.


9921622:
@tylerecouture review

tylerecouture (changes requested):
This looks awesome!  Hugely needed (the fact that maps need to be manually updated is not intuitive..)

Two main things:

1. Put the mixin in the Map (djcytoscape) app once and use it where needed.  They all looked identical to me, or close enough that you could refactor to be identical.
Why are these specific signals hardcoded?  Does HasAPrereq or IsAPrereq not cover the models needed?  Where does this specific list of models come from?  If it's something specific to maps, then that list of apps should be in some constant within the maps app, and the maps app should recieve the signal and handle them all I think?  And since those signals are pretty much model agnostic, one signal function should cover all cases, no need to repeat them for each different model (I think, please correct me if I'm confused =)

src/badges/apps.py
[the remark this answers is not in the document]
9921622: its so the signal in badges is initialized. But since im moving all map related signals to djcytoscape app its now removed

9921622:
1. put signals and view mixin in djycytoscape app
2. since i put the signals in the same file, I used a helper function to reduce boilerplate code

src/badges/views.py
```diff
@@ -47,11 +47,26 @@ def badge_list(request):
     return render(request, "badges/list.html", context)
 
 
+class BadgeUpdateDeleteMixin:
```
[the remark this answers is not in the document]
tylerecouture: @9921622 I agree.

src/badges/views.py
```diff
@@ -57,7 +59,7 @@ class BadgePrereqsUpdate(ObjectPrereqsFormView):
     model = Badge
 
 
-class BadgeDelete(NonPublicOnlyViewMixin, DeleteView):
+class BadgeDelete(NonPublicOnlyViewMixin, UpdateMapMessageMixin, DeleteView):
```
tylerecouture, line 62: This is cool.

src/djcytoscape/signals.py
```diff
@@ -0,0 +1,63 @@
...
+    # get related maps
+    related_map_ids = list(CytoScape.objects.get_related_maps(instance).values_list('id', flat=True))
+    if not related_map_ids:
+        return
+
+    # run task in background
+    regenerate_map.apply_async(args=[related_map_ids], queue='default')
+
+
+@receiver([post_save, post_delete], sender=Badge)
```
tylerecouture, line 31: This is a way better implementation.  I think we can simplify it even more by adding mutiple senders to a single method like this:

```
@receiver([post_save, post_delete], sender=Quest)
@receiver([post_save, post_delete], sender=Badge)
@receiver([post_save, post_delete], sender=Rank)
def regenerate_related_maps(instance):
    ...
```

You can probably leave the Prereq one as is since it's gotta do some extra stuff.

tylerecouture: These can go directly on the helper function, no need for an intermediary.

tylerecouture (changes requested):

src/djcytoscape/signals.py
9921622: done

tylerecouture (approved):

Issue #1670, opened 2024-07-29
Fix built-in "full_clean" errors


tylerecouture:
On a new database with a new deck, populated with "generate content" and a new student user (just create one with all the defaults)
1. run full_clean() management command
2. report any errors here so we can fix the problems in the code base.

9921622:
using this https://github.com/bytedeck/bytedeck/pull/1672 to check.
The only problem `full_clean` shown was that profiles generated by `generate_content` has invalid grad years
![image](https://github.com/user-attachments/assets/20e24ade-50e5-483a-ab4f-7e517c792de3)

tylerecouture:
@9921622 

* Can you change grad year to  blank=True, null=True and add the migration, to solve that one (also note #1273 assigned to you).
* Can you investigate why the CytoElements aren't getting a valid URL?

yujinyuz:
@tylerecouture Might be better to just pass in `blank=True` instead of both.

`blank=True` will allow having empty grad_year but if you update it, but it will prevent saving users if the grad_year is not provided.

`blank=True and null=True` will allow updating a user even when no grad_year is supplied

tylerecouture:
Closing as resolved on `develop`. Both errors originally reported here are gone:

- **grad_year** — the field no longer exists on `Profile`, so it can't raise "This field cannot be blank."
- **CytoElement `href`** — hrefs now carry trailing slashes (e.g. `/quests/665/`, and interlink `/maps/27/665/286/` validates), so the "URL is missing a trailing slash" error no longer occurs.

I also ran the full check this issue describes — `generate_content` a deck, then `full_clean()` **every** object across all local apps (`quest_manager`, `profile_manager`, `announcements`, `comments`, `notifications`, `courses`, `prerequisites`, `badges`, `djcytoscape`, `portfolios`, `utilities`, `siteconfig`, `tags`) — and it reported **zero** validation errors. Nothing left to fix.

- Claude Code

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #1680, opened 2024-08-07
Fix: cytoscape href and submission draft_comment full_clean errors

9921622:
### What?
Fixed the full clean error for CytoElements with href validation error (url is invalid) and QuestSubmission draft_comment validation error (draft_comment cannot be blank)

### Why?
See #1670

### How?

CytoElement
- replaced URLField with Charfield and a custom validator for local urls
- URLField is for full urls not local ones like `/quest/1/`

QuestSubmission
- added `blank=True` for draft_comment. Since `model.full_clean` throws an error when cleaning a None variable that has `null=True` without `blank=True`
 
### Testing?
- added testing for new validator

### Anything Else?
I noticed badges url did not have a trailing slash like most other urls. So i added a trailing slash.

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
	- Updated URL routing to ensure consistent matching with trailing slashes.
	- Introduced enhanced URL handling capabilities for better flexibility in the `href` field.
	- Modified `draft_comment` in the `QuestSubmission` model to allow for flexibility in form submissions.

- **Bug Fixes**
	- Improved error messages for validation issues related to missing trailing slashes in URLs.

- **Tests**
	- Added tests to validate the functionality of the new URL validation logic.


src/badges/urls.py
```diff
@@ -10,7 +10,7 @@
     path('', views.badge_list, name='list'),
     path('list/', views.badge_list, name='badge_list'),
     path('create/', views.badge_create, name='badge_create'),
-    path('<int:badge_id>', views.detail, name='badge_detail'),  # shows assertions of current students only
+    path('<int:badge_id>/', views.detail, name='badge_detail'),  # shows assertions of current students only
```
tylerecouture, line 13: Nice catch, thank you!

src/djcytoscape/models.py
```diff
@@ -15,6 +15,29 @@
 from quest_manager.models import Category
 
 
+def validate_url(url):
```
tylerecouture, line 18: I'm pretty sure we are already doing this elsewhere in the app.  It doesn't make sense to have to maintain two versions of a local url (relative path) validator.  Can you take a look around and see if you can find it? I may be imagining things.

If you do find it, pick the best implementation (or extend one if the needs are not exactly the same)

tylerecouture: Alternately, can we just USE the flatpages validator mentioned in the docstring?

src/djcytoscape/models.py
```diff
@@ -119,7 +142,7 @@ class CytoElement(models.Model):
                              help_text="if present, will be used as node label instead of id")
     min_len = models.IntegerField(default=1,
                                   help_text="number of ranks to keep between the source and target of the edge")
-    href = models.URLField(blank=True, null=True)
+    href = models.CharField(blank=True, null=True, max_length=50, validators=[validate_url])
```
tylerecouture, line 145: maybe "validate_local_url"? for greater clarity? if I see validate url I'm thinking http...etc

tylerecouture (changes requested):

src/djcytoscape/models.py
9921622: image
image

flatpages doesnt have its own validator. Im pretty sure it would take more work to jury-rig flatpage form directly than what currently exist

9921622: found an alternative `URLOrRelativeURLField`

9921622: not entirely sure to keep the validate url tests. `URLOrRelativeURLField` is basically `URLField` but if it starts with a `/` it adds `www.example.com` at the start

tylerecouture (approved):

PR #1688, opened 2024-08-15
Chore: locking django-tenants version to 3.5.0; Closes #1663

9921622:
### What?
Locked django-tenants version and added a comment explaining why

### Why?
See #1663

### How?
Set `django-tenants==3.5.0` and added a comment explaining that our codebase uses python 3.8 and `3.5.0` is the last version that supports 3.8

In our README
image

django-tenants changelog
image

### Testing?
None. `django-tenants 3.5.0` is our current version rn. 

### Review request
@tylerecouture

## Summary by CodeRabbit

- **Chores**
	- Updated the version specification of the `django-tenants` package to a newer range, maintaining compatibility while ensuring stability within the application.

### DIFF
--- a/requirements-production.txt
+++ b/requirements-production.txt
@@ -1,4 +1,4 @@
-django-tenants>=3.4,<3.6
+django-tenants>=3.5,<3.6  # latest version that supports python 3.8
 backports.csv>=1.0.7,<1.1
 beautifulsoup4>=4.12.3,<4.13
 celery>=5.2.2,<5.4


9921622:
@tylerecouture review

tylerecouture (changes requested):
Although there may only be a 3.5.0, better to pin the version to <3.6, so that in the off chance there is a big fix to the 3.5 version (i.e. 3.5.1) we don't miss it.

Because 3.5.0 is the current working version, it should be pinned at >=3.5,<3.6

requirements-production.txt
```diff
@@ -1,4 +1,4 @@
-django-tenants>=3.4,<3.6
+django-tenants>=3.5,<3.6  # latest version that supports python 3.8
```
tylerecouture, line 1: Ok let's be more specific, in a year from now what does thisean?

tylerecouture (approved):

PR #1701, opened 2024-08-21
Feat: update Django to 4.2 LTS; Closes #1691

### LINKED ISSUE #1691: Update Django to version 4.2
Django 3.2 has reached end of life.  We need to update to a supported version.  Then next LTS is 4.2, so let's start with that Upgrade.  5.2 LTS won't be released until mid 2025.

Step 1 Clean up other dependencies:

To make life easier, before upgrading Django (which may create other incompatibilities with our current dependency versions), please simplify our requirements.txt files.  Currently they are a bit of a mess, and include almost all of our sub-dependencies, with inconsistent version pinning.

- [ ] Use `pipdeptree` (run from command line within venv) to view the root dependacies, and their sub dependancies.  Ideally we only have our root dependancies in our requirements files, and rely on those to properly select the version of subdependancies they need.  So remove all subdependancies. 
- [ ] Sometimes, two root dependancies rely on the same sub-dependancy and we have to pin the sub-dependancy due to make sure they are both compatible.  if so, make sure to leave a comment in the file. 
- [ ] All dependancies should by default be pinned like this: >=current.version.bugfix, <next.version (i.e. allow for bugfix updates if the dependancy uses semantic versioning).  Any dependancies that are pinned differently than this should have a comment explaining why.
- [ ] Submit this requirements update (removal of subdependancies) as its own PR before starting on the django version upgrade.

Step 2 Some Docs:
Some notes to review before starting:

- Django supported versions and timeline: https://www.djangoproject.com/download/#supported-versions
- Django Release Process: https://docs.djangoproject.com/en/dev/internals/release-process/

Step 3 Upgrade:
Read the backwards incompatibilities and upgrade guide here (along with anything else that is relevant)
 https://docs.djangoproject.com/en/5.1/releases/4.2/ includes:

9921622:
### What?
Upgraded Django to 4.2 and fixed all the issues that came with it.

### Why?
See #1691 

### How?
Django lists all removed deprecations in their site. The ones listed below are the only ones I found relevant.
https://docs.djangoproject.com/en/dev/internals/deprecation/
```
4.1
	- TransactionTestCase.assertQuerysetEqual() will no longer automatically call repr() on a queryset when compared to string values.
4.0
	- django.utils.encoding.force_text() and smart_text() will be removed.
	- django.utils.translation.ugettext(), ugettext_lazy(), ugettext_noop(), ungettext(), and ungettext_lazy() will be removed.
	- The HttpRequest.is_ajax() method will be removed.	
	- Support for the pre-Django 3.1 django.core.signing.Signer signatures (encoded with the SHA-1 algorithm) will be removed.
	- Support for the pre-Django 3.1 django.core.signing.dumps() signatures (encoded with the SHA-1 algorithm) in django.core.signing.loads() will be removed.		
	- The providing_args argument for django.dispatch.Signal will be removed.
	- django.conf.urls.url() will be removed.
```

How i fixed each issue

> TransactionTestCase.assertQuerysetEqual() will no longer automatically call repr() on a queryset when compared to string 

fixed this by removing instances of using `repr` in tests

> django.utils.encoding.force_text() and smart_text() will be removed.

using `smart_str` as it is the 4.2 equivalent

>  django.utils.translation.ugettext(), ugettext_lazy(), ugettext_noop(), ungettext(), and ungettext_lazy() will be removed.

The only one out of the list above we use is `ugettext_lazy` which I replaced with `gettext_lazy`

> The HttpRequest.is_ajax() method will be removed.	

Replaced any `is_ajax` calls with `xml_http_request_required` method decorator (implemented in #1617)
This method decorator uses the intended way to check for ajax `request.META.get('HTTP_X_REQUESTED_WITH') == 'XMLHttpRequest'`

> Support for the pre-Django 3.1 django.core.signing.Signer signatures (encoded with the SHA-1 algorithm) will be removed.
> Support for the pre-Django 3.1 django.core.signing.dumps() signatures (encoded with the SHA-1 algorithm) in django.core.signing.loads() will be removed.	

I'm not entirely sure if this was the reason why `tenant.tests.test_views.TenantCreateViewTest.test_default`'s email confirmation link test was failing, but the issue seemed related. The `allauth` dependancy uses signer.dumps() + loads() which generate different keys when the timestamps change. (ie. different on monday vs tuesday)

I fixed this by using `freeze_time` to lock the timestamps to make the keys the same when comparing

Looking in Django's source code, it uses HMAC (HMAC uses timestamps according to Google). Additionally, the Django 3.2 release notes mention that "signing.dumps() and loads() have become shortcuts for TimestampSigner.sign_object() and unsign_object()."

image

But this doesn't explain how `tags.tests.test_views.AutoResponseViewTests` doesn't fail even though it uses `signing.dumps`

> The providing_args argument for django.dispatch.Signal will be removed.

Fixed by removing `providing_args` for `notifications.singnals.notify`. I left the providing args as a comment for future reference.

> django.conf.urls.url() will be removed.

`django.conf.urls.url` was replaced with `django.urls.re_path`. Therefore, replaced everything with `url` with `re_path`

---
#### Stuff that has changed but i cant find a source:

you cannot use reverse foreign key relations on none saved objects

This caused `SemesterCreateUpdateFormsetMixin.get_formset_queryset` to crash on `SemesterCreate` because it uses an unsaved instance of `ExcludeDates`

Fixed by checking if `self.object.pk` exist 

`pytz` has been removed in favor for `zoneinfo`. This screwed with some datetime stuff in tests but i fixed it.

`response.wsgi_request._messages` carries over messages after each consecutive post request. I couldn't find an elegant way to clear the messages before each test. So i added a get request which does clear the messages.

### Testing?
Ran all automated tests individually.
Manually tested the site, making sure nothing on the js side caused any issues.

No tests but some tests were modified because of the changed above.

### Anything Else?
I changed the return value of `xml_http_request_required` from a `JsonRepsonse 400` to a `HTTPResponse 403`
Reason was because of https://stackoverflow.com/a/5298527 saying how 403 is preferred over 400

- changed some tests regarding status_codes to accommodate for this change
- also had to add some tests because of `codecov` test
- added some status_code tests for Badge and Rank popups.
- noticed `Ajax_TagChart` doesn't have any tests associated with it
- I noticed i accentally put `django-querysetsequence>=0.16` twice. I removed one of them

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Updated to Django 4.2, potentially enhancing performance and adding new functionalities.
  - Enhanced AJAX views to ensure they are only accessible through AJAX requests.

- **Bug Fixes**
  - Adjusted expected HTTP status codes in multiple tests to reflect correct access control responses.

- **Documentation**
  - Updated comments and documentation to align with Django 4.2 changes.

- **Refactor**
  - Replaced `url()` with `re_path()` for improved URL routing flexibility across several modules.

- **Tests**
  - Added new tests to verify status codes for AJAX endpoints.
  - Improved assertions in tests by comparing object instances directly rather than their string representations. 

- **Chores**
  - Cleaned up unnecessary code and comments within various files for better maintainability.

### DIFF
--- a/requirements-production.txt
+++ b/requirements-production.txt
@@ -1,4 +1,4 @@
-Django==3.2.25
+Django==4.2
 django-allauth>=0.54,<0.55
 django-bootstrap-datepicker-plus>=5.0.5,<6.0
 django-celery-beat>=2.2,<2.7
--- a/src/announcements/urls.py
+++ b/src/announcements/urls.py
@@ -1,20 +1,20 @@
 from announcements import views
 
-from django.conf.urls import url
+from django.urls import re_path
 
 app_name = 'announcements'
 
 urlpatterns = [
-    url(r'^$', views.list, name='list'),  # function based view
-    url(r'^archived/$', views.list, name='archived'),
+    re_path(r'^$', views.list, name='list'),  # function based view
+    re_path(r'^archived/$', views.list, name='archived'),
     # url(r'^$', views.List.as_view(), name='list'),  # CBV
     # url(r'^create/$', views.create, name='create'),
-    url(r'^create/$', views.Create.as_view(), name='create'),  # CBV
-    url(r'^(?P<pk>\d+)/delete/$', views.Delete.as_view(), name='delete'),
-    url(r'^(?P<pk>\d+)/edit/$', views.Update.as_view(), name='update'),
-    url(r'^(?P<ann_id>\d+)/copy/$', views.copy, name='copy'),
-    url(r'^(?P<ann_id>\d+)/publish/$', views.publish, name='publish'),
-    url(r'^(?P<ann_id>\d+)/comment/$', views.comment, name='comment'),
-    url(r'^(?P<ann_id>\d+)/$', views.list, name='list'),
+    re_path(r'^create/$', views.Create.as_view(), name='create'),  # CBV
+    re_path(r'^(?P<pk>\d+)/delete/$', views.Delete.as_view(), name='delete'),
+    re_path(r'^(?P<pk>\d+)/edit/$', views.Update.as_view(), name='update'),
+    re_path(r'^(?P<ann_id>\d+)/copy/$', views.copy, name='copy'),
+    re_path(r'^(?P<ann_id>\d+)/publish/$', views.publish, name='publish'),
+    re_path(r'^(?P<ann_id>\d+)/comment/$', views.comment, name='comment'),
+    re_path(r'^(?P<ann_id>\d+)/$', views.list, name='list'),
 
 ]
--- a/src/badges/forms.py
+++ b/src/badges/forms.py
@@ -64,7 +64,6 @@ class BulkBadgeAssertionForm(forms.Form):
             search_fields=['name__icontains'],
         )
     )
-    # students = forms.ModelMultipleChoiceField(queryset=None)
     students = forms.ModelMultipleChoiceField(
         # TODO just use the user objects here instead of profile
         # Goign back to the user just to sort by profile string is...a hack.  How to do that properly?!
--- a/src/badges/tests/test_models.py
+++ b/src/badges/tests/test_models.py
@@ -260,7 +260,7 @@ class BadgeAssertionTestModel(TenantTestCase):
         values = []
         for _ in range(num):
             badge_assertion = self.badge_assertion_recipe.make()
-            values.append(repr(badge_assertion))
+            values.append(badge_assertion)
 
         qs = badge_assertion.get_duplicate_assertions()
         self.assertQuerysetEqual(list(qs), values, )
--- a/src/badges/tests/test_views.py
+++ b/src/badges/tests/test_views.py
@@ -265,6 +265,9 @@ class BadgeViewTests(ViewTestUtilsMixin, TenantTestCase):
         self.assertEqual(len(messages), 1)
         self.assertTrue(scape.name in str(messages[0]))
 
+        # to clear any messages before next test
+        self.assert200('badges:badge_list')
+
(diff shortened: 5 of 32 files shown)


9921622:
@tylerecouture should be ready for review

9921622:
@tylerecouture 
ran `python -Wa manage.py test > stdout.txt 2> stderr.txt` from this [stack answer](https://stackoverflow.com/a/5289137)

[stderr.txt](https://github.com/user-attachments/files/16734282/stderr.txt)
[stdout.txt](https://github.com/user-attachments/files/16734300/stdout.txt)

I can only see a couple unique deprecation warnings

#### Deprecations in our dependencies
django-tenants: `TenantMiddleware` is deprecated. Its now called `TenantMainMiddleware`
django-taggit: `index_together` is deprecated. Used in `django-taggit`
import export: DeprecationWarning: 'queryset' must be supplied as a named parameter

#### Deprecations in our repo
`assertQuerysetEqual` is renamed to `assertQuerySetEqual`
`Please remove debug_toolbar.panels.logging.LoggingPanel`
`RemovedInDjango50Warning: The USE_L10N setting is deprecated. Starting with Django 5.0`
`DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.10 it will stop working`
`DeprecationWarning: The 'text' argument to find()-type methods is deprecated. Use 'string' instead.`
`PydanticDeprecatedSince20: The `json` method is deprecated; use `model_dump_json` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide...`

Issue #1736, opened 2025-01-29
Error when acessing Library tab for Campaigns


tylerecouture:
When importing a campaign from the Library tab, on the Campaigns page, get this error when trying to access the list of campaigns:

Internal Server Error: /library/campaigns/

NoReverseMatch at /library/campaigns/
Reverse for 'import_category' with arguments '('',)' not found. 1 pattern(s) tried: ['library/import\\-campaign/(?P<campaign_name>[^/]+)/\\Z']

Request Method: GET
Request URL: http://hackerspace.bytedeck.com/library/campaigns/
Django Version: 4.2.16
Python Executable: /usr/local/bin/uwsgi
Python Version: 3.8.5
Python Path: ['/app/src', '.', '', '/usr/local/lib/python38.zip', '/usr/local/lib/python3.8', '/usr/local/lib/python3.8/lib-dynload', '/usr/local/lib/python3.8/site-packages', '/usr/local/lib/python3.8/site-packages/odf', '/usr/local/lib/python3.8/site-packages/odf', '/usr/local/lib/python3.8/site-packages/odf', '/usr/local/lib/python3.8/site-packages/odf', '/usr/local/lib/python3.8/site-packages/odf', '/usr/local/lib/python3.8/site-packages/odf', '/usr/local/lib/python3.8/site-packages/odf']
Server time: Tue, 28 Jan 2025 15:55:10 -0800
Installed Applications:
('grappelli',
 'django_tenants',
 'tenant.apps.TenantConfig',
 'django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.sites',
 'django.contrib.staticfiles',
 'django.contrib.flatpages',
 'allauth',
 'allauth.account',
 'allauth.socialaccount',
 '[allauth.socialaccount.providers.google](http://allauth.socialaccount.providers.google/)',
 'crispy_forms',
 'django_summernote',
 'bytedeck_summernote',
 'captcha',
 'bootstrap_datepicker_plus',
 'embed_video',
 'django_select2',
 'url_or_relative_url_field',
 'import_export',
 'django_celery_beat',
 'colorful',
 'hackerspace_online',
 'storages',
 'taggit',
 'quest_manager',
 'profile_manager',
 'announcements',
 'comments',
 'notifications',
 'courses',
 'prerequisites',
 'badges',
 'djcytoscape',
 'portfolios',
[...]

PR #1747, opened 2025-02-09
Fix error when accessing library campaigns

yujinyuz:
### What?

Fix for #1736 

### Why?

Bug

### How?

Add condition to check if the title is not empty

### Testing?

Manually tested
### Review request
@tylerecouture

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated the campaign list view so that the import option is now only displayed when a campaign has a valid title, improving the interface consistency.


tylerecouture (changes requested):
I don't think this is a good idea.  The title is a required field (not not null and not blank, so somewhere Categories are being allowed into the database improperly.

I suspect this is occurring as part of the "import" code, where a full_clean() isn't being done and bad data is being allowed in.  I can manuall fix the Library deck that is missing data, but the code fix needs to prevent this from happening in the first place.

Does that make sense?


```
class Category(IsAPrereqMixin, models.Model):
    """ Used to group quests into 'Campaigns'
    """
    title = models.CharField(max_length=50, unique=True)
```

This means we probably needs some full_clean() calls before saving as part of the `class QuestResource(resources.ModelResource)` in quest_manager.admin.py.

tylerecouture (commented):
Does a test trying to import (via django-import-export) a Category without a Title fail?

yujinyuz:
@tylerecouture I think I found the culprit. It's this snippet: 

https://github.com/bytedeck/bytedeck/blob/a45a31c6e0839bf03bc2be9f1679b56122a17960/src/quest_manager/admin.py?plain=1#L102-L105

But it seems that we are already handling empty campaigns

https://github.com/yujinyuz/bytedeck/blob/a45a31c6e0839bf03bc2be9f1679b56122a17960/src/quest_manager/admin.py?plain=1#L142-L154

Could you check if the library campaigns listed are correct?

tylerecouture:
@thefoxwithin before you work any more on the Library stuff, can you do a full review of this bugfix?  Apologies the PR descirption is not very detailed.

thefoxwithin:
Can confirm that when using import_ids for campaigns this isn't an issue, specifically with what I found
image
image

Issue #1556, opened 2024-01-12
Add start and end fields to video


fjbagley:
When selecting a video to add, it would be great if there was an optional start and end times. 

This would add the ;start=____ and/or end=_____ code to the src section of the iframe code.

tylerecouture:
Note to develeopers: this is the Summernote plugin we are currently using for video embeds:
https://github.com/DiemenDesign/summernote-video-attributes

fjbagley:
The plugin hasn't been updated in 4 years and doesn't include the start and end attributes. I was able to modify the script to add those fields. This seems to be working now, but I will need to test more.  I have also modified it to enable JavaScript control which will allow for the Video Tracking feature.

tylerecouture:
We have a copy of the plugin in the bytedeck project, so our copy of it can just be modified directly:

src/static/js/summernote-video-attrbutes.js

tylerecouture:
@fjbagley also note, the google embed code only supports start OR end, but not both at the same time =(

If possible to include this in the plugin edits somehow (even if just helptext on the form field that it will only use one if both values are entered), that would prevent confusion.

fjbagley:
It works if you do start=#&end=#. I have it modified to output this. I just
need to test it some more.

tylerecouture:
Oh cool. Maybe it just doesn't show both on the timeline with those white brackets (or maybe that's only in Firefox!)

fjbagley:
Here are the added fields. User can add any combination of the 4 values. 
![Image](https://github.com/user-attachments/assets/9c9f9234-aef7-40d1-92dd-97a66cde53b0)


Output of above. enablejsapi=1 is for video watch tracking coming later.

![Image](https://github.com/user-attachments/assets/83e4a0d0-b519-4f9c-809d-8496057d9677)

 If start>end, then can't click OK.  Error shows on float

![Image](https://github.com/user-attachments/assets/0f3b44f6-f8a3-4aa8-98c8-463a511d922d)


Should I change any of the working or functionality before pull request?

tylerecouture:
This looks great to me!

tylerecouture:
@thefoxwithin this already has a PR at #1763 , I just need you to pull it locally and try to break it (test it) and make sure it works as expected.
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

thefoxwithin:
[...]

PR #1763, opened 2025-03-07
Add Video Start End Attributes (#1556)

fjbagley:

Frontend look with check for URL
image

Check to see that start time is not after end time
image

What is added to code
image

### Review request
@tylerecouture

### DIFF
--- a/src/static/js/summernote-video-attributes.js
+++ b/src/static/js/summernote-video-attributes.js
@@ -6,7 +6,8 @@
   }else{
     factory(window.jQuery);
   }
-}(function($){
+}
+(function($){
   $.extend(true,$.summernote.lang,{
     'en-US':{ /* English */
       videoAttributes:{
@@ -26,6 +27,14 @@
         alignmentOption2:'Right',
         alignmentOption3:'Initial',
         alignmentOption4:'Inherit',
+        start: 'Start Time (Optional)',
+        startHrs: 'Hours',
+        startMin: 'Minutes',
+        startSec: 'Seconds',
+        end: 'End Time (Optional)',
+        endHrs: 'Hours',
+        endMin: 'Minutes',
+        endSec: 'Seconds',
         suggested:'Show Suggested videos when the video finishes',
         controls:'Show player controls',
         autoplay:'Autoplay',
@@ -50,6 +59,8 @@
       var $editable=context.layoutInfo.editable;
       var options=context.options;
       var lang=options.langInfo;
+      var startSeconds = 0;
+      var endSeconds = 0;
 
       context.memo('button.videoAttributes',function(){
         var button=ui.button({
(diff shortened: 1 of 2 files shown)


tylerecouture:
@thefoxwithin can you download this PR and test it for me?  I don't know if you'll have permissions to do this, but please try:
https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally

thefoxwithin:
> > @thefoxwithin can you download this PR and test it for me? I don't know if you'll have permissions to do this, but please try: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/checking-out-pull-requests-locally
> 
> @thefoxwithin did you miss this?

Yes I did, I don't remember looking at this.

thefoxwithin:
While testing I had to create a new deck with the branch checked-out to see the changes.

The biggest error is that on the quest creation form the `Create` button didn't work. I checked-out another branch and it worked, I created another deck to see if it would work on this new one, it didn't.

Since I can't actually create a quest with these attributes applied I can't see what it looks like outside of the quest creation form.

For further testing, I tried updating a quest and that didn't work either. It seems there is an error when trying to post with this change. I've tried further debugging and it doesn't appear that the button does anything, there's no data on my local server log that suggests anything is happening when I click the green `Create` or `Update` buttons.

A minor thing I noticed was that I can make the start and end times longer than the video. This doesn't really break things but it does make the video end immediately. I wonder if there's a way to make it so that at least the start time can't be longer than the video length?

src/static/js/summernote-video-attributes.js
```diff
@@ -72,7 +83,7 @@
           '<div class="form-group">'+
             '<label for="note-video-attributes-href" class="control-label col-xs-3">'+lang.videoAttributes.href+'</label>'+
             '<div class="input-group col-xs-9">'+
-              '<input type="text" id="note-video-attributes-href" class="note-video-attributes-href form-control">'+
+              '<input type="text" id="note-video-attributes-href" class="note-video-attributes-href form-control" required>'+
```
thefoxwithin, line 86: @fjbagley 
<details>
<summary>If trying to create a quest without any videos it gets blocked by this required tag</summary>

image

I had to put the video into every field to create the quest

image


</details>

After altering the code to remove the `required` tag I was able to create a quest without needing a video href included. The start and end attributes still worked without `required`

<details>
<summary>A side effect of removing `required` is that as soon as I interact with start or end time I can hit the ok button.</summary>

### **Before interacting with start or end time:**
image

### **After interacting with just start time:**
image

### **After interacting with just end time:**
image

</details>

Would you like to continue editing this Bagley, or would you like me to pick up where you left off and continue this?

thefoxwithin (changes requested):

src/static/js/summernote-video-attributes.js
fjbagley: Go for it @thefoxwithin . I had put the required tag in to avoid someone missing putting in a url before submitting and creating an empty video container. If you can easily find a way to require the tag then great, but it isn't critical to the functionality.

fjbagley: @thefoxwithin I know you can get the length of the video through the YouTube iFrame API itself, but then you would have to load the iFrame solely to get the video length. It would also take a relatively significant amount of code to make that happen. Better to work on the other problems.

tylerecouture (changes requested):
@thefoxwithin please resolve the conflicts and go for it, per Fergus's notes above.

Issue #1748, opened 2025-02-09
Missing info in Campaign detail view


tylerecouture:
Instead of blank, the defaults should be:

Quests in this campaign: 0
Total XP available: 0

![Image](https://github.com/user-attachments/assets/e0eb031b-88a4-41d7-9d38-57ac19af6d62)

location example: https://hackerspace.bytedeck.com/quests/campaigns/64/

PR #1769, opened 2025-05-13
Added fallback value for number of quests and total xp now default of…

thefoxwithin:
… 0 will be displayed; Closes #1748

Please ensure you are familiar with our Pull Request Description expectations described here: https://www.pullrequest.com/blog/writing-a-great-pull-request-description/
### What?
Added 0 as a default value to the number of quests in the campaign and the total amount of xp within the campaign.

### Why?
To resolve issue #1748

### How?
The firstof command in Django truncates any value registered as None, so it doesn't display 0. However they implemented a way to add a fallback value, by enclosing 0 in quotation like so "0", instead of returning nothing it returns 0.

### Testing?
Since this altered the HTML file directly I used the real time view of my test website to see what changes were made.
Tested src && flake8 with 0 issues.

### Screenshots (if front end is affected)
Screenshot 2025-05-13 124922

### Review request
@tylerecouture

## Summary by CodeRabbit

- **Bug Fixes**
	- Display now shows "0" for quest counts and total XP when no data is available, enhancing clarity.
- **User Interface Updates**
	- Renamed quest count label to "Published quests in this campaign."
	- Removed the "Active" column from the campaigns list for a cleaner view.
	- All notifications in the dropdown now include a remove option for easier management.
- **Documentation**
	- Expanded setup instructions with detailed steps for creating tenants via web form or command line, including environment activation guidance.

PR #1774, opened 2025-05-16
Added fallback value for number of quests and total xp and specified …

thefoxwithin:
…what quests are shown

default for number of quests 0
default for amount of xp 0

### Why?
closes issue #1748
### Testing?
As done in the last pull request, I'm sorry that I made an oopsie and in the last one more files were edited/committed than I thought, won't happen again. 

Screenshot 2025-05-16 114757

### Anything Else?
I'm so sorry, I started messing around with the git graph trying to discover things and merged the old campaign-detail-view with some other changes so when I updated it I had more commits than I wanted/needed. Won't happen again, and I'll make sure my issue solving is for sure separated from now on. Thank you for under standing.

Everything should be resolved now and I made the change from your comment:

"""
While you are in this code, can you verify whether this shows ALL quests in a campaign, or if it only shows published quests? I think it only shows published quests.

If you confirm it is only published quests, can you change the text to "Published quests in this campaign"?

If it includes all of them, leave it as is.
"""
### Review request
@tylerecouture

### DIFF
--- a/src/quest_manager/templates/quest_manager/category_detail_content.html
+++ b/src/quest_manager/templates/quest_manager/category_detail_content.html
@@ -15,8 +15,8 @@
     <div class="row">
       <div class="col-sm-6">
         <ul class="list-unstyled">
-          <li>Quests in this campaign: {% firstof category_quest_count category.quest_count %}</li>
-          <li>Total XP available: {% firstof category_total_xp_available category.xp_sum %}</li>
+          <li>Published quests in this campaign: {% firstof category_quest_count category.quest_count "0" %}</li>
+          <li>Total XP available: {% firstof category_total_xp_available category.xp_sum "0" %}</li>
           <li>Active: {% firstof category_active category.active %}</li>
         </ul>
       </div>


tylerecouture (changes requested):
Please provide a full PR description including whatever info you put in the old PR.

thefoxwithin:
> Please provide a full PR description including whatever info you put in the old PR.

Adding some context here from the previous (now closed) #1769 PR that initially addressed issue #1748.

In the original PR, I made the following changes:
- Added `"0"` as a default fallback value for both the number of quests and the total XP in the campaign.
- This was necessary because Django’s `{% firstof %}` tag skips `None` and empty values, which can result in those fields rendering as blank. By adding `"0"` as a string fallback, it ensures that the UI consistently shows `0` when there’s no data instead of nothing at all.

This new PR:
- Keeps that `"0"` fallback behavior in place for both quest count and XP.
- Updates the label to **"Published quests in this campaign"** based on feedback in the previous PR and confirmation that only published quests are shown.

Also, just to reiterate what I noted in the description — the previous PR had some unrelated changes accidentally bundled in due to local merges. This one is clean and scoped only to the issue.

Thanks for your patience and review!

tylerecouture (approved):

Issue #1773, opened 2025-05-16
Disable delete buttons on Campaigns with quests.


tylerecouture:
When a campaign has available / visible quests, then disable the Delete button, and change the hover text of the delete button to:

"Can't delete a campaign that has published quests; you must unpublish or delete all quests in the campaign first".  

![Image](https://github.com/user-attachments/assets/1ba4d95c-d836-45d2-9a4d-78d04a6d7c4b)

thefoxwithin:
What change do you want to happen to the delete icon visually if any,

Right now I've modified the icon, I played around with modifying the colour as well. Knowing what change before the PR will make it as smooth as possible.

tylerecouture:
@thefoxwithin visually, I want the disabled styling
 * See Bootstrap styling for disabled buttons: https://getbootstrap.com/docs/3.3/css/#buttons-disabled
 * Change the tooltip/hover text
 * Button doesn't do anything

PR #1785, opened 2025-05-22
Disable campaign delete

thefoxwithin:
### What?
Introduced a disabled state for the "Delete Campaign" button when the campaign has associated quests (`quest_count != 0`), and ensured that a tooltip explains why the button is disabled.

### Why?
Previously, the "Delete Campaign" button was always enabled, even when the campaign contained quests. Per [issue #1773](https://github.com/bytedeck/bytedeck/issues/1773), we want to prevent deletion in such cases and clearly communicate the reason via a tooltip.

### How?
- Added a condition to apply the `.disabled` class to the button only when `object.quest_count != 0`.
- Wrapped the button in a `<span>` element, which holds the `title`.
- This allows the tooltip to appear on hover even when the button is disabled.
- Included Bootstrap suggested safety features `tab-index="-1"` and `aria-disabled="true"`

> Note: The `<span>` element is there because the `.disabled` class forces `pointer-events: none;`
> which would cause the title of the button to not display when hovered

### Testing?
- Manually verified that:
  - When `quest_count == 0`, the button is active and links to the delete URL.
  - When `quest_count > 0`, the button appears disabled and the tooltip is visible on hover.

<details>
<summary>Campaign with Quests (Disabled Button with Tooltip)</summary>

Screenshot 2025-05-22 095532
Screenshot 2025-05-22 095803

</details>

<details>
<summary>Campaign without Quests (Active Button)</summary>

Screenshot 2025-05-22 101241
Screenshot 2025-05-22 101254

</details>

### Review request
@tylerecouture

### DIFF
--- a/src/quest_manager/templates/quest_manager/category_detail.html
+++ b/src/quest_manager/templates/quest_manager/category_detail.html
@@ -12,10 +12,24 @@
                 title = "Edit this campaign" >
             <i class="fa fa-edit"></i>
             </a>
-            <a class="btn btn-danger" href="{% url 'quests:category_delete' object.id %}" role="button"
+            {% if object.quest_count != 0 %}
+            <span title="Unpublish or delete all quests before deleting campaign">
+                <a class="btn btn-danger disabled"
+                    href="#"
+                    tabindex="-1"
+                    aria-disabled="true"
+                    role="button"
+                    title = "Unpublish or delete all quests before deleting campaign" >
+                <i class="fa fa-trash-o"></i>
+                </a>
+            </span>
+            {% else %}
+            <a class="btn btn-danger"
+                href="{% url 'quests:category_delete' object.id %}" role="button"
                 title = "Delete this campaign" >
             <i class="fa fa-trash-o"></i>
             </a>
+            {% endif %}
         </div>
     {% endif %}
     {% if object.active or request.user.is_staff %}
--- a/src/quest_manager/templates/quest_manager/tab_campaigns_list.html
+++ b/src/quest_manager/templates/quest_manager/tab_campaigns_list.html
@@ -37,12 +37,25 @@
                title="Edit this campaign">
               <i class="fa fa-edit"></i>
             </a>
-            <a class="btn btn-danger"
-               href="{% url 'quests:category_delete' object.id %}"
-               role="button"
-               title="Delete this campaign">
-              <i class="fa fa-trash-o"></i>
-            </a>
+            {% if object.quest_count != 0 %}
+            <span title="Unpublish or delete all quests before deleting campaign">
+              <a class="btn btn-danger disabled"
+                href="#"
+                tabindex="-1"
+                aria-disabled="true"
+                role="button"
+                title="Unpublish or delete all quests before deleting campaign">
+                <i class="fa fa-trash-o"></i>
+              </a>
+            </span>
+            {% else %}
+              <a class="btn btn-danger"
+                href="{% url 'quests:category_delete' object.id %}"
+                role="button"
+                title="Delete this campaign">
+                <i class="fa fa-trash-o"></i>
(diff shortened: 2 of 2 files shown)


tylerecouture:
> **Actionable comments posted: 0**
> 🧹 Nitpick comments (4)
> 
> > src/quest_manager/templates/quest_manager/category_detail.html (3)> `16-25`: **Ensure tooltip activation and attribute consistency.**
> > > 
> > > * If you rely on Bootstrap’s styled tooltips, add `data-toggle="tooltip"` to the `<span>` (or inner `<a>`) and initialize tooltips in your JS, since native `title` attributes won’t trigger Bootstrap tooltips.
> > > * You can remove the duplicate `title` on the inner `<a>` and keep it only on the `<span>`.
> > > * Clean up spacing around `=` in attributes (e.g., `title="…"` rather than `title = "…"`).
> > > 
> > > Please verify that tooltips are initialized globally or add:
> > > ```
> > > <span data-toggle="tooltip" title="Unpublish or delete all quests before deleting campaign">
> > >   …
> > > </span>
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `26-32`: **Standardize attribute formatting on active delete link.**
> > > The active `<a>` in the `else` branch is correct. For consistency, remove spaces around `=` in your attributes:
> > > ```diff
> > > - href="{% url 'quests:category_delete' object.id %}" role="button"
> > > - title = "Delete this campaign"
> > > + href="{% url 'quests:category_delete' object.id %}" role="button"
> > > + title="Delete this campaign"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > `15-32`: **Extract delete-button markup into a shared include.**
> > > This disabled/active delete-button snippet is duplicated in multiple templates (`category_detail.html` and `tab_campaigns_list.html`). Consider moving it into a partial (e.g., `_delete_campaign_button.html`) and including it with the relevant context (`quest_count` and `object.id`). This will reduce duplication and ease future updates.
> > 
> > 
> > 
> > src/quest_manager/templates/quest_manager/tab_campaigns_list.html (1)> `41-50`: **Tooltip integration and accessibility for disabled state.**
> > > 
> > > * Add `data-toggle="tooltip"` to the `<span>` (or `<a>`) and ensure tooltips are initialized in JS if you want Bootstrap‐style tooltips.
> > > * You can remove the redundant `title` from the inner `<a>` and keep it on the `<span>`.
> > > 
> > > Confirm that tooltips are active and styled properly in the campaigns list.
> 
> 📜 Review details

@thefoxwithin can you take a look at these nitpicks?  The data-tooltip looks like a good one.  I can't remember if we are using the bootstrap styled tooltips or browser defaults.  Please investigate!  I just want them to be consistant.

.github/workflows/codeql-analysis.yml
tylerecouture, line 1: Something in your process is funky for this to keep happening. Maybe you can show me your workflow? Maybe you aren't created your new feature branches off of develop?

tylerecouture (changes requested):

.github/workflows/codeql-analysis.yml
thefoxwithin: Could it be that I started working on this branch before I rebased to develop again? I worked on this branch, checked for upstream then rebased onto my origin develop that was all caught up with the upstream.

thefoxwithin: I think I found the issue, when using github `source control` to push changes if I've rebased to shrink the number of commits, it pulls first then pushes. I'll use the bash from now on to force push changes when that happens so it doesn't pull changes unrelated to the issue.

Issue #1661, opened 2024-07-20
Library MVP doesn'


tylerecouture:

yujinyuz:
Notes:

When importing a quest, consider the following:

* Download the icon from s3 (or from the provided URL)
 * We cannot simply just copy the image URL provided from the file since tenants importing it might not have access to the URL provided.
* When performing the import, upload the downloaded icon from s3

tylerecouture:
@yujinyuz We can just copy the image url.  Currently uploaded files are not tenant specific.  In the future we might change that, but it won't affect past urls, so for now just go with copying urls!

tylerecouture:
Confusing status of icons in Library vs how they appear in the library tab of another deck:

![Image](https://github.com/user-attachments/assets/99698606-391c-4b54-9355-6245245647d5)

thefoxwithin:
This should be closed, from the testing deck.

<img width="1233" height="919" alt="Image" src="https://github.com/user-attachments/assets/8ba2883a-6c4b-4b95-a573-e38f1e4e4615" />

PR #1793, opened 2025-06-07
Fix: Show correct campaign title on library list; Closes #1764

### LINKED ISSUE #1764: Library and Studio Tyee Campaign Name Differences
I have put 3 quests on the library. They are part of the campaign called Abstract - The Art of Design. When I look at the library tab, they show up as Python Interactive II. This is due to Abstract (on Library) having the same ID (54) as Python Interactive II (Studio Tyee). Perhaps a UUID is necessary for campaign?

Image

Image

thefoxwithin:

### What?

Fixes incorrect campaign titles and icons when importing quests from the Library; ensures reliable cross-schema linking using `import_id`.  
Closes [#1764](https://github.com/bytedeck/bytedeck/issues/1764). and potentially [#1661](https://github.com/bytedeck/bytedeck/issues/1661)

### Why?

Previously, imported quests displayed the campaign title and icon from the *local* schema, which could be incorrect or missing. This caused confusion for users importing quests from the Library, as the campaign context was not accurately reflected. Cross-schema linking was also brittle due to reliance on local database IDs.

### How?

- Added a new `import_id` (UUID) field to the `Category` (campaign) model for consistent identification across schemas.
- Updated the quest import logic to match campaigns by `import_id` instead of local DB `id`.
- Adjusted logic to retrieve and display the **correct campaign title and icon** from the **Library schema**, not the local one.
- Modified template logic and ensured tests confirm expected output.

### Testing?

- Manually verified that importing a quest with an existing `import_id` shows the correct Library campaign title and icon.
- Added test assertions to ensure duplicate-import warnings appear when a campaign with the same `import_id` already exists.
- Ran the test suite to confirm no regressions in quest import functionality.

#### **- ✅ Campaign title and icon shown on imported quests list matches Library campaign**

<details>
<summary>Library quests viewed from the library</summary>

I've underlined the things that are inconsistent and will become consistent across tenants with changes applied
Screenshot 2025-06-07 142525

</details>

<details>
<summary>Library quests before changes viewed from different deck</summary>

I've underlined the things that are different from the Library tenant to the local one
Screenshot 2025-06-07 143353

</details>

<details>
<summary>Library quests after changes viewed from different deck</summary>

I've underlined the things that are being showed correctly now after the changes have been added
Screenshot 2025-06-07 142511

</details>

### Anything Else?

- This change lays groundwork for consistent cross-schema linking between other models too (e.g., badges).
- You may need to run a data migration or manually set `import_id`s on existing `Category` rows if cross-schema import is used on older data.

I still want to work on this either in a different issue or in the same one, but this page even after the changes doesn't show the right information.

<details>
<summary>Import Quest Confirmation page</summary>

image
</details>

It's giving the wrong icon and the wrong campaign name. Let me know if you want me to change this page in the same PR or if you want me to make another issue and do it from there.

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Campaigns now have a unique import ID for improved linking and identification across the Library.
  * Import and export of quests and campaigns now use the import ID for more reliable data transfers.

* **Bug Fixes**
  * Import links for campaigns are only displayed when a valid import ID is available.

* **Improvements**
  * Import confirmation pages now display the campaign's Import ID when relevant.
  * URLs and tests have been updated to use import IDs instead of campaign names for better consistency and reliability.

* **Documentation**
  * Enhanced documentation for campaign import/export processes within the admin interface.


tylerecouture (changes requested):
Please resolve conflicts

src/library/tests/test_views.py
```diff
@@ -163,9 +163,13 @@ def test_import_quest_to_current_deck(self):
...
         with library_schema_context():
-            library_quest = baker.make(Quest)
+            campaign = baker.make(Category, import_id=uuid.uuid4())
+            library_quest = baker.make(
+                Quest,
+                visible_to_students=True,
+                campaign=campaign,
+            )
 
-        # Sanity check that the library quest does not exist in the local test schema
```
tylerecouture, line 168: Why removing this helpful comment?

src/library/tests/test_views.py
```diff
@@ -249,14 +253,20 @@ def test_all_library_category_page_status_codes_for_staff(self):
...
-        import_url = reverse('library:import_category', args=(local_campaign.name,))
+            # Create a category in the library tenant
+            library_category = baker.make(Category, title='Existing Campaign')
+
+        # Create a category in the current tenant with the same import_id
+        current_category = baker.make(Category, import_id=library_category.import_id, title=library_category.name)
+
+        import_url = reverse('library:import_category', args=[library_category.import_id])
+
+        import_url = reverse('library:import_category', args=(current_category.import_id,))
```
tylerecouture, line 269: why are we passing a tuple to one of these reverse() methods and a list to the other right above it? just seems weird the inconsistancy

tylerecouture, line 269: Actually, why are we just overwriting the import_url variable?  What's the purpose of the first reverse() call?

src/library/views.py
```diff
@@ -29,7 +29,8 @@ def quests_library_list(request):
     with library_schema_context():
         # Get the active quests and force the query to run while still in the library schema
         # by calling list() on the queryset
-        library_quests = list(Quest.objects.get_active())
+        library_quests = Quest.objects.select_related('campaign').filter(visible_to_students=True).order_by('name')
```
tylerecouture, line 32: Why remove get_active()?  It has additional checks, including visible_to_students (ie..visible()), removes archived quests, removes quests in campaigns that are inactive, etc.  Seems way better.
```
    def get_active(self):
        return self.get_queryset().datetime_available().not_expired().visible().active_or_no_campaign()

```

src/quest_manager/admin.py
```diff
@@ -194,22 +229,58 @@ def generate_simple_prereqs(self, parent_object, data_dict):
...
+            - A new Category Object may be created if no existing match is found.
+        """
+        campaign_title = data_dict.get('campaign_title')
+        campaign_icon = data_dict.get('campaign_icon')
+        campaign_short_description = data_dict.get('campaign_short_description')
+        campaign_import_id = data_dict.get('campaign_import_id')
+
+        campaign = None
+
+        # Try to find import_id
```
tylerecouture, line 258: Can you make it clearer WHERE it's looking for the campaign?  Is this on the local deck or on the library?  I think local because we aren't switching contexts.

src/quest_manager/admin.py
```diff
@@ -194,22 +229,58 @@ def generate_simple_prereqs(self, parent_object, data_dict):
...
+        if campaign_import_id:
+            try:
+                campaign = Category.objects.get(import_id=UUID(campaign_import_id))
+            except (Category.DoesNotExist, ValueError):
```
tylerecouture, line 262: Why Value Error?  DoesNotExist makes sense, but why are we ignoring ValueError here?

src/quest_manager/models.py
```diff
@@ -30,6 +30,13 @@ class Category(IsAPrereqMixin, models.Model):
         help_text="Quests that are a part of an inactive campaign won't appear on quest maps and won't be available to students."
     )
 
+    import_id = models.UUIDField(
+        default=uuid.uuid4,
+        blank=True,
```
tylerecouture, line 35: `blank=False`, this should never be blank, it's a required field.  blank=True while null=False is just asking for problems with our database I think?

src/quest_manager/admin.py
```diff
@@ -218,8 +289,10 @@ def after_import(self, dataset, result, using_transactions, dry_run, **kwargs):
                 parent_quest = Quest.objects.get(import_id=import_id)
 
                 self.generate_simple_prereqs(parent_quest, data_dict)
-
-                self.generate_campaign(parent_quest, data_dict)
+                if (data_dict.get('campaign_title') or data_dict.get('campaign_import_id')):
```
tylerecouture, line 292: can you remove the outer parenthesis? Why they there?  this isn't javascript or C++ =D

src/quest_manager/templates/quest_manager/tab_campaigns_list.html
```diff
@@ -52,9 +52,18 @@
...
-               title="Import this Campaign into your Deck.  The campaign will appear in your Inactive tab."><i class="fa fa-download"></i></a>
+            {% if object.import_id %}
+               <a href="{% url 'library:import_category' object.import_id %}"
+                  class="btn btn-sm btn-default"
+                  title="Import this Campaign into your Deck.  The campaign will appear in your Inactive tab.">
+                 <i class="fa fa-download"></i>
+               </a>
+            {% else %}
+               <button type="button" class="btn btn-sm btn-default disabled"
+                       title="No import ID available — please generate one in Admin">
```
tylerecouture, line 63: In what cases would this occur?

tylerecouture (changes requested):
Some notes:

> "You may need to run a data migration or manually set import_ids on existing Category rows if cross-schema import is used on older data."

AI blather.  Don't we set a default on the model field?  If so, it will auto generate during the migration, correct?

> It's giving the wrong icon and the wrong campaign name. Let me know if you want me to change this page in the same PR or if you want me to make another issue and do it from there.

Yes! Please add a separate issue for this.

Let lots of little comments for your consideration.  Thanks for all the great work on this!!

thefoxwithin:
> AI blather. Don't we set a default on the model field? If so, it will auto generate during the migration, correct?

Good question! While we do set `default=uuid.uuid4` on the model field, that default only applies when a new `Category` instance is created through Django’s ORM. It does **not** automatically backfill existing rows in the database during the migration.

Because of this, we need an explicit data migration that iterates over all existing `Category` records and assigns a unique UUID to the `import_id` field. Without this, existing rows would have `NULL` or duplicate values, causing issues when we add the `unique=True` constraint.

So, the default covers new data going forward, but for older data, we still need to run a backfill migration or manually set those IDs.

I've looked into this further and refined the process based on what I learned. There are **two migrations**:

1. **First migration**  
   Adds the `import_id` field with:
   - `null=True`
   - `blank=True`
   - **no** `default=uuid.uuid4`
   - **no** `unique=True`  
   
   It also includes a **data migration** to backfill existing rows with unique UUIDs.  
   This avoids issues where:
   - all rows might end up with the same UUID (if default is applied at the migration level),
   - or uniqueness violations occur (since all rows start with `None`).

2. **Second migration**  
   Updates the `import_id` field by:
   - removing `null=True`
   - adding `default=uuid.uuid4`
   - adding `unique=True`
   - **removing `blank=True`** to make the field required in forms (not just in the database)

   This ensures:
   - all new categories get a unique UUID automatically,
   - admin and form-based input treat the field as required,
   - and the database enforces uniqueness.

---

This two-step approach prevents migration errors while keeping data integrity intact.  
And there shouldn't be any other errors — all that would need to happen is running:

```bash
python src/manage.py migrate_schemas
```

…and it’ll work as expected. ✅

src/library/tests/test_views.py
thefoxwithin: This was too long ago to remember I'm sorry, I'll add it back in though.

src/library/tests/test_views.py
thefoxwithin: I had both in there because flake8 was mad at me for not using the `current_category` variable. Idk what I was thinking though because the obvious solution is to just remove the variable but leave the `baker.make...` code.

thefoxwithin: The first reverse call is the one that should be used, and the second one was there for flake8 purposes, see the other comment for a bit more details. (It's a very dumb reason)

src/library/views.py
thefoxwithin: I must not have been able to right click it and `go to definition` and so I didn't know what it was doing.

I've seen a lot more of the codebase since then and I believe that this wouldn't have been done with my current knowledge. It's crazy how different navigating the code space feels after almost 2 months.

src/quest_manager/admin.py
thefoxwithin: Yeah it's looking on the local deck. I've adjusted it to be "Try to find import_id on the local deck"

src/quest_manager/admin.py
thefoxwithin: The `ValueError` is caught here because if the `campaign_import_id` isn't a valid UUID string, trying to convert it with `UUID()` will throw that error. So we catch it to avoid crashing in case the data is malformed or unexpected — basically a safety net.  

And we may get malformed `uuid`s accidentally if a superuser decides to edit it based off of this help text
`help_text="Used to link this object across schemas. Don't edit manually unless necessary."`

It's fairly edge case, but it is necessary.

thefoxwithin (commented):
Thanks for your insightful review. Now that I think about it I might've been able to clear up some of this myself before your review. (It was a lot of things that I've since learned to do better)

Cool to see my own growth too. =)

tylerecouture:
> Good question! While we do set `default=uuid.uuid4` on the model field, that default only applies when a new `Category` instance is created through Django’s ORM. It does **not** automatically backfill existing rows in the database during the migration.
> 
> Because of this, we need an explicit data migration that iterates over all existing `Category` records and assigns a unique UUID to the `import_id` field. Without this, existing rows would have `NULL` or duplicate values, causing issues when we add the `unique=True` constraint.

Is this AI generated response?  I don't think this is true.  

Django docs say `db_default` will do it: https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.Field.db_default

> db_default will still be set at the database level and will be used when inserting rows outside of the ORM **or when adding a new field in a migration.**

But I've never used that, because I think db_default just fall back to  `default` if it's not provided for explicitly.

Can you just try it?  I'm pretty sure `default` will be used on existing data during a migration.

tylerecouture:
> > I'm pretty sure `default` will be used on existing data during a migration.

Confirmed:
https://stackoverflow.com/questions/54855018/how-to-set-default-value-for-new-model-field-in-existing-model-with-data

thefoxwithin:
> > > I'm pretty sure `default` will be used on existing data during a migration.
> 
> Confirmed: https://stackoverflow.com/questions/54855018/how-to-set-default-value-for-new-model-field-in-existing-model-with-data

I read the stack overflow and isn't that pretty much what is being done?


From the code in [0040_category_import_id.py](https://github.com/bytedeck/bytedeck/pull/1793/files#diff-f513e0b457c035c1ba192a3b3b6d13ef4397d5518bd23caa9d0af94aa7fa0087)
```
def backfill_category_import_ids(apps, schema_editor):
    Category = apps.get_model('quest_manager', 'Category')
    for category in Category.objects.filter(import_id__isnull=True):
        category.import_id = uuid.uuid4()
        category.save(update_fields=["import_id"])


class Migration(migrations.Migration):

    dependencies = [
        ('quest_manager', '0039_category_short_description'),
    ]

    operations = [
        migrations.AddField(
            model_name='category',
            name='import_id',
            field=models.UUIDField(
                blank=True,
                null=True,
                help_text="Used to link this object across schemas. Don't edit manually unless necessary.",
            ),
        ),
        migrations.RunPython(backfill_category_import_ids, reverse_code=migrations.RunPython.noop),
    ]
```

From the final part of the stack overflow you sent
```
# Generated by Django 2.0.2 on 2019-02-24 19:45

from django.db import migrations, models

def copy_value_to_short_value(apps, schema_editor):
    MetaDataValue = apps.get_model('app', 'metadatavalue')
    db_alias = schema_editor.connection.alias
    from django.db.models import F
    MetaDataValue.objects.using(db_alias).all().update(
        short_value=F('value')
    )


class Migration(migrations.Migration):

    dependencies = [
        ('app', '0001_initial_commit'),
    ]

    operations = [
        migrations.AddField(
            model_name='metadatavalue',
            name='short_value',
            field=models.CharField(default='value', max_length=200, verbose_name='short_value'),
            preserve_default=False,
        ),
        migrations.RunPython(copy_value_to_short_value),
    ]
```

thefoxwithin:
> Can you just try it? I'm pretty sure default will be used on existing data during a migration.

@tylerecouture
I've tried a few different times and a few different ways,

https://github.com/user-attachments/assets/0eff9482-3347-4d61-a80a-63796b924a2e

I've tried it multiple times and the result is the same, the default does give a uuid value to the pre-existing categories, the issue is that they're all the same. Since we want them to be unique, we write a function in the migration to give them all a different uuid. While this isn't the exact same as what was in the stack overflow it is the same concept.

We're both right and I was just being unclear and didn't realize it fully. We need the backfill not because django doesn't backfill, but that it doesn't backfill the way we want. I'm sorry that this distinction ended up causing us both a headache.

Also I can't include `default=uuid.uuid4` in the first migration otherwise the same issue occurs, django backfills the null values with a single default uuid, then the helper function runs and doesn't find any null values to replace, so it basically does nothing. I actually knew this when I made the other comment, it's the reason I made the second migration.

With all of that in mind, I think this is a simple two step-migration process that works and shouldn't require any other steps.
Since nothing about the process has changed, just the understanding has, there's no commit to push.

src/library/views.py
tylerecouture: In case this happens in the future, as a back up you can `Ctrl + Shift + F` and search the entire codebase for the function.  Sometimes VS Code doesn't make the connections.

src/quest_manager/admin.py
tylerecouture: Is it possible to change a uuid to a non uuid value?  Can you try it?  That should not validate nor save if you try to do that?

tylerecouture:
@thefoxwithin Can you take a look at what I did last time I added import_id field to an existing model?  See if you can dig out the commits and migrations (see git blame on the import_id field for quests, for example).  I'm just curious, I don't recall any of these complications, but maybe I just forgot because it's been a while?

thefoxwithin:
> @thefoxwithin Can you take a look at what I did last time I added import_id field to an existing model? See if you can dig out the commits and migrations (see git blame on the import_id field for quests, for example). I'm just curious, I don't recall any of these complications, but maybe I just forgot because it's been a while?

Yeah I found the first migration for quest `import_ids` and which is fairly straightforward
```
# Generated by Django 2.0.13 on 2019-03-12 20:51

from django.db import migrations, models
import uuid


class Migration(migrations.Migration):

    dependencies = [
        ('quest_manager', '0003_auto_20190111_1312'),
    ]

    operations = [
        migrations.AddField(
            model_name='quest',
            name='import_id',
            field=models.UUIDField(default=uuid.uuid4,
                                   help_text='Only edit this if you want to link to a quest in another system so that when importing from that other system, it will update this quest. Otherwise do not edit this or it will break existing links!',
                                   null=True),
        ),
    ]
```
But then the next commit is using a unique generation for the quests `import_id`s so it does seem this is how it was resolved in the past
```
# Generated by Django 2.0.13 on 2019-03-12 21:59
import uuid

from django.db import migrations


def gen_uuid(apps, schema_editor):
    MyModel = apps.get_model('quest_manager', 'Quest')
    for row in MyModel.objects.all():
        row.import_id = uuid.uuid4()
        row.save(update_fields=['import_id'])


class Migration(migrations.Migration):

    dependencies = [
        ('quest_manager', '0004_quest_import_id'),
    ]

    operations = [
        # omit reverse_code=... if you don't want the migration to be reversible.
        migrations.RunPython(gen_uuid, reverse_code=migrations.RunPython.noop),
    ]
```
And there's two more migrations that reference quest `import_id`s so currently the campaigns `import_id`s would have less migrations then the quests.

image

src/quest_manager/admin.py
thefoxwithin: You're right it needs to be a valid uuid for it to be saved, so I'll remove the ValueError as it's a case that should never happen.

image

src/quest_manager/models.py
```diff
@@ -30,6 +30,12 @@ class Category(IsAPrereqMixin, models.Model):
         help_text="Quests that are a part of an inactive campaign won't appear on quest maps and won't be available to students."
     )
 
+    import_id = models.UUIDField(
+        default=uuid.uuid4,
+        unique=True,
+        help_text="Used to link this object across schemas. Don't edit manually unless necessary."
```
tylerecouture, line 36: This help text should be more informative.  Users don't know what `schemas` are.   Let's change to this:

"This value links your campaign to the corresponding campaign within the Library.  Only change this value if you want to disconnect your campaign from the Library."

tylerecouture (changes requested):
Just a help text update!

tylerecouture (approved):

PR #1795, opened 2025-06-14
Library accordion; Closes #1757

### LINKED ISSUE #1757: Library quest list accordion doesn't expand or load.
Image

thefoxwithin:
### What?

This PR adds the front-end and back-end support for the Library tab under Quests. It includes:

- A new `bootstrap-table-accordion.js` and corresponding CSS for the Library section
- AJAX views and URL routes for displaying quest information dynamically
- New HTML templates and tab structure updates (`tab_library_quests.html`)
- A comprehensive test suite covering the AJAX endpoints and rendering logic

### Why?

The existing accordion JS used in the `quest_manager` section was tightly coupled with deck-specific logic and assumptions. We needed a separate version for the Library to:
- Avoid clashing with identifiers or UI behaviors used in user decks
- Introduce library-specific formatting and dynamic info loading via AJAX
- Maintain visual consistency while preserving the separation of concerns between "My Decks" and the shared Library

Addresses issue [#1757](https://github.com/bytedeck/bytedeck/issues/1757) I started fixing this on another issue, and since I'd done the work I thought I'd make this PR.

### How?

- **Frontend**
  - Created a new JS module: `bootstrap-table-accordion.js` scoped for Library use
  - Added Library-specific CSS rules for styling accordion rows in a way that has similarities with local deck quest tabs
  - Updated `tab_library_quests.html` to use this new accordion behavior and hook into the AJAX quest info view

- **Backend**
  - Added `ajax_quest_library_list` view that returns metadata for quests in the Library schema
  - Added `ajax_quest_info_by_import_id` view to return rendered HTML for quest details via POST
  - Registered these views under `library/urls.py` with appropriate namespacing

- **Tests**
  - Added `test_ajax_quest_library_list` to verify JSON response and icon metadata
  - Added `test_ajax_quest_info_returns_html` to validate correct rendering and status
  - Added `test_ajax_quest_info_invalid_import_id_returns_404` to ensure robustness for invalid IDs

### Testing?

Tests added in `library.tests.test_views.QuestLibraryTestsCase`:

- ✅ `test_ajax_quest_library_list`
- ✅ `test_ajax_quest_info_returns_html`
- ✅ `test_ajax_quest_info_invalid_import_id_returns_404`

Manual verification was also performed to confirm:
- Accordion toggle and quest expansion work as expected
- Dynamic quest info loads via AJAX and displays correctly
- Error states are gracefully handled

https://github.com/user-attachments/assets/baa53ae5-6239-460b-87f7-0f8e4289d65c

### Anything Else?

- The new JS/CSS will only be loaded on Library pages
- The accordion behavior intentionally avoids assuming local deck ownership
- Future work will include UI enhancements and more granular filtering on Library quests
### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Quest library now features an interactive, expandable quest list with status indicators for repeatable and blocking quests.
  * Collapsible detail views with loading spinners are available for each quest in the library.
  * The "Library" tab in the quest manager displays a badge showing the total number of library quests.

* **Improvements**
  * Quest list rendering is more dynamic and responsive, with enhanced accessibility and mobile-friendly display.
  * Optimized quest data loading for improved performance.
  * Button actions in the quest manager adapt based on whether you are viewing library quests or your own quests.

### DIFF
--- a/src/library/views.py
+++ b/src/library/views.py
@@ -4,9 +4,6 @@ from django.contrib.auth.decorators import login_required
 from hackerspace_online.decorators import staff_member_required
 from django.core.exceptions import PermissionDenied
 from django.shortcuts import get_object_or_404, redirect, render
-from django.template.loader import render_to_string
-from django.http import JsonResponse
-from django.views.decorators.http import require_POST
 from quest_manager.models import Quest
 from quest_manager.models import Category
 
@@ -29,7 +26,7 @@ def quests_library_list(request):
 
     with library_schema_context():
         # Get the active quests and force the query to run while still in the library schema
-        # by calling list() on the queryset
+        # by calling len() on the queryset
         library_quests = Quest.objects.get_active().select_related('campaign').prefetch_related('tags')
         num_library = len(library_quests)
 
@@ -41,20 +38,6 @@ def quests_library_list(request):
         }
 
     return render(request, 'library/library_quests.html', context)
-
-
-@require_POST
-@login_required
-def ajax_quest_info(request, id):
-    """
-    AJAX POST returning detailed HTML of a quest by id,
-    always checking library schema.
-    """
-    with library_schema_context():
-        quest = get_object_or_404(Quest, pk=id)
-        template = 'quest_manager/preview_content_quests_avail.html'
-        html = render_to_string(template, {'q': quest}, request)
-        return JsonResponse({'quest_info_html': html})
 
 
 @login_required
--- a/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
+++ b/src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
@@ -51,6 +51,7 @@ function loadQuestOrSubmissionContent(id) {
   // Determine the correct AJAX URL based on the current page
   var currentURL = window.location.href;
   var ajax_url;
+  var postData = { csrfmiddlewaretoken: window.contextData.csrfToken };
 
   if (currentURL.includes("/inprogress/")) {
       ajax_url = `${window.contextData.ajax_submission_root}${id}/`;
@@ -61,7 +62,8 @@ function loadQuestOrSubmissionContent(id) {
   } else if (currentURL.includes("/approvals/")) {
       ajax_url = `${window.contextData.ajax_approval_root}${id}/`;
   } else if (currentURL.includes("/library/")) {
-      ajax_url = `${window.contextData.ajax_library_root}${id}/`;
+      ajax_url = `${window.contextData.ajax_quest_root}${id}/`;
+      postData.use_schema = "library";
   } else {
       ajax_url = `${window.contextData.ajax_quest_root}${id}/`; // Default for available quests or drafts
   }
@@ -70,7 +72,7 @@ function loadQuestOrSubmissionContent(id) {
   $.ajax({
       type: "POST",
       url: ajax_url,
-      data: { csrfmiddlewaretoken: window.contextData.csrfToken },
+      data: postData,
       success: function (data) {
           $contentContainer.html(data.quest_info_html).addClass("ajax-content-loaded");
           $('div.pack').pack();
--- a/src/library/templates/library/library_bttns.html
+++ b/src/library/templates/library/library_bttns.html
@@ -1,9 +1,0 @@
-{% if request.user.is_staff %}
-  <div id="preview-btns-{{ q.id }}" class="pull-right" style="margin-bottom: 1em;">
-    <a href="{% url 'library:import_quest' q.import_id %}"
-       class="btn btn-primary"
-       title="Import this Quest into your Deck. The Quest will appear in your Drafts tab.">
-      <i class="fa fa-download"></i> Import Quest
-    </a>
-  </div>
-{% endif %}
--- a/src/library/urls.py
+++ b/src/library/urls.py
@@ -4,10 +4,6 @@ from library import views
 app_name = 'library'
 
 urlpatterns = [
-    # AJAX endpoints
-    path('quests/ajax_quest_info/<int:id>/', views.ajax_quest_info, name='ajax_library_info'),
-    path('quests/ajax_quest_info/', views.ajax_quest_info, name='ajax_library_root'),
-
     # Library pages
     path('quests/', views.quests_library_list, name='quest_list'),
     path('campaigns/', views.campaigns_library_list, name='category_list'),
--- a/src/quest_manager/static/quest_manager/css/bootstrap-table-accordion.css
+++ b/src/quest_manager/static/quest_manager/css/bootstrap-table-accordion.css
@@ -150,27 +150,3 @@ tr.detail-view > td > div > ul > .list-group-item:last-child,
 tr.detail-view > td > div > .list-group-item:last-child  {
   border-bottom-color: var(--accordian-table-border-color);
 }
-
-/* Library-specific: Flex layout for quest header */
-.quest-header {
-  display: flex;
-  align-items: flex-start;
-  justify-content: space-between;
-  flex-wrap: wrap;
-}
-
-/* Button container for preview/detail views */
-.library-preview-btns {
-  margin-left: 1em;
-  margin-bottom: 1em;
-}
-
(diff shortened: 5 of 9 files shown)


thefoxwithin:
This comment thread is relevant to this PR

@tylerecouture asked:
> Is this duplication necessary? Can they not both be handled with the same files by adding some internal logic? This seems like an unecessary duplication.

My response:
>Yeah, I get that—it definitely looks like duplication at first glance. But in this case I think it's worth it, and here's why:

> The Library tab behaves pretty differently from the regular deck views. It pulls quests from a separate schema, loads details via AJAX differently, and doesn’t follow the same visibility or deck-based logic. If I tried to cram all of that into the existing accordion JS, I’d need to add a bunch of conditionals and special cases that would make the original code harder to follow and more fragile.
> 
> The main difference is that the rest of the tabs use integer ids for quests, while the Library tab references quests by their import_id (UUID). That affects how rows are identified, how AJAX URLs are built, and how selectors are matched in the DOM. Trying to support both formats in a single file would’ve introduced a lot of branching logic, so splitting it out made it cleaner and safer.
> 
> Having a separate bootstrap-table-accordion.js just for the Library means:
> 
> I can tweak and test Library-specific behavior without risking regressions in other tabs.
> It keeps the logic cleaner and more maintainable long term.
> And if the Library evolves—like supporting search filters, inline previews, or more metadata—it’ll be easier to build on this without touching unrelated code.
> Same deal with the CSS—bootstrap-table-accordion.css for the Library lets us:
> 
> Isolate any visual differences or fixes without accidentally affecting the main deck views.
> Avoid clashing with styles that assume a certain structure or ID format.
> Keep things modular in case the Library UI keeps diverging from the main one.
> So yeah, it’s technically some duplication now, but it’s more like intentional separation to prevent future headaches. Happy to revisit if things start overlapping more later!

thefoxwithin:
> > The main difference is that the rest of the tabs use integer ids for quests, while the Library tab references quests by their import_id (UUID).
> 
> Why doesn't the library tab use the integer id too? Was that pre-existing? whether it uses the import_id or the id, it will still need to reach into the library schema to get the data. If it used `id` like the rest of the views, then we might not have to unnecessarily duplicate so much code. The differences should be in the view code, in python, as much as possible, so that our front end code can be as simple and reusable as possible. Can you take a think through this and let me know if this seems feasable?

You're absolutely right and I should've tried this first, I don't know why I got so hung up on using import_id, for some reason I thought it needed to use import_id. I've force pushed a branch that doesn't make new bootstrap-table-accordion files. I think I forgot to add tests but that's an easy fix.

thefoxwithin:
I've added the tests and here's an updated video of the accordion, it's got more info because it'll render properly now. In my old method none of that info like "Requires Approval" would render properly.


https://github.com/user-attachments/assets/15acf634-81df-42de-b34f-fe6bbe8be600

It's still a little bit goofy about campaigns, but that should be fixed when the campaigns have import_ids like quests and badges.

tylerecouture (changes requested):
I haven't looked at the code yet, but I noticed in your video (THANK YOU FOR THESE! SUPER HELPFUL!) that the header doesn't go blue when a quest is expanded.  Can you make sure that wasn't broken?

image

thefoxwithin:
I spent a really long time trying to figure out the colouring on the library accordion to figure out it had something to do with how I initialized the table using data-url causing bootstrap to fill in things on it's own making it so that the classes defined in the template wouldn't load.

TL;DR
I spent a long time figuring out all I had to do was remove one line of code haha.

Still working on some minor things that'll keep the table looking correct.

src/library/templates/library/preview_content_lib.html
```diff
@@ -0,0 +1,54 @@
...
+    <div class="col-xs-6">
+      <p>Campaign: {{ quest.campaign|default:"-" }}</p>
+    </div>
+    <div class="col-xs-6">
+      <p>{{ config.custom_name_for_tag }}s: <small>{{ quest.tags.all|join:", "|default:"-" }}</small></p>
+    </div>
+  </div>
+  <hr class="visible-xs-block">
+
+  <div id="preview-btns-{{ quest.id }}" class="pull-right" style="margin-bottom: 1em;">
```
tylerecouture, line 12: This file appears identical to `quest_manager/preview_content_quests_avail.html`  except for the set of buttons here.  Instead of duplicating all the code here, can we instead introduce some logic that just changes the buttons based on the library context, when on the library tab?

The `quest_manager/buttons.html` already has a lot of logic in it to determine which buttons to present, so might make sense to use that.

There's a reason for this, for example you've already changed this so it's different in ways we don't want it to be different from preview_content_quests_avail.html:

` <p>{{ quest.short_description|default:"No description." }}</p`    
vs   
` <p>{{q.short_description|safe}}</p>`

These lines should be identical.

src/library/views.py
```diff
@@ -29,18 +30,33 @@ def quests_library_list(request):
     with library_schema_context():
         # Get the active quests and force the query to run while still in the library schema
         # by calling list() on the queryset
```
tylerecouture, line 32: Now we aren't calling list at all.  Is this code comment accurate or not?  Do we need to force evaluation within the library schema context or not?

src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
tylerecouture, line 1: Nice. awesome that you got it working within the existing accordion script!

tylerecouture (changes requested):
This is looking WAY better!  Thank you!

I have a request to try to remove one more (possibly) unnecessarily duplicate file.  

Please take a look.

src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
thefoxwithin: It took a good amount of learning but that's what I'm here for, at least that's part of what I'm here for. Thank you again for this opportunity again.

src/library/views.py
thefoxwithin: The comment is not accurate anymore — we're not calling `list()`.  
However, we are forcing evaluation with `len()`, so the query still runs inside the `library_schema_context()`, which is correct.  
The logic is fine, but I should have updated the comment. Do you want me to add back the `list()` call just in case?

thefoxwithin (commented):
These PR reviews have been really valuable for me — they help me learn, catch errors earlier, and engage more deeply with the project. It makes me feel like I’m contributing to something meaningful.

That said, I’m open to suggestions on how to make my PRs smoother or more efficient to review. Any guidance would be welcome!

src/quest_manager/static/quest_manager/js/bootstrap-table-accordion.js
```diff
@@ -60,6 +61,9 @@ function loadQuestOrSubmissionContent(id) {
       ajax_url = `${window.contextData.ajax_submission_root}${id}/past/`;
   } else if (currentURL.includes("/approvals/")) {
       ajax_url = `${window.contextData.ajax_approval_root}${id}/`;
+  } else if (currentURL.includes("/library/")) {
+      ajax_url = `${window.contextData.ajax_quest_root}${id}/`;
+      postData.use_schema = "library";
```
thefoxwithin, line 66: Here's how we trigger the `from_library_schema_first()`

tylerecouture (approved):
Awesome!

tylerecouture:
> 
> These PR reviews have been really valuable for me — they help me learn, catch errors earlier, and engage more deeply with the project. It makes me feel like I’m contributing to something meaningful.
> 
> That said, I’m open to suggestions on how to make my PRs smoother or more efficient to review. Any guidance would be welcome!
> 
> 

I'm not sure I have any suggestions? Part of the pain is just you learning the codebase. Another part of the pain is me forgetting how everything works 😂.

Maybe ask more questions about anything that's ambiguous? Though I feel like you're doing that well now.

Always check for how similar things are done elsewhere in the code to improve consistency.  You'll have noticed I'm big on DRY code (don't repeat yourself =)

All in all, PRs are getting better every time and I suspect they'll keep improving till the end!

PR #1802, opened 2025-06-20
Feat: archive tab; Closes #1800

### LINKED ISSUE #1800: Archived quests are only accessible through Django
If a user archives a quest they are not longer able to access this unless they have Superuser status and only through the Django backend.

Maye add a small tab in quests called Archived?

thefoxwithin:
### What?

- Implements a dedicated **"Archived"** tab for quests.
- Modifies the logic for displaying archived quests: staff can now access them in the new "Archived" tab (ensuring they don't appear in any other tabs).
- Archived quests now display **"View Details"** and **"Unarchive"** buttons.
- Ensures only staff can unarchive quests.
- Updates `get_queryset()` to default to `include_archived=False` to prevent unintended inclusion.
- Refines unarchive logic:
  - Unarchived quests are now sent to the **Drafts** tab.
  - If the quest is already unarchived, a message is shown linking to its detail page in the **Drafts** tab.
- Adds tests for unarchive behavior and Drafts fallback.
- Updates and clarifies related comments throughout.

### Why?

For [issue #1800](https://github.com/bytedeck/bytedeck/issues/1800)

- The new "Archived" tab gives staff a proper interface for managing archived quests.
- Previously, archived quests were only accessible via the Django admin, which not all staff can access.
- This ensures archived quests are *only* visible in the Archived tab, and the unarchive behavior is predictable and user-friendly.

### How?

- Added new logic to display archived quests in a dedicated tab for staff users.
- Modified the unarchive view to:
  - Redirect unarchived quests to the **Drafts** tab if appropriate.
  - Show a link to the quest’s detail page.
- Wrote tests in `QuestViewQuickTests` to cover these behaviors.
- Refactored and clarified comments in views and tests.

### Testing?

- Ran full test suite, including new tests for unarchiving and Drafts tab logic.
- Manually verified in the UI:
  - Staff do **not** see archived quests in "Available" or "Drafts".
  - Staff **do** see archived quests in the "Archived" tab.
  - Unarchiving a quest sends it back to the appropriate location (usually Drafts).
  - Attempting to unarchive an already unarchived quest shows an appropriate message with a working link.

No major UI changes, but here is a video showing the new "Archived" tab and the other modifications

https://github.com/user-attachments/assets/ab66d939-753b-4c48-a872-be1fa4635d9f

Here's the updated `help_text` when choosing to archive a quest or not 
image

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Introduced an "Archived Quests" tab and sidebar menu for staff users, allowing easy access to archived quests.
  * Added the ability for staff to unarchive quests directly from the interface.
  * Staff can now view and manage archived quests, including a dedicated tab with quest counts.
  * Enhanced quest copying to include archived quests for staff users.

* **Bug Fixes**
  * Improved handling of archived quests in various views to ensure correct visibility and permissions.

* **Tests**
  * Added tests to verify staff permissions for unarchiving quests and proper access restrictions for non-staff users.

* **Documentation**
  * Updated help text for the archived quest field for clarity.


thefoxwithin:
@coderabbitai review

thefoxwithin:
@coderabbitai review?

src/quest_manager/models.py
```diff
@@ -398,14 +397,32 @@ def get_pk_met_list(self, user):
 
 
 class QuestManager(models.Manager):
-    def get_queryset(self, include_archived=False):
+    def get_queryset(self, include_archived=True):
```
thefoxwithin, line 400: To answer questions regarding the `get_archive()` default of `include_archived=True`

When clicking any button regarding an archived quest it can't find it

<details>
<summary>Different URLs (in order) while the default is False</summary>

### View Details
image

### Edit
image

### Copy
image

### Past Submissions
image

### Summary Data
image

### Delete
image

### Unarchive (NEW)
image

</details>

 But when the default is `True` all of those buttons work. Which means that the views need the default to be `True` otherwise the quests won't appear in the queryset and they won't be found, the 404 error.

src/quest_manager/migrations/0040_alter_quest_archived.py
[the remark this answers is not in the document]
tylerecouture: Wait until the other one is merged, then re-make this migration (you may have to delete the migration file on the branch first and then makemigrations again to regenerate it with the new number)

thefoxwithin:
> @thefoxwithin is this still relevant or has this been superseded by the new layout we chose to go with?

@tylerecouture I'm not quite sure what you mean. If you're referring to the library stuff I thought that the archive tab was separate to the library. If you are referring to the library then all of that shouldn't affect this PR as it has no reference to the library.

src/quest_manager/models.py
tylerecouture: There's something bigger going on that you changed if you broke all the views that worked before with the previous default value!  Figure out why they broke and then we can discuss if the default needs to change, but just change and hope isn't a strategy I want to use =D

src/quest_manager/models.py
```diff
@@ -405,14 +404,32 @@ def get_pk_met_list(self, user):
...
+        Returns:
+            QuestQuerySet: The queryset, filtered according to the include_archived flag.
+        """
         qs = QuestQuerySet(self.model, using=self._db)
         if not include_archived:
             qs = qs.not_archived()
         return qs
 
     def get_active(self):
-        return self.get_queryset().datetime_available().not_expired().visible().active_or_no_campaign()
```
tylerecouture, line 415: Why are we changing this?

src/quest_manager/models.py
```diff
@@ -439,13 +456,40 @@ def get_available_without_course(self, user):
         return qs.not_in_progress_completed_or_cooldown(user)
 
     def all_drafts(self, user):
-        qs = self.get_queryset().filter(visible_to_students=False)
```
tylerecouture, line 442: I don't understand why you changed the default in get_queryset, and then proceeded to pass the opposite parameter in all these methods!

src/quest_manager/models.py
```diff
@@ -439,13 +456,40 @@ def get_available_without_course(self, user):
...
+        Return a queryset of archived quests for staff users only.
+
+        Args:
+            user: The user requesting archived quests
+
+        Returns:
+            QuerySet of archived quests if user is staff, empty list otherwise
+        """
+        # If the default of include_archived=False make sure this includes archived
+        qs = self.get_queryset().filter(archived=True)
```
tylerecouture, line 487: Is this the problem? you need to use `self.get_queryset(include_archived=True)`
Also not need to re-filter, it's already done in get_queryset()

src/quest_manager/models.py
```diff
@@ -439,13 +456,40 @@ def get_available_without_course(self, user):
...
+        if user.is_staff:
+            return qs
```
tylerecouture, line 489: save some cycles and check for staff before generating the queryset

src/quest_manager/views.py
```diff
@@ -248,16 +248,20 @@ class QuestCopy(QuestCreate):
     def get_form_kwargs(self):
         kwargs = super().get_form_kwargs()
 
-        # by default, set the quest this was copied from as the new_quest_prerequisite
-        # If this is changed in the form it will be overwritten in form_valid() from QuestFormViewMixin
-        copied_quest = get_object_or_404(Quest, pk=self.kwargs["quest_id"])
+        # Use a queryset that includes archived quests for staff
+        qs = Quest.objects.all()
```
tylerecouture, line 252: by default this won't include archived quests (because the get_queryset ddefault)  Maybe read up on get_queryset method override that we did , so you don't have to change everything uneccessarily, and can just get the queryset you want exactly.

tylerecouture (changes requested):
Please return the default value for get_queryset (that's intentional).  explicitly include the archived quests in your queryset by passing True to get_queryset, when you need them.  Or create a new get_archived() method on the manager.

thefoxwithin (commented):
I hope these videos and extra explanation give the required context as to why I changed `get_queryset` the way I did.

src/quest_manager/models.py
```diff
@@ -74,7 +74,7 @@ def condition_met_as_prerequisite(self, user, num_required=1):
 
         # get all the active quests in this campaign/category
         # active = not expired, past availability date and time, not archived, visible to students (not draft)
-        quests = self.quest_set.get_active()
+        quests = Quest.objects.get_active().filter(campaign=self)
```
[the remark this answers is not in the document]
thefoxwithin: I can't find the reason I made this change, which means I should've documented it more, or not done it at all. I have a suspicion it was to cover something super specific maybe? But if that's the case I don't know what to replicate to find what this is meant to do. I'll be more diligent in covering why I would do something like this in the future, if not for you then for myself.

Since I can't replicate or remember why, I have reverted it.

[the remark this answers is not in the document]
thefoxwithin: I believe I have found why I changed it, using `self.quest_set.get_active()` when `get_active()` has the argument `include_archived=False` uses a version of get_queryset() that doens't have `include_archived` as an argument, so it should be changed to `Quest.objects.get_active()` the `campaign=self` shouldn't matter though so that can stay gone.

thefoxwithin:
@tylerecouture I did as thorough a search I could, and I believe this covers most things, I haven't gotten to tackling the issues I found in the search yet.

# Archived Quests Log

##  Campaign + Map Visibility

- Made a campaign (**Campaign 26**) with:
  - an active quest (**Quest 33**)
  - an archived quest (**Quest 54**)

- Both quests show up in the campaign when logged in as a staff member, and the campaign shows the correct number of published quests and correct amount of XP.

- Made **Quest 33** appear on the map by making its prerequisite the ByteDeck proficiency badge.

- Set **Quest 33** as the prerequisite for the archived **Quest 54** — it **does not** appear on the map.

- Created an archived quest (**Quest 53**) not in a campaign, gave it the ByteDeck proficiency badge as a prerequisite — **did not** appear on the map after refresh.

- Unarchived **Quest 53** → regenerated the map → **now appears** on the map.

- Archived **Quest 53** again → quest was **immediately removed** from the map without regenerating.

- When editing a quest, I **can select archived quests** as prerequisites. These quests won’t display on the map until unarchived — this appears to be **intended behavior**.

---

##  Library Quest Behavior

- Created a quest on the library (**Quest 35**) that is both **published and archived**.

- Switched to a different deck, reloaded the library tab, searched for the quest name — **not found**, which is the **desired behavior**.

---

##  Approval Tabs Review

- Submitted tab: No archived quests  
- Returned tab: No archived quests  
- Approved tab: No archived quests  
- Flagged tab: No archived quests

(Checked while logged in as a teacher. Before archiving any quests that would go in here)

---

##  Edge Case Testing

### ✅ Archived After Approval

- Created a published quest (**Quest 56**) with prerequisite: **Welcome to ByteDeck!**
- Student submitted **Quest 56**.
- Teacher approved **Quest 56**, then archived it.
- Returned to the approved tab → **archived quest appears**.  
  → This seems **intentional**.

---

### ⚠️ Archived Prerequisite Behavior

- Created a new published quest (**Quest 55**) with **Quest 56** (now archived) as its prerequisite.
- Student could **see Quest 55**, but:
  - Clicking "Start" → **404 error**.
  - Clicking “View Details” showed:

  """
  Quest Preview  
  You do not yet meet the prerequisites to start this quest. This preview allows you to view the content of future quests.
  """

- Even though the student **previously submitted and had Quest 56 approved**, they **could not access Quest 55** because **Quest 56 is archived**.

- Clicking the link to **Quest 56** from the details page of Quest 55 shows the **same misleading message**:

  """
  Quest Preview  
  You do not yet meet the prerequisites to start this quest. This preview allows you to view the content of future quests.
  """

> This error message is **misleading**. It implies the student hasn’t met the prerequisites, but the real issue is that the prerequisite is archived.  

---

### ✅ Fixing the Archived Prerequisite

- Teacher **unarchived Quest 56**.
- Student accessed **Quest 55** again:
  - No error on details page.
  - Was able to **start and submit** the quest.
- Map view shows both **Quest 56 and 55**.

---

### ✅ Submission Visibility When Archiving

- Teacher **did not approve Quest 55**, then archived it.
- Checked the submitted tab — **submission disappeared**.
- Unarchived **Quest 55** — submission **reappeared**.
  → This seems **intended**.

---

### ✅ Returned + Archived Submission

- Created **Quest 57** with prerequisite: **Welcome to ByteDeck!**
- Student submitted the quest.
- Teacher **returned** the submission, then **archived** the quest.
- Returned tab: **no submission visible**.
- Student: Couldn’t find submission except via **notifications**.
  - Clicking the notification → **404 error**.

---

### ✅ Flagged + Archived Submission

- Created **Quest 58** with prerequisite: **Welcome to ByteDeck!**
- Student submitted.
- Teacher **flagged** the submission.
- Flagged tab: Quest shows up.
- Archived the quest — it **no longer appears** in the flagged tab.  
  → This seems **intended**.

---

##  Summary / Follow-ups

- Most functionality behaves as expected.
- However, a few issues are worth flagging:
  - ⚠️ **Misleading error message** when a prerequisite quest is archived.
  - ⚠️ **Returned submission becomes inaccessible** when the quest is archived — even through notification links.
- Let me know if there’s anything else to test — this covers most scenarios I could think of for now.

tylerecouture (changes requested):
> * Both quests show up in the campaign when logged in as a staff member,

This is not good. Archived quests are basically deleted quests and should not be considered as existing anywhere in the entire a app except in the archived tab (aka the trash bin?) which is why the default queryset used archived=False).

> * When editing a quest, I can select archived quests as prerequisites.

This is bad, we do not want archived quests to be part of the prerequisite options.

In fact, we have an issue https://github.com/bytedeck/bytedeck/issues/946 to fix the problem of when you archive a quest that is itself a prerequisite for something else. Archived quests should not be prerequisites to anything at all. Actually we have a PR waiting in the wings that I never reviewed https://github.com/bytedeck/bytedeck/pull/1535

One thing that hasn't been checked, is the new map generation form at /maps/generate/. Archived quests should not appear in the Initial Object widget.

I am leaning back toward NOT changing the default.  For the app's architecture, it makes sense that not including archived is the default, because the app assumes they are gone, deleted, and don't exists.  Because of that, it's the archived tab that should have it's own stand alone logic to pull the archived quests and display them.

tylerecouture:
@thefoxwithin Thank you for the thorough review and checking everything so diligently, and I apologies that this is a frustrating issue/task!  We'll figure it out.

thefoxwithin:
> > @thefoxwithin Thank you for the thorough review and checking everything so diligently, and I apologize that this is a frustrating issue/task! We'll figure it out.
> 
> Having a closer look at the problems you outline, I don't think we need any of those buttons on the archived tab. The only thing they need is to be able to unarchive the quest --> Make sure it's a draft, make the message say it has been moved to the drafts folder. Then they can do all the stuff. Archived is basically deleted, and just like the trash bin, you can't do anything to files in the trash bin except un trash them =)

Okay, that makes a lot more sense, I don't believe I had the context that is was going to be the trash bin when I first made the changes. But that makes a lot more sense for how the quests have been handled, and yes this will be a much easier implementation than I thought. Your original comment about it being easy makes a lot more sense too.

src/quest_manager/views.py
```diff
@@ -1086,6 +1100,41 @@ def approvals(request, quest_id=None, template="quest_manager/quest_approval.htm
...
+        request: HTTP request object
+        quest_id: ID of the quest to unarchive
+
+    Returns:
+        Redirect to quests list with success message
+    """
+    quest = Quest.objects.get_queryset(include_archived=True).get(id=quest_id)
+
+    # Make the link that leads to the quests detail page to include in the message
+    url = reverse('quests:quest_detail', args=[quest.pk])
```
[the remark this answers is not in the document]
thefoxwithin: Thanks this worked as a good practice run for updating the urls of the imports on a different issue.

src/quest_manager/templates/quest_manager/buttons.html
```diff
@@ -8,7 +8,15 @@
       </a>
     {% endif %}
 
-    {% if  request.user.is_staff or request.user == q.editor %}
+    {% if q.archived and request.user.is_staff %}
+      <form action="{% url 'quests:unarchive' q.id %}" method="post" style="display:inline">
```
tylerecouture, line 12: We shouldn't be using inline styles, just use the appropriate bootstrap styling or the style classes already used elsewhere in the project for equivalent stuff.

If it's because the form is causing the buttons to look off, look at the other places in the project where we mix POST form buttons and GET (url link) buttons.

src/quest_manager/templates/quest_manager/buttons.html
```diff
@@ -8,7 +8,15 @@
...
+        {% csrf_token %}
+        <button type="submit" class="btn btn-default" title="Unarchive Quest - send it to the drafts tab">
```
tylerecouture, line 14: For consistency with all the other button title/help text which use colons instead of dashes:
"Unarchive this quest: move it into your drafts tab so you can edit it"

src/quest_manager/models.py
```diff
@@ -127,8 +127,7 @@ class XPItem(models.Model):
     )
     archived = models.BooleanField(
         default=False,
-        help_text='Setting this will prevent it from appearing in admin quest lists.  '
-        'To un-archive a quest, you will need to access it through Site Administration.'
+        help_text="Move this quest into the Archived tab."
```
tylerecouture, line 130: We'll need to update this help_text with more info once we decide what else happens when you archive a quest (for example removing it as a prerequisite, deleting any submissions, etc)

tylerecouture (changes requested):
Down to the final boss now...lol.

tylerecouture:
https://github.com/bytedeck/bytedeck/pull/1802#discussion_r2194000447  Not resolved or is it?

PR #1814, opened 2025-07-02
Campaign import filter; Closes #1812

### LINKED ISSUE #1812: Importing an active campaign with only drafts doesn't work
When importing a campaign that only has quests in the drafts panel of the library the campaign doesn't show up in the inactive or active tabs. (similarly to an empty campaign)

<details>

Image

Image

Image

</details>

What do we want to do with active campaigns that only have draft quests in them? Currently draft quests are being treated as invisible/non-existent.

Do we want these campaigns to show up on the import page? If it's an active campaign with inactive quests do we still want to be able to import the quests?

thefoxwithin:
### What?

Refactor the campaign library listing and import confirmation views to filter campaigns and quests based on their current status. Specifically:

- Use a subquery with `Exists` to efficiently filter campaigns to only those with at least one active (visible and not archived) quest.
- Update the import confirmation view to display only current quests (`visible_to_students=True` and `archived=False`) for clarity during import.
- Add tests to verify that:
  - The campaign library list only shows campaigns with current quests.
  - The import confirmation page only shows current quests for the selected campaign.

### Why?

Previously, the campaign list view simply fetched all active campaigns without filtering out those without any visible or unarchived quests. This could result in campaigns appearing in the list that don’t have any current quests available, potentially confusing users.

Additionally, the import confirmation view displayed all quests in a campaign regardless of their visibility or archival status, which could cause confusion about what would actually be imported.

Filtering campaigns and quests at the database/query level improves performance, correctness, and user experience by:

- Showing only campaigns with at least one active (visible and not archived) quest.
- Avoiding confusion by displaying only quests that are currently importable.
- Ensuring data consistency between what users see and what they can import.

This change resolves issue [#1812](https://github.com/bytedeck/bytedeck/issues/1812).

### How?

- Used Django ORM's `Exists` and `OuterRef` expressions to filter campaigns with active quests directly in the query.
- Replaced usage of `.quest_set.all()` with `.current_quests()` in the import confirmation view to only show active quests.
- Added two tests (`test_campaigns_library_list__filters_by_current_quests` and `test_import_campaign_view__shows_only_current_quests`) following repository naming conventions to cover these cases.
- Ensured tests run within the appropriate library schema context.

References:
- [Django's Exists](https://docs.djangoproject.com/en/5.2/ref/models/expressions/#exists-subqueries)
- [Django's OuterRef](https://docs.djangoproject.com/en/5.2/ref/models/expressions/#django.db.models.OuterRef)

### Testing?

- Confirmed all existing and new tests pass.
- Verified the campaign list page excludes campaigns without current quests.
- Verified the import confirmation page only lists current quests.
- Manual UI spot checks to confirm no inactive or archived quests appear on import confirmation.

<details>
<summary>Before the changes from this branch</summary>

### From the local Deck
image
image

### From the Library Deck
It's all the same so showing these won't add context.

</details>

<details>
<summary>After the changes from this branch</summary>

### From the Library Deck

</details>

### Review request
@tylerecouture

## Summary by CodeRabbit

* **Bug Fixes**
  * Campaign listings now only display campaigns with at least one quest that is visible and not archived.
  * When importing a campaign, only quests that are visible and not archived are shown in the confirmation view.
  * Inactive campaigns are excluded from campaign listings even if they contain visible, unarchived quests.

* **Tests**
  * Added tests to ensure campaign and quest filtering behaves correctly in listings and import confirmation views.


src/library/views.py
```diff
@@ -54,9 +55,22 @@ def campaigns_library_list(request):
     """
 
     with library_schema_context():
-        # Get the active quests and force the query to run while still in the library schema
-        # by calling list() on the queryset
-        library_categories = list(Category.objects.filter(active=True))
+        # Filter quests that are visible to students and not archived,
+        # linked to the campaign (category) via OuterRef for subquery use.
+        filtered_quests = Quest.objects.filter(
```
tylerecouture, line 60: Let's put this logic in a method in the Category's Manager class (which it doesn't have at the moment, but you can see almost every other model does have one).

Then we can just do `library_categories = Category.objects.all_active_with_available_quests()` in the view.

And that also allows us to properly unit test the `all_active_with_available_quests()` separately from the view method.

I'm not sure this is the cleanest way to do the query, but lets refactor into a Category manager method first, with tests, then we can tinker with the insides of the method if needed.

tylerecouture: Some reading for you:  https://docs.djangoproject.com/en/4.2/topics/db/managers/

tylerecouture (changes requested):

src/library/views.py
tylerecouture: Can you explain that more?  When I first made most of the models I went overkill with creating model QuerySet classes.  I think because this is the only one we have, it's fine to access via the a manager...unless you have a good reason to add both Manager and QuerySet for this single method?

src/quest_manager/models.py
```diff
@@ -19,12 +19,52 @@
...
+        return self.filter(active=True).annotate(
+            has_available_quests=Exists(filtered_quests)
+        ).filter(has_available_quests=True)
+
+
+class CategoryManager(models.Manager):
+    """
+    Custom model manager for Category (campaign) objects.
+    """
+    def get_queryset(self):
```
tylerecouture, line 47: No need to override this if we aren't changing anything

src/quest_manager/models.py
```diff
@@ -19,12 +19,52 @@
...
+        """
+        Returns the base QuerySet for Category objects,
+        using the custom CategoryQuerySet with additional filtering utilities.
+        """
+        return CategoryQuerySet(self.model, using=self._db)
+
+    def all_active_with_available_quests(self):
```
tylerecouture, line 54: Just put all the code here. I don't think we need  CategoryQuerySet at this point, so simplify the code and just put it all in a single Manager method.

src/quest_manager/models.py
```diff
@@ -19,12 +19,52 @@
...
+        quest that is visible to students and not archived.
+        Uses a correlated subquery with EXISTS for performance.
+        """
+        filtered_quests = Quest.objects.filter(
+            campaign=OuterRef('pk'),
+            visible_to_students=True,
+            archived=False
+        )
+
+        return self.filter(active=True).annotate(
```
tylerecouture, line 38: lots of steps here on a single line, seperate them out into their own lines of code, and then add a code comment for each explaining WHY you are doing what you're doing.

tylerecouture: in the end, just return the queryset without any additional manipulation on the return line.

src/quest_manager/models.py
```diff
@@ -19,12 +19,52 @@
...
+    """
+    Custom queryset for Category model (campaigns).
+    """
+    def with_available_quests(self):
+        """
+        Annotates and filters categories that are active and have at least one
+        quest that is visible to students and not archived.
+        Uses a correlated subquery with EXISTS for performance.
+        """
+        filtered_quests = Quest.objects.filter(
```
tylerecouture, line 32: There should be an existing Quest manager method for this that is already tested, if not, create it there and call it. That  helps us keep Quest logic in the Quest model and can reuse it later, and also properly unit test it.

Also, "filtered_quests" is a terrible name because it tells you nothing about what the queryset is for or contains!

tylerecouture: Also, it's not clear what you mean by "available" since that's not a field.  There are a thousand ways quests could be available or unavailable, so be explicit here.  By default, when I read "available" it means available to a student, which doesn't make sense here.

What's this method for?  Maybe it needs the context added to the docstring.

tylerecouture (changes requested):
Apologies, I think this just needs better code comments and documentation about what's going on and more importantly, why.

src/quest_manager/models.py
```diff
@@ -19,12 +19,47 @@
 from tags.models import TagsModelMixin
 
 
+class CategoryManager(models.Manager):
+    """
+    Custom manager for Category model that provides
+    queryset methods for active campaigns with available quests.
```
tylerecouture, line 25: This is a really specific class docstring that will be irrelevant the second we add a new method to the class.  I don't think a docstring is required here at all because these are standard django classes.  Or, just a custom manager with a link to custom manager documentation is all we need.

src/quest_manager/models.py
```diff
@@ -19,12 +19,47 @@
...
+        return qs
+
+
 class Category(IsAPrereqMixin, models.Model):
     """ Used to group quests into 'Campaigns'
     """
     title = models.CharField(max_length=50, unique=True)
     icon = models.ImageField(upload_to='icons/', null=True, blank=True)
     short_description = models.CharField(max_length=500, blank=True, null=True)
+    objects = CategoryManager()
```
tylerecouture, line 62: `objects`  should go after all the fields, with a line break. Check out the other models for consistancy

src/quest_manager/models.py
```diff
@@ -19,12 +19,47 @@
...
+class CategoryManager(models.Manager):
+    """
+    Custom manager for Category model that provides
+    queryset methods for active campaigns with available quests.
+    """
+
+    def all_active_with_importable_quests(self):
+        """
+        Returns a queryset of active campaigns (categories) that have
+        at least one quest visible to students and not archived.
```
tylerecouture, line 31: Maybe add some additional context here. Because the Library isn't accessible to students `visible_to_students` is actually a deceptive name for the field.  Also, it's confusing (the name of that field, because quests where `visible_to_students` is True, might not literally be visible to students!  Can you create a **new issue** to change `Quest.visible_to_students` to  `Quest.published`?  That's actually how it appears in the front end because verbose name is "published".

For this docstring, and anywhere else you see it from now on, please refer to it as "visible_to_students (i.e. published)" similar to how we refer to "Category (i.e. Campaign)"

This will prevent confusion with a quest's ACTUAL visibility to students, which is a combo of archived, expiry date, prerequisites being met, archived, etc, etc.

src/quest_manager/models.py
```diff
@@ -19,12 +19,47 @@
...
+
+        This helps efficiently filter campaigns relevant for display
+        or import in the library context.
+        """
+
+        # Start with all categories filtered to only those marked active
+        qs = self.get_queryset()
```
tylerecouture, line 38: the code comment above this line?  that's not what this is doing.  this is not filtered at all, so what does that code comment mean?

src/quest_manager/models.py
```diff
@@ -19,12 +19,47 @@
...
+
+        # Annotate each category with a count of related quests that:
+        # - are visible to students (not drafts)
```
tylerecouture, line 41: boom, not drafts, exactly!  I am realizing how inconstant we are throughout this project.  Campaign.active, Quest.visible_to_students , Badge.active, Announcement.draft

These should all be the same field name  * CRY *

src/library/tests/test_views.py
```diff
@@ -293,3 +293,59 @@ def test_import_campaign_success(self):
 
         # all imported quests should be inactive for this campaign
         self.assertEqual(imported_library_campaign.quest_set.filter(visible_to_students=False).count(), 3)
+
+    def test_campaigns_library_list__filters_by_current_quests(self):
```
tylerecouture, line 297: We need a `test_campaigns_library_list__excludes inactive_campaigns()` because I believe that would have failed...

tylerecouture (changes requested):
This is great, so much better and easier to read.  I think missing a critical filter (and test), so please confirm.

src/quest_manager/models.py
thefoxwithin: Fair enough, I was just trying to do my best to keep everything documented like the functions.

src/quest_manager/models.py
```diff
@@ -19,6 +19,35 @@
 from tags.models import TagsModelMixin
 
 
+class CategoryManager(models.Manager):
+
+    def all_active_with_importable_quests(self):
```
tylerecouture, line 24: Nicely done. I like this method.

tylerecouture (approved):

PR #1835, opened 2025-07-15
importing now searches all local quests including archived; Closes #1834

### LINKED ISSUE #1834: Importing and Archiving interaction
If I import a quest then archive it I'm allowed to import the quest again, clicking the button doesn't appear to do anything, but it does give me an error. I shouldn't be allowed to try to import the quest again, so we'll have to include archived quests in the search on the local deck.

https://github.com/user-attachments/assets/285da845-0362-4c47-96fd-b052fe7c2cdc

thefoxwithin:
added tests for finding a local quest in various states

### What?

This PR fixes a bug where archived quests on the local deck were not being detected during library import, resulting in a confusing UX where clicking “Import” appeared to do nothing.  
The fix ensures archived quests are included in the lookup when checking if a quest has already been imported.

### Why?

Previously, when a user archived an imported quest and tried to re-import the same quest from the library, the system failed to detect the existing (archived) quest.  
This allowed the "Import" button to appear active but did nothing on click, leading to an error.  
We now include archived quests in this lookup so users aren't allowed to import duplicates — even if archived.

https://github.com/bytedeck/bytedeck/issues/1834

### How?

- Updated `Quest.objects.filter(...)` to `Quest.objects.all_including_archived().filter(...)` in the relevant library import views.
- Ensures the check for existing quests includes archived ones.

### Testing?

Added test cases that:
- Create local quests in all three states: visible, invisible (draft), and archived.
- Create corresponding library quests with matching `import_id`s.
- Confirm that attempting to import any of the matching library quests (even if the local one is archived) results in a message:  
  `'Your deck already contains a quest with a matching Import ID'`.

N/A

### Review request
@tylerecouture

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved import checks to prevent importing quests with matching IDs, including archived and unpublished quests.

* **Tests**
  * Expanded test coverage to verify import restrictions and user notifications for quests in various states.

### DIFF
--- a/src/library/tests/test_views.py
+++ b/src/library/tests/test_views.py
@@ -144,14 +144,34 @@ class QuestLibraryTestsCase(LibraryTenantTestCaseMixin):
         # import_id already on the current deck
         # TODO: When we add an overwrite feature, this quest will need to be modified to test that feature
 
-        # Create a quest in the local test schema
+        # Create quests in the local test schema with varying states
         quest = baker.make(Quest)
+        quest_2 = baker.make(Quest, visible_to_students=False)
+        quest_3 = baker.make(Quest, archived=True)
 
-        # create a quest in the library schema with same import_id
+        # create quests in the library schema with same import_ids
         with library_schema_context():
             library_quest = baker.make(Quest, import_id=quest.import_id)
+            library_quest_2 = baker.make(Quest, import_id=quest_2.import_id)
+            library_quest_3 = baker.make(Quest, import_id=quest_3.import_id)
 
         url = reverse('library:import_quest', args=[library_quest.import_id])
+
+        # Check that it sends you to the confrimation page with
+        self.assertContains(
+            self.client.get(url),
+            'Your deck already contains a quest with a matching Import ID'
+        )
+
+        url = reverse('library:import_quest', args=[library_quest_2.import_id])
+
+        # Check that it sends you to the confrimation page with
+        self.assertContains(
+            self.client.get(url),
+            'Your deck already contains a quest with a matching Import ID'
+        )
+
+        url = reverse('library:import_quest', args=[library_quest_3.import_id])
 
         # Check that it sends you to the confrimation page with
         self.assertContains(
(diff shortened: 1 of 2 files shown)


src/library/views.py
```diff
@@ -134,7 +134,7 @@ def get(self, request, quest_import_id):
                 - local_quest: The local quest with a matching import_id, if one exists.
         """
         # Check for local existence to warn the user before importing
```
tylerecouture, line 136: A confusing code comment.  How does this "check for local existence"? It just gets the quest (or None) and sends to the Template.  More context needed here.

tylerecouture (changes requested):
Simple enough, just one tweak!

src/library/tests/test_views.py
```diff
@@ -144,12 +144,16 @@ def test_import_quest_to_current_deck(self):
         # import_id already on the current deck
         # TODO: When we add an overwrite feature, this quest will need to be modified to test that feature
 
-        # Create a quest in the local test schema
+        # Create quests in the local test schema with varying states
```
tylerecouture, line 147: While we're in here, can you improve the docstring on the test method?  Also, can you be more explicit in this code comment "varying states" instead list the specific states being tested.

tylerecouture (changes requested):

src/library/tests/test_views.py
thefoxwithin: I replaced that comment with this.
```
        # Create quests in the local test schema
        # First is a published quest that's not archived
        # Second is a quest that's not published also not archived
        # Third is a quest that's archived
```

There's also labels over the actual testing of the quests. I did think of a better way to resolve this now though, what if I change the names of the variables to be `published_quest`, `unpublished_quest` and `archived_quest`? Would that suffice, since it should be able to be seen from the code that the quest has `published=False` or `archived=True`.

I didn't think of this before since we were talking about the code comment and not the variable names. Also "varying states" is vague so looking at that comment it might've been hard to tell what I was talking about.

src/library/views.py
thefoxwithin: replaced with
```
# Look for a matching local quest (even if archived) to show a warning if it already exists
```

tylerecouture (approved):

PR #1837, opened 2025-07-18
Disable category delete; Closes #1824

### LINKED ISSUE #1824: Disable campaign deletion
The current prevention for deleting a campaign that contains quests works only by disabling the delete button in the UI. However, manual URL manipulation allows access to the delete form and the campaign can still be deleted.

How it was tested:
- Tried deleting a campaign via the button: correctly prevented.
 
Image

- Manually navigated to the deletion URL: deletion form accessible and campaign deletable.

Image

Expected:
- Server-side validation blocking deletion regardless of access method.
- Complete protection from deleting campaigns with quests.

Actual:
- Only client-side or UI-level protection in place.
- Manual URL access circumvents the safeguard.

Recommendation:  
Add backend checks in the deletion view to enforce this constraint, ensuring campaigns with quests cannot be deleted under any circumstance.

Since this involves url manipulation and knowing the id before hand this is low priority however it was an oversight in a previous implementation. Sorry.

thefoxwithin:
### What?

This PR enhances the existing `CategoryDelete` class-based view by adding safeguards to prevent deletion of campaigns (categories) that contain published quests.  
It also updates the delete confirmation template to disable the delete button with a tooltip when published quests exist.  
Tests were modified to verify that these deletion restrictions are properly enforced.

### Why?

Previously, the `CategoryDelete` view allowed deletion of campaigns even if they contained published quests, which could cause data integrity issues.  
Disabling the delete button on the `category_confirm_delete.html` and adding a tooltip informs users upfront why deletion is restricted.

https://github.com/bytedeck/bytedeck/issues/1824

### How?

- Added `post()` and overridden `delete()` methods in the existing `CategoryDelete` view to check for published quests before allowing deletion.  
- If published quests exist, the delete is blocked, an error message is shown, and the user is redirected.  
- The delete confirmation template disables the delete button and adds a tooltip if the campaign has published quests.  
- Updated tests to confirm that campaigns with published quests cannot be deleted while empty campaigns can be.

### Testing?

- Automated tests verify deletion is blocked for campaigns with published quests.  
- Confirmed deletion succeeds for empty campaigns.  
- Manual testing validated the disabled button and tooltip display in the UI.

image

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Deletion of campaigns is now prevented if the campaign contains any published quests. Users will see a disabled delete button with an explanatory tooltip in such cases.

* **Bug Fixes**
  * Improved user feedback and accessibility when attempting to delete campaigns with published quests.
  * Added server-side validation to block deletion if published quests exist, handling race conditions gracefully.

* **Tests**
  * Expanded test coverage to verify that campaigns with published quests cannot be deleted, while empty campaigns can be deleted successfully.
  * Added tests for race condition scenarios to ensure deletion is correctly blocked when quests become published during the process.

### DIFF
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -165,18 +165,23 @@ class CategoryDelete(NonPublicOnlyViewMixin, DeleteView):
 
     def delete(self, request, *args, **kwargs):
         """
-        Attempt to delete the Category object only if it has no quests
-        published. If published quests exist, prevent deletion
-        and display an error message.
+        Attempt to delete the Category object only if it has no published quests.
+        If published quests exist, prevent deletion and display an error message.
+
+        This logic protects against race conditions where the user loads the delete
+        confirmation page while the campaign appears deletable (e.g., only has
+        unpublished quests), but by the time they submit the form, one or more
+        quests may have been published.
+        https://github.com/bytedeck/bytedeck/pull/1837#discussion_r2232233693
 
         Returns:
             HttpResponseRedirect on successful deletion or redirect back with error.
         """
         self.object = self.get_object()
-        self.object.refresh_from_db()
 
-        # Check for any published quests in this category
-        if self.object.quest_set.filter(published=True).exists():
+        # Use a fresh queryset directly from Quest.objects instead of self.object.quest_set
+        # to avoid stale cached related objects and ensure we check the latest DB state.
+        if Quest.objects.filter(campaign=self.object, published=True).exists():
             messages.error(
                 request,
                 "You can't delete this campaign because it contains published quests"
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -156,6 +156,35 @@ class CategoryDelete(NonPublicOnlyViewMixin, DeleteView):
     model = Category
     success_url = reverse_lazy("quests:categories")
 
+    def post(self, request, *args, **kwargs):
+        """
+        Handle POST requests by delegating to the delete method to
+        enforce custom deletion logic and validation.
+        """
+        return self.delete(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs):
+        """
+        Attempt to delete the Category object only if it has no quests
+        published. If published quests exist, prevent deletion
+        and display an error message.
+
+        Returns:
+            HttpResponseRedirect on successful deletion or redirect back with error.
+        """
+        self.object = self.get_object()
+        self.object.refresh_from_db()
+
+        # Check for any published quests in this category
+        if self.object.quest_set.filter(published=True).exists():
+            messages.error(
+                request,
+                "You can't delete this campaign because it contains published quests"
+            )
+            return redirect(self.success_url)
+
+        return super().delete(request, *args, **kwargs)
+
 
 class QuestDelete(NonPublicOnlyViewMixin, UserPassesTestMixin, UpdateMapMessageMixin, DeleteView):
     def test_func(self):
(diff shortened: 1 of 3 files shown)


src/quest_manager/views.py
```diff
@@ -155,6 +155,34 @@ class CategoryDelete(NonPublicOnlyViewMixin, DeleteView):
...
+        """
+        Handle POST requests by delegating to the delete method to
+        enforce custom deletion logic and validation.
+        """
+        return self.delete(request, *args, **kwargs)
+
+    def delete(self, request, *args, **kwargs):
+        """
+        Attempt to delete the Category object only if it has no quests
+        visible_to_students. If visible quests exist, prevent deletion
```
tylerecouture, line 168: visible_to_students reference

src/quest_manager/views.py
```diff
@@ -155,6 +155,34 @@ class CategoryDelete(NonPublicOnlyViewMixin, DeleteView):
...
+        and display an error message.
+
+        Returns:
+            HttpResponseRedirect on successful deletion or redirect back with error.
+        """
+        self.object = self.get_object()
+
+        # Check for any visible quests in this category
+        if self.object.quest_set.filter(visible_to_students=True).exists():
```
tylerecouture, line 177: needs an update

tylerecouture (changes requested):

src/quest_manager/views.py
thefoxwithin: Published now

src/quest_manager/views.py
thefoxwithin: Published now

src/quest_manager/views.py
```diff
@@ -155,6 +155,34 @@ class CategoryDelete(NonPublicOnlyViewMixin, DeleteView):
...
+        """
+        self.object = self.get_object()
+
+        # Check for any visible quests in this category
+        if self.object.quest_set.filter(published=True).exists():
+            messages.error(
+                request,
+                "You can't delete this campaign because it contains published quests"
+            )
+            return redirect(self.success_url)
```
tylerecouture, line 182: I think it makes more sense on failure to return to the Campaign detail view (where it should list the quests in the campaign and they can be dealt with) instead of going back to the campaign list view.

Also, can you note in the docstring or comment HOW this view could be accessed if a published quest exists?  Is this only accessible if a user enters the url manually?

tylerecouture (changes requested):

src/quest_manager/views.py
thefoxwithin: So there was one scenario that I could think of that this would trigger,
1. Campaign has 1 unpublished quest (published=False).

2. User A visits the delete campaign page:

      - quest_count == 0 → delete button is enabled.

3. User B publishes the quest in the meantime (published=True).

4. User A submits the form (via enabled button).

5. Backend check — this should now block the deletion...
    - But it doesn’t — deletion goes through.
    - > I've tested it a few times and this is still the case.

Since this is happening I though of something different because I like this suggestion.
> I think it makes more sense on failure to return to the Campaign detail view (where it should list the quests in the campaign and they can be dealt with) instead of going back to the campaign list view.

I think it provides a better user experience, so should we instead leave the delete buttons enabled and rely on this backend redirect to lead them to the links to the quests so they can either be unpublished or taken out of the campaign? Or do we continue to disable the delete buttons and remove the backed protection?

tylerecouture: > Backend check — this should now block the 

@thefoxwithin 
Ok, that's a great explanation.  I assume it's because the quest model is cached.  I'm fine with this (please add a link to your comment here, or paste it into the code to explain).  Alternately, you could try just re-fetching the object from the database (like how we have to do in tests all the time) before checking using [refresh_from_db()](https://docs.djangoproject.com/en/5.2/ref/models/instances/#django.db.models.Model.refresh_from_db) which might be better (still good to have the explanation WHY we are doing that though)

thefoxwithin: Done

PR #1846, opened 2025-07-25
Archive update; Closes #946

### LINKED ISSUE #946: Archived Quests Remain Prerequisites

thefoxwithin:
### What?

Adds the ability to **archive quests** via a confirmation page. Archiving a quest:

- Unpublishes and hides it from students  
- Deletes all associated submissions  
- Revokes awarded XP  
- Disables re-archiving if the quest is used as a prerequisite  
- Introduces UI warnings and restrictions around these actions  

### Why?

Previously, there was no way to retire or hide a quest in a way that also removed its impact on XP and progress. This caused issues for instructors who needed to deprecate content but couldn't remove its downstream effects. This change:

- Gives instructors control over retiring quests cleanly  
- Ensures XP totals and progress reflect only active, relevant quests  
- Prevents data inconsistency from archiving prerequisite-linked quests  

https://github.com/bytedeck/bytedeck/issues/946

### How?

- Added a **confirmation page** (`quest_confirm_archive.html`) explaining the implications of archiving  
- Used `Quest.get_reliant_objects(active_only=False)` to show which quests rely on the current one  
- Disabled the "Archive" button in the template if the quest is used as a prerequisite  
- In the `QuestArchive` view:
  - Prevents archiving if the quest is a prerequisite (with message + redirect)  
  - If not blocked:
    - Sets `archived=True` and unpublishes the quest  
    - Deletes all submissions (using `_base_manager` to bypass default filters)  
    - Invalidates cached XP for all affected users  
    - Displays a success message and redirects to the archive list  

### Testing?

- Added integration tests:
  - Archiving succeeds for a non-prerequisite quest  
  - Archiving is blocked for a quest used as a prerequisite (with error message + redirect)  
  - Context includes `prereq_of` to populate the dependent quest list  
- Verified XP invalidation occurs for each user with submissions before archival  
- Confirmed UI state matches logic: archive button is disabled when it should be  

<details>
<summary>Screenshots of `quest_confirm_archive.html` when quest is a prerequisite vs not</summary>

**When the quest is a prerequisite of another model**
image

**When the quest is not a prerequisite of anything else**
image

</details>

### Anything Else?

- Archiving is **reversible**, but submissions and XP are not restored on unarchival (by design)  
- We intentionally prevent archiving based on *any* dependency (active or not), to keep the logic conservative and predictable for instructors  
- `is_used_prereq()` is evaluated at runtime and separately surfaced in the view context to power both template logic and backend enforcement  

### Review request
@tylerecouture


thefoxwithin:
This is strange, that's the third time a test has failed due to seemingly unrelated tests. Each time the failing tests have been different.

First odd failure today: https://github.com/bytedeck/bytedeck/actions/runs/16527037702/job/46742922154
Second odd failure today: https://github.com/bytedeck/bytedeck/actions/runs/16530284092/job/46753724608
Third odd failure today: https://github.com/bytedeck/bytedeck/actions/runs/16533048051/job/46762404862

The first and second were due to the notifications but the third is about the map generation, which is unrelated to the other two. What's more strange is that from what I can tell none of them are related to the PRs they're failing in.

src/quest_manager/models.py
```diff
@@ -938,7 +935,7 @@ def calculate_xp_to_date(self, user, date):
         total_xp = 0
 
         # Get all of the user's XP granting submissions for the active semester
-        submissions_qs = self.all_approved(user, up_to_date=date).grant_xp()
+        submissions_qs = self.all_approved(user, up_to_date=date).grant_xp().exclude_archived_quests()
```
tylerecouture, line 938: Since you changed the `all_approved method`, I think this is now redundant?  archived are already excluded:

image

src/quest_manager/templates/quest_manager/quest_confirm_archive.html
```diff
@@ -0,0 +1,60 @@
...
+{% block heading_inner %}Archive Quest{% endblock %}
+
+{% block content %}
+<form action="" method="post">{% csrf_token %}
+  <div class="alert alert-warning" role="alert">
+    <strong>Warning:</strong>
+    Archiving a quest will hide it from students and remove all existing submissions. Any XP awarded for this quest will be revoked. This action is reversible.
+    {% if object.is_used_prereq %}
+    <br>
+    <strong>Note:</strong> You cannot archive this quest while it is a prerequisite for other quests.
```
tylerecouture, line 13: "while it is a prerequisite for other quests." <-- What if it's a prereq for a badge or something else?  Maybe say "...while it is a prerequisite for another object (e.g. a quest or a badge)"

tylerecouture (changes requested):
Two small comments.  Also, a couple requests for the Confirmation template.

1. Can we make the quest name more prominent/obvious?  Maybe like it is on the Quest Summary page?  Or another way?
image

2. If can't archive, can you put that in it's own red well at the top above the warning well?

3. Can you display the prereqs listed like they are formatted in the other prereq lists, where they show the model AND the name?
image


Thank you!

src/quest_manager/models.py
thefoxwithin: Thanks!

src/quest_manager/templates/quest_manager/quest_confirm_archive.html
thefoxwithin: Got it, since it was so easy to add quests as prerequisites I was mostly thinking of them.

thefoxwithin:
> This needs to be covered by a test. There are a few more spots too. Please look at the automated Codecov report in the PR
> 
> image

That codecov report is old and I don't think it applies anymore, I searched for that in the code and couldn't find it. Probably because it was to restore xp to students who had submissions, but since the submissions get deleted then there's no need to try to restore xp.

tylerecouture (changes requested):
Thanks for the changes.  I noticed that the Warning and the note below it are almost duplicate.  How about we just put all those notes in the warning and remove the duplication?

PR #1853, opened 2025-07-28
Urlize fix; Closes #1826

### LINKED ISSUE #1826: urlize() only links if entire string starts with URL prefix
The `urlize()` function includes logic to detect and preserve list-like prefixes such as "1." or "a." at the start of the entire input string to avoid mistakenly linking numbered or lettered list markers as URLs.

However, this prefix detection only checks for such a prefix at the very beginning of the input text. As a result, if the text contains multiple URLs each prefixed with a number or letter followed by a dot (e.g., a numbered list of URLs), only the very first prefix is detected and preserved. Subsequent prefixed URLs within the text are processed as normal links, causing incorrect trimming or linking.

For example, given the input:

1.www.testing.com 2.www.example.com 3.www.notarealsite.com

only the "1." prefix at the start is recognized and preserved. The "2." and "3." prefixes later in the text are not recognized as prefixes and thus linked incorrectly.

This limitation prevents correct handling of multiple prefixed URLs in a single string, causing inconsistent or broken URL transformations in lists.

How it was tested:
- Submitted text containing multiple prefixed URLs in a single string.
- Observed that only the initial prefix was preserved; others were treated as part of the URL.

Expected:
- The prefix detection logic should apply to each URL or segment individually, not just the start of the entire string.
- This would enable proper linking of multiple prefixed URLs in lists without breaking formatting.

Actual:
- Only the prefix at the very beginning of the string is preserved.
- Subsequent prefixed URLs are linked incorrectly.

Recommendation:  
Update the `urlize()` function to analyze each URL candidate separately for prefixes, allowing correct preservation of list-like prefixes throughout the text while efficiently processing multiple URLs.

Image

Initial implementation of `urlize()` https://github.com/bytedeck/bytedeck/issues/930

thefoxwithin:
### What?

Refactors and improves the `urlize` function to correctly linkify plain-text URLs while preserving formatting like list prefixes and `<br>`-separated input.

### Why?

The previous implementation produced false positives in unit tests and did not accurately reflect how URLs were rendered in the actual comment system. In particular:
- Prefixes like `1.www.site.com` would wrap the entire string in a link, including the list number.
- Input with `<br>` line breaks wasn't parsed correctly.
- Existing anchor tags were not excluded from re-processing, risking broken or nested links.

https://github.com/bytedeck/bytedeck/issues/1826

### How?

- Added logic to detect list prefixes (e.g. `1.`, `a.`) using a regular expression and preserve them outside the linkified portion.
- Split text by `<br>` and process each line independently.
- Used `bleach.linkify` with a custom `trim` callback to optionally shorten display text.
- Skipped processing if the text already includes an anchor tag.

### Testing?

- Rewrote and expanded the test suite to more accurately simulate real comment input:
  - Covered list-prefixed URLs.
  - Verified preservation of existing anchor tags.
  - Confirmed trimming behavior and safe handling of edge cases like `javascript:` URLs.
- Confirmed output is consistent with expected link wrapping and does not corrupt HTML.

https://github.com/user-attachments/assets/aa9bce37-9e96-49e4-99f8-2a0bec706dd9

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Added enhanced URL linkification that converts URLs in text to clickable links while preserving list prefixes and formatting.
  * Introduced optional trimming of long URLs in link display text.

* **Bug Fixes**
  * Improved handling to avoid double-linking when anchor tags already exist.
  * Ensured that potentially unsafe URLs (e.g., javascript:) are not linkified.

* **Tests**
  * Added comprehensive tests for URL linkification, formatting preservation, and security handling.


src/utilities/tests/test_html.py
```diff
@@ -23,3 +24,199 @@ def test_textify_donot_ignore_links(self):
...
+        and the URL text is preserved.
+        """
+        text = "Visit www.example.com"
+        result = urlize(text)
+        self.assertIn('<a href="http://www.example.com"', result)
+        self.assertIn('www.example.com', result)
+
+        text = "1.www.example.com"
+        result = urlize(text)
+        self.assertIn('<a href="http://www.example.com"', result)
```
tylerecouture, line 54: All these `assertIn` seem redundant with the assertTrue after it, since if the second one is true, obviously it contains the substring tested the line before?

tylerecouture (changes requested):

PR #1854, opened 2025-07-29
Export quest to library; Closes #1847

### LINKED ISSUE #1847: Create a function to share a quest to the library
Much like the importer, when a quest is shared to the library set `published=False` so the owner of the library can review the quest and decide whether it will be available or not. This function should check the library for a matching `import_id` before trying to share.

Since this function is sharing a single quest, should it mimic the importer and only import the quest and leave whatever campaign it may be attached to locally behind?

Should the sharing functions get their own file in the library folder, or does it make sense for them to share the importer.py file?

Specifically part 3. "A way for deck owners to SHARE (push) a quest or a campaign of quests to the library."

thefoxwithin:
### What?

Adds the ability for users to export quests from a local tenant to the shared library, with appropriate permissions, a confirmation step, and a notification to the library owner from the library schema.

### Why?

To allow deck owners (and optionally staff members) to share their locally created quests with others by exporting them to the central shared library.

https://github.com/bytedeck/bytedeck/issues/1847

### How?

- Implemented `ExportQuestView` as a `FormView` that confirms the export before proceeding.
- Added `exporter.py` whose only function at the moment is `export_quest_to_library()`
- Added `require_export_permission()` to enforce that:
  - Only the quest owner can export unless the new config flag `allow_staff_export` is enabled.
  - The quest must not already exist in the library (based on `import_id` uniqueness).
- If the quest is exportable:
  - It is copied into the library schema (via `export_quest_to_library()`).
  - A success message is shown and the user is redirected back to the local quest.
  - A notification is sent (from `deck_ai` to `deck_owner`) in the library schema.
    - Includes fallback handling to ensure the sender and recipient aren't the same,
      since notifications won't be delivered in that case.
- If the quest already exists or the user lacks permission:
  - A `PermissionDenied` is raised.

### Testing?

Added test coverage to `QuestLibraryTestsCase` in `test_views.py`:
- Export succeeds for owner and is denied for non-owner.
- Staff can export only if config flag is enabled.
- Export is denied if a quest with the same `import_id` already exists in the library.
- Notification is sent to library owner upon successful export.
- Proper redirection and success message on completion.

<details>
<summary>Screenshots of the export confirmation page</summary>

If the quest already exists in the library
image

Exportable quest confirmation form
image

Validation error when skipping the Creative Commons checkbox
image

</details>

The new SiteConfig field in the owner only section
image

These are the current links used in the confirmation UI — please let me know if they need to be updated:
https://creativecommons.org/licenses/by-sa/4.0/
https://github.com/bytedeck/bytedeck/discussions

### Anything Else?
No buttons yet

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Added the ability for deck owners (and optionally staff) to export quests to a shared library, with a confirmation interface and license agreement.
  * Introduced a permission setting to allow staff members to export quests if enabled by the deck owner.
  * Implemented notifications for deck owners upon successful export to the shared library.

* **Bug Fixes**
  * Prevented duplicate exports of quests with the same import ID to the shared library.

* **Tests**
  * Added comprehensive tests for export permissions, export flows, duplicate prevention, and notification creation.

* **Documentation**
  * Updated the user interface to clarify export permissions and license requirements.


tylerecouture:
@thefoxwithin I think you should chunk this more.  For examle, adding the siteconfig option and migrations can be it's own stand alone PR

src/library/views.py
```diff
@@ -283,6 +292,85 @@ def post(self, request, campaign_import_id):
...
+                sender = (
+                    User.objects
+                    .filter(is_active=True, is_staff=True)
+                    .exclude(id=recipient.id)
+                    .order_by('id')
+                    .first()
+                )
+
+            # Only send the notification if both sender and recipient are valid.
+            # This avoids silent failures in setups with only one user.
```
tylerecouture, line 356: Why would there only be one user?  A new deck has at least two, the admin and the owner, doesn't it?

src/library/views.py
```diff
@@ -283,6 +292,85 @@ def post(self, request, campaign_import_id):
...
+            export_quest_to_library(source_schema=source_schema, quest_import_id=quest.import_id)
+
+            # Get the newly exported quest
+            exported_quest = Quest.objects.get(import_id=quest.import_id)
+
+            config = SiteConfig.get()
+            recipient = config.deck_owner
+            sender = config.deck_ai
+
+            # Fallback: If sender and recipient are the same, the notification system will skip sending.
```
tylerecouture, line 344: Can you explain why this would happen and why its a problem?

src/library/views.py
tylerecouture: This seems like something to fix in the notification app rather than here.  Maybe create an issue for it if it's a legit problem.

src/siteconfig/migrations/0030_siteconfig_allow_staff_export.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('siteconfig', '0029_siteconfig_map_auto_update'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='siteconfig',
+            name='allow_staff_export',
+            field=models.BooleanField(default=False, help_text='If enabled, staff will be able to export quests and campaigns to the Shared Library.', verbose_name='Allow staff to export quests to the library'),
```
tylerecouture, line 16: Not clear the status when off.  So maybe change to: "If enabled, staff will be able to export quests and campaigns to the Shared Library, otherwise only the owner can export quests to the Library"

 (if you change this please squash the migrations so we only have one, or delete this migration so it regenerates it fresh... stupid that it generates a migration when helptext is changed)

tylerecouture: Can we hyperlink this helptext to the Library help/wiki page?

tylerecouture (changes requested):
Some minor comments.

src/library/views.py
tylerecouture: I'm confused why these are silent failures.  What is failing?  Why would you want to send a notification to yourself?

Can you add some context? Which deck is this in, the library or the local deck sharing the quest?  Can you add docstrings to these methods with details?

src/library/views.py
tylerecouture: No, this prevents someone from getting a notification from themselves.  Why are we trying to force this notification?  There will always be an owner and an admin on every deck as well.  And if for some reason there is only one, then seems like a notification is redundant anyway and we can just let it skip.

tylerecouture (changes requested):

src/library/views.py
tylerecouture: What's the problem with sending it via the deck_ai user? That's the purpose of that user. If they're the same user as the owner then no notification.

What specifically is this one a notification for?

thefoxwithin: It's a notification to the Library Owner that a Quest has been exported

tylerecouture: Ok. Let's change this and send this notification to all staff on the Library deck, coming from the deck_ai.

If using the deck_ai causes problems with testing, just create an AI user and assign it the AI role for the test.

PR #1863, opened 2025-08-10
Add export quest button; Closes #1848

### LINKED ISSUE #1848: Add the button and confirmation page for sharing a single quest to the library
Quests should have a button to allow the Owner of the library to share quests to the library. (Is this going to be limited to the owners or will staff members also have this option?)

This button should lead to a confirmation page, this confirmation page should describe that the quest is being shared under Creative Commons Attribution Share Alike license. They need to agree to that before being able to share the quest. Also on the confirmation page should be another button to "Share" the quest to the library that becomes available after they've agreed to the CC license, along with a cancel button.

If there is a quest on the library that already has the same `import_id` the share button will be disabled for now. We can handle options in later PRs.

thefoxwithin:
### What?
Added an export quest button that is only visible to authorized users outside the library schema. This button allows exporting quests to the shared library.

### Why?
To ensure only users with proper permissions—either the deck owner or staff when allowed—can export quests. Also, to prevent showing the export button when browsing quests already in the library schema.

https://github.com/bytedeck/bytedeck/issues/1848

### How?
- Created a `can_export` template tag that checks if the current user can export based on site config and deck ownership.
- Updated quest templates to conditionally render the export button using the `can_export` tag.
- Added logic to hide the button when the current schema is the library schema.

### Testing?
- Verified the export button appears for the deck owner and staff (if allowed) on non-library schemas.
- Verified the export button is hidden when viewing quests in the library schema.
- Manual tests clicking the export button confirm correct routing and permissions.

image
image
image
image
image

### Anything Else?
Draft quests can still be exported at this time, I was unsure if we wanted to allow that or not. Do we want the templatetag check if the quest already exists on the library as well? They won't be able to export either way.

### Review request
@tylerecouture

## Summary by CodeRabbit

- New Features
  - Added an “Export this quest to the Library” button for eligible users, visible outside the Library view and placed alongside existing actions.

- Bug Fixes
  - After exporting, users are redirected to the main Quests list (instead of drafts).

- Tests
  - Expanded tests to verify export button visibility, can_export behavior across roles and schema contexts, and the updated redirect.

### DIFF
--- a/src/quest_manager/templates/quest_manager/buttons.html
+++ b/src/quest_manager/templates/quest_manager/buttons.html
@@ -52,6 +52,12 @@
             title="Archive this quest" >
             <i class="fa fa-fw fa-archive"></i>
           </a>
+          {% if can_export %}
+            <a class="btn btn-default" href="{% url 'library:export_quest' q.import_id %}" role="button"
+              title="Export this quest to the Library">
+              <i class="fa fa-fw fa-upload"></i>
+            </a>
+          {% endif %}
           <a class="btn btn-danger" href="{% url 'quests:quest_delete' q.id %}" role="button"
             title = "Delete this quest" >
             <i class="fa fa-trash-o"></i>
--- a/src/library/tests/test_views.py
+++ b/src/library/tests/test_views.py
@@ -372,8 +372,7 @@ class QuestLibraryTestsCase(LibraryTenantTestCaseMixin):
         url = reverse('library:export_quest', args=[self.local_quest.import_id])
         response = self.client.post(url)
 
-        self.assertEqual(response.status_code, 302)
-        self.assertEqual(response.url, reverse('quests:drafts'))
+        self.assertRedirects(response, reverse('quests:quests'))
 
         with library_schema_context():
             exported_quest = Quest.objects.get(import_id=self.local_quest.import_id)
--- a/src/library/views.py
+++ b/src/library/views.py
@@ -352,7 +352,7 @@ class ExportQuestView(View):
             quest_import_id (UUID): The import ID of the quest to export.
 
         Returns:
-            HttpResponseRedirect: Redirect to the drafts list after successful export.
+            HttpResponseRedirect: Redirect to the main quests list after successful export.
 
         Raises:
             PermissionDenied: If a quest with the same import ID already exists in the library.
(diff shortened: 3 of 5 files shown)


src/quest_manager/templates/quest_manager/buttons.html
```diff
@@ -52,6 +52,13 @@
             title="Archive this quest" >
             <i class="fa fa-fw fa-archive"></i>
           </a>
+          {% can_export as export_allowed %}
+          {% if export_allowed %}
```
tylerecouture, line 56: Why are you renaming the variable?  Why not just say `{% if can_export %}` ?

src/quest_manager/templates/quest_manager/buttons.html
```diff
@@ -52,6 +52,13 @@
...
+            <a class="btn btn-default" href="{% url 'library:export_quest' q.import_id %}" role="button"
+              title="Export this quest to the library">
```
tylerecouture, line 58: Capital L for Library.

src/quest_manager/templatetags/quest_tags.py
```diff
@@ -14,3 +16,33 @@ def is_hidden(quest, user):
     if not user or not quest:
         return None
     return user.profile.is_quest_hidden(quest)
+
+
+@register.simple_tag(takes_context=True)
+def can_export(context):
```
tylerecouture, line 22: What's the reason for doing this in the template instead of passing it as a context variable in the view?

src/quest_manager/templatetags/quest_tags.py
```diff
@@ -14,3 +16,33 @@ def is_hidden(quest, user):
...
+
+    Returns False if:
+    - The current schema is the library schema (i.e., user is already in the library), or
+    - The request or user is not available or not authenticated.
+    """
+    request = context.get('request')
+    if not request or not getattr(request, 'user', None) or not request.user.is_authenticated:
+        return False
+
+    current_schema = getattr(getattr(request, 'tenant', None), 'schema_name', None) or getattr(request, 'schema_name', None)
```
tylerecouture, line 39: don't we have a method to get the current deck?

src/quest_manager/templatetags/quest_tags.py
```diff
@@ -14,3 +16,33 @@ def is_hidden(quest, user):
...
+    Returns True if:
+    - The current user is the deck owner for this site, or
+    - The site configuration allows staff to export quests and user is staff.
+
+    Returns False if:
+    - The current schema is the library schema (i.e., user is already in the library), or
+    - The request or user is not available or not authenticated.
+    """
+    request = context.get('request')
+    if not request or not getattr(request, 'user', None) or not request.user.is_authenticated:
```
tylerecouture, line 36: how is an unauthenticated user accessing this page? how would there not be a request?  What are all these things checking?  Are they all tested?

tylerecouture (changes requested):

PR #1867, opened 2025-08-12
Add export campaign feature; Closes #1849

### LINKED ISSUE #1849: Create a function to share a campaign to the library
Much like the campaign importer, when a campaign is shared to the Library it should set `active=False`. (This may be changed to `published=False` in the future.) This function should check the Library for an existing campaign with a matching `import_id ` before proceeding.

When sharing a campaign, the associated quests should also be shared. The function should preserve the order and prerequisites between quests, except for the starting prerequisite on the first quest, which should be removed in case the referenced quest doesn't exist in the Library. Furthermore all quests in the campaign should set `published=False`.

Do we want the share function detect if the quests already exist and identify them like the importer does?
If we do want the existing quests to be detected, how do we want to handle them? Would we want them to preserve their published state (i.e. if they're published or not) on the library after being shared in a campaign?

Specifically part 3. "A way for deck owners to SHARE (push) a quest or a campaign of quests to the library."

thefoxwithin:
### What?
Add the ability to export a full campaign (category) and all its quests from a tenant schema to the shared library schema, including confirmation views and permission checks.

### Why?
This feature enables users to share campaigns and their quests centrally in the shared library, improving content reuse and management across tenants. 

https://github.com/bytedeck/bytedeck/issues/1849

### How?
- Implement `ExportCampaignView` with GET confirmation and POST export logic.
- Add permission checks to restrict export to deck owners or staff if `allow_staff_export=True`.
- Use import/export logic with `QuestResource` to transfer campaigns and quests.
- Notify staff users after a successful export.
- Add tests covering permission enforcement, export success, and error cases.

### Testing?
- Unit tests verify export permission scenarios for deck owners and staff.
- Tests ensure forbidden access when a campaign already exists in the library.
- Tests confirm successful export redirects and notifications are sent.
- Manual testing of confirmation pages and export flows in UI.

#### With no Conflicts
image

#### One or more Quests exist in the Library already
image

#### Campaign already exists in the Library already
image

#### Success notification displayed on Library
image

> The success notification uses the same format as the quests, Library `deck_ai`, user who exported, the deck it was exported from, and finally the name of the campaign. Clicking the notification will send you to the Campaign detail page

### Anything Else?
- Future improvements include handling duplicate quests in library by creating copies.

### Review request
@tylerecouture

## Summary by CodeRabbit

- New Features
  - Export entire campaigns (category + published quests) to the shared library with a confirmation page, license acceptance, conflict warnings, routing, and post-export notifications.

- Improvements
  - Centralized export permission checks; clearer conflict warnings and library indicators in lists; exported items are created unpublished in the library; isolated cross-tenant operations; refined redirects and success messaging.

- Tests
  - Expanded coverage for permissions, conflict handling, successful quest/campaign exports, notifications, and cross-tenant scenarios.

### DIFF
--- a/src/library/exporter.py
+++ b/src/library/exporter.py
@@ -1,6 +1,6 @@
 from django_tenants.utils import schema_context
 from quest_manager.admin import QuestResource
-from quest_manager.models import Quest
+from quest_manager.models import Quest, Category
 
 from .utils import library_schema_context
 from django.core.exceptions import ValidationError
@@ -50,5 +50,58 @@ def export_quest_to_library(*, source_schema, quest_import_id):
         imported_quest.published = False
         imported_quest.full_clean()
         imported_quest.save()
+
+    return result
+
+
+def export_campaign_to_library(*, source_schema, campaign_import_id):
+    """
+    Exports a campaign (Category) and all its quests from the given tenant schema
+    into the shared library schema.
+
+    All exported quests and the campaign will be unpublished in the library by default.
+    This ensures the shared library remains a neutral, reviewable repository.
+
+    Args:
+        source_schema (str): The tenant schema where the campaign currently resides.
+        campaign_import_id (UUID): The import ID of the campaign to export.
+
+    Returns:
+        ImportResult: The result of the import operation into the library schema.
+
+    Raises:
+        Category.DoesNotExist: If the campaign is not found in the source schema.
+        ValidationError: If no published quests exist for the campaign.
+    """
+    with schema_context(source_schema):
+        category = Category.objects.get(import_id=campaign_import_id)
+        quests = Quest.objects.filter(published=True, campaign=category)
+        if not quests.exists():
+            raise ValidationError("Cannot export a campaign without any published quests.")
+        export_data = QuestResource().export(quests)
+
+    # Prepare visibility map for library: all quests unpublished
+    exported_visibility_map = {str(q.import_id): False for q in quests}
+
+    with library_schema_context():
+        try:
+            result = QuestResource().import_data(
+                export_data,
+                dry_run=False,
+                raise_errors=True,
+                use_transactions=True,
+                import_campaign=True,
+                local_visibility_map=exported_visibility_map,
+            )
+        except (ValidationError, IntegrityError) as e:
+            # Raise a clearer validation error with context
+            raise ValidationError(f"Failed to import campaign to library schema: {e}") from e
+
+        # Explicitly unpublish imported campaign
+        imported_category = Category.objects.filter(import_id=campaign_import_id).first()
+        if imported_category:
+            imported_category.published = False
+            imported_category.full_clean()
+            imported_category.save(update_fields=['published'])
 
     return result
--- a/src/library/templates/library/confirm_export_campaign.html
+++ b/src/library/templates/library/confirm_export_campaign.html
@@ -1,0 +1,59 @@
+{% extends "quest_manager/base.html" %}
+
+{% block head_title %}Library | Export Campaign to Shared Library{% endblock %}
+
+{% block heading %}
+  <i class="fa fa-book pull-right"></i>
+  {% block heading_inner %}Export Campaign to Library{% endblock %}
+{% endblock %}
+
+{% block content %}
+
+  {% if existing_campaign %}
+    <div class="alert alert-danger">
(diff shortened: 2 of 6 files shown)


src/library/exporter.py
```diff
@@ -52,3 +52,59 @@ def export_quest_to_library(*, source_schema, quest_import_id):
...
+                use_transactions=True,
+                import_campaign=True,
+            )
+        except (ValidationError, IntegrityError) as e:
+            # Raise a clearer validation error with context
+            raise ValidationError(f"Failed to import campaign to library schema: {e}") from e
+
+        # Unpublish exported quests
+        exported_quests = Quest.objects.filter(import_id__in=[q.import_id for q in quests])
+        for eq in exported_quests:
```
tylerecouture: Code comment with context for why this loop is needed.   Will staff get a notification for EVERY quest in the campaign that is sent to the library?  Maybe we don't need signals to trigger on this on the library?  Need some more comments why we are doing it this way and what the consequences are.

tylerecouture (approved):

src/library/exporter.py
tylerecouture: I'm curious how much lag there is when doing this for real, since it's not asynchronous, how slow is this process?  If I export a campaign of 20 quests, does it take 2 seconds or 10 or 30?

tylerecouture (changes requested):
A couple comments about efficiency.  Can you test this and how long it takes in testing to export a big campaign of like 30 quests?  try auto generating some big campaigns using the admin tool

src/library/exporter.py
thefoxwithin: With 31 quests it took a little over 6 seconds to complete. Approximately 6.3 to be more precise, which works out to about 0.20 seconds per quest. That seems like a reasonable ballpark, but let me know if that feels acceptable.

I also tested the scenario where every quest already existed on the library. In that case, the export took about 6.77 seconds, a small increase that works out to 0.21 seconds per quest. This suggests the existence check is only a minor additional cost.

To see how this scales, I added 19 more quests (total of 50), with 31 of them already present on the library. The export finished in 10.44 seconds, which is consistent with the ~0.21 seconds per quest observed earlier.

As an experiment, I also tried switching the “unpublish quests” step from looping through each quest and calling .save() to a single bulk update that skips model validation and signals. Surprisingly, this was slower, about 11.40 seconds for the same 50-quest export. On investigation, this makes sense: the loop was already very cheap (just one boolean flip and update per row), and the real bottleneck in export is the serialization/import logic and database writes, not this small unpublish step. A bulk update might even incur extra overhead if the published field is indexed, since PostgreSQL has to touch all affected rows at once. So the performance difference here is negligible, and the choice between bulk update vs loop is more about clarity and signal handling than speed.

> Will staff get a notification for EVERY quest in the campaign that is sent to the library?

No, only a single notification for the campaign itself is sent. Clicking it takes staff to the campaign page where they can see all the included quests.

Looking at it a bit closer the exporter and importer are doing it differently, I'll change the exporter to act more like the importer and this should remove the need to loop over all the quests. (The importer doesn't even do that which is what tipped me off that they're behaving differently)

thefoxwithin: After updating it to be a lot more like the importer it appears there is no change in speed.

still about 10.5 seconds for exporting a campaign with 50 quests.

tylerecouture: Anything that takes more than a few seconds should be done asychronously as a celery task.  I don't think you've done any of these yet, so might as well learn it!

You'll need to update the message so that it lets the user know the quests are being pushed to the library in the background.

Create a celery tasks and run it.  The user will have their next page load right away and the actually processing will be done concurrently.

tylerecouture: Alternately, you an just leave the message as is, is there any difference to the user whether the quests appear instantly in the library or take 10 seconds?

tylerecouture: An example if you want one, is the "regenerate all maps" view `regenerate_all` in djcytoscape/views.py.

It checks if there are more than 5 maps to regenerate, it sends it as a celery task/parallel instead of doing it in series.

tylerecouture: One more thing, is that local development environment might be a lot slower than production, so maybe test it in production first.  If uploading 30 quests takes 4 seconds or less in production, we can leave it as is.

PR #1881, opened 2025-08-17
add archived quests to the delete view queryset and test it; Closes #1874

### LINKED ISSUE #1874: Can't delete an Archived quest
This one is more a minor inconvenience since you can unarchive and then delete the quest, however it might be worth adding the ability to delete archived quests without the extra steps.

Image

Image

When trying to delete it by using the url

Image

Which means that archived quests aren't included in the deletion queryset.

thefoxwithin:
### What?  
- Updated the `QuestDelete` view to include archived quests in its queryset by using `Quest.objects.all_including_archived()`.  
- Added test coverage to ensure teachers can delete archived quests.  

### Why?  
- Previously, archived quests were excluded from the queryset by default, so attempting to delete them would raise `404`.  
- This ensures that `QuestDelete` works correctly for both active and archived quests, matching expected behavior.  

[#1874](https://github.com/bytedeck/bytedeck/issues/1874)  

### How?  
- Overrode `get_queryset()` in `QuestDelete` to call `Quest.objects.all_including_archived()`.  
- Added a section in `test_teacher_can_create_and_delete_quests` to create an archived quest and verify it can be deleted.  

### Testing?  
- Unit test confirms that teachers can delete both active and archived quests.  
- Existing tests for permission checks and deletion of active quests remain unaffected.  

image

image

### Anything Else?  
- Ensures consistency between manager methods and CBV behavior.  

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Staff can now delete archived quests from the quest management interface; archived quests show a Delete option matching the active-quest workflow and redirect appropriately after confirmation.

* **Tests**
  * Test coverage expanded to create and delete archived quests, verifying archived-quest visibility and successful removal via the delete view.

### DIFF
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -225,6 +225,12 @@ class QuestDelete(NonPublicOnlyViewMixin, UserPassesTestMixin, UpdateMapMessageM
     def test_func(self):
         return self.get_object().is_editable(self.request.user)
 
+    def get_queryset(self):
+        """
+        Include archived quests so that archived quests can be deleted using this views.
+        """
+        return Quest.objects.all_including_archived()
+
     model = Quest
     success_url = reverse_lazy("quests:quests")
 
--- a/src/quest_manager/templates/quest_manager/buttons.html
+++ b/src/quest_manager/templates/quest_manager/buttons.html
@@ -24,6 +24,11 @@
         </button>
       </form>
 
+      <a class="btn btn-danger" href="{% url 'quests:quest_delete' q.id %}" role="button"
+        title = "Delete this quest" >
+        <i class="fa fa-trash-o"></i>
+      </a>
+
     {% elif  request.user.is_staff or request.user == q.editor %}
         {% if is_library_view %}
           <a href="{% url 'library:import_quest' q.import_id %}"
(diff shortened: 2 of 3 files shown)


src/quest_manager/views.py
```diff
@@ -225,6 +225,12 @@ class QuestDelete(NonPublicOnlyViewMixin, UserPassesTestMixin, UpdateMapMessageM
     def test_func(self):
         return self.get_object().is_editable(self.request.user)
 
+    def get_queryset(self):
+        """
+        Include archived quests using the manager's helper method.
```
tylerecouture: The WHAT is usually obvious from the code itself, in this case the docstring just repeats what the code is already saying.  More important and not at all obvious is the WHY.  Make sure to explain WHY in your docstrings and code comments, it's much more important, because you or me or anyone else in the future needs to know that.

tylerecouture (changes requested):

src/quest_manager/views.py
thefoxwithin: Got it, changed the docstring to
"""
Include archived quests so that archived quests can be deleted using this views.
"""

tylerecouture (approved):

Issue #1878, opened 2025-08-17
Highlight warning message


thefoxwithin:
This message should be turned into a red well

<img width="1110" height="591" alt="Image" src="https://github.com/user-attachments/assets/100d3903-6238-44a2-b340-bbd49c3826d3" />

Like this one when importing a campaign

<img width="1081" height="140" alt="Image" src="https://github.com/user-attachments/assets/1d2b8603-02e4-4018-b5aa-4c9b102dede7" />

PR #1883, opened 2025-08-18
added alert alert-danger to the warning message

thefoxwithin:
### What?
Highlighted the warning message displayed by `confirm_import_quest.html` when the quest already exists locally.
 
### Why?
The rest of the confirmation pages use alert wells to display the warning messages. The message didn't feel like a warning and wasn't very visible before.

https://github.com/bytedeck/bytedeck/issues/1878
### How?
```html
    <div class="alert alert-danger">
      Your deck already contains a quest with a matching Import ID.  Overwriting existing quests is not yet supported.
      If you still want to import this quest from the Library, you need to delete the existing quest from your deck
      or change it's Import ID:
    </div>
```
the div used to be <p>

### Testing?
Manually tested that it showed up as a red well.

image

### Review request
@tylerecouture

## Summary by CodeRabbit

- **Style**
  - Updated the warning message on the quest import confirmation page to use a Bootstrap danger alert for clearer visual emphasis.
  - Improves readability and consistency with the app’s UI components, making critical warnings more noticeable to users.

### DIFF
--- a/src/library/templates/library/confirm_import_quest.html
+++ b/src/library/templates/library/confirm_import_quest.html
@@ -3,11 +3,11 @@
 {% block heading_inner %}Import Quest from Library{% endblock %}
 {% block content %}
   {% if local_quest %}
-    <p>
+    <div class="alert alert-danger">
       Your deck already contains a quest with a matching Import ID.  Overwriting existing quests is not yet supported.
       If you still want to import this quest from the Library, you need to delete the existing quest from your deck
       or change it's Import ID:
-    </p>
+    </div>
     <div class="well">
       <p>
         Existing Quest: <a href="{{ local_quest.get_absolute_url }}">{{ local_quest.name }}</a>


src/library/templates/library/confirm_import_quest.html
```diff
@@ -3,11 +3,11 @@
...
-    <p>
+    <div class="alert alert-danger">
       Your deck already contains a quest with a matching Import ID.  Overwriting existing quests is not yet supported.
       If you still want to import this quest from the Library, you need to delete the existing quest from your deck
       or change it's Import ID:
-    </p>
+    </div>
     <div class="well">
       <p>
         Existing Quest: <a href="{{ local_quest.get_absolute_url }}">{{ local_quest.name }}</a>
```
tylerecouture, line 13: Some confusion: Why does the link here say [Shared Library]  before the quest.  What does that mean?  Seems like a link to the library quest instead of the local quest, but the code here seems to indicate it's a local url?

image

tylerecouture (changes requested):

src/library/templates/library/confirm_import_quest.html
thefoxwithin: Sorry for the confusion, on this deck I've deleted and then imported the [Shared Library] quests. This was a while ago to test the campaign import functions. It is actually pointing to the local quest.

thefoxwithin: I've renamed them on that deck so this confusion won't happen again.

tylerecouture: Oh, that was the actual quest name...hahaha.

PR #1885, opened 2025-08-18
added buttons to export campaigns, and restricted exporting to campaigns with published quests; Closes #1850

### LINKED ISSUE #1850: Add the button and confirmation page for sharing a campaign to the Library
Campaigns should have a button to allow the deck Owner to share entire campaigns to the Library. (Is this going to be limited to the owners or will staff members also have this option?)

This button should lead to a confirmation page that explains the campaign is being shared under the Creative Commons Attribution Share Alike license. The user must agree to these terms before being able to share the campaign. On the confirmation page, there should be a "Share" button that becomes enabled only after agreeing to the license, along with a cancel button.

If there is a campaign already on the Library with the same `import_id`, the share button should be disabled for now. Handling of duplicates or overwrite options can be implemented in later PRs.

After sharing, a notification should be sent to Library staff so they can review and publish the imported campaign and quests manually.

Let me know if there's anything I'm missing for the confirmation page.

thefoxwithin:
### What?
Add buttons that lead to the export confirmation page. Much like importing, exporting a campaign that has no published quests is not allowed. The way we got around this with importing is not showing campaigns without published quests to decks other than the library. With exporting there's an explanation when hovering over the button.

### Why?
To allow users to get to the export confirmation page in ways that are explicitly protected and don't force url manipulation.

https://github.com/bytedeck/bytedeck/issues/1850

### How?
If the user is allowed to export (i.e. they are the owner or they are staff and `allow_staff_export=True`) the button is shown. If the campaign has no published quests then the button is disabled with help text explaining why.

### Testing?

- Wrote a test `test_export_get__disables_button_if_no_quests` that:
  - Creates a campaign with no published quests.
  - Renders the export confirmation page.
  - Confirms the page shows the alert message.
- Manually verified that campaigns with published quests display an **enabled export button** and navigate correctly to the confirmation page.

image

image

image

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Added an Export action for campaigns in the detail view and campaign list (visible when export is allowed); disabled and accessible when no published quests.  

* **Bug Fixes / UX**
  * Confirmation screen shows a red alert: "You cannot export a campaign with no published quests." Share button now has an explicit DOM id for client-side use.  

* **Permissions**
  * New site-config permission check controlling who can export to the library.  

* **Tests**
  * Added tests for empty-campaign export UI and export permission logic.

### DIFF
--- a/src/library/templates/library/confirm_export_campaign.html
+++ b/src/library/templates/library/confirm_export_campaign.html
@@ -13,6 +13,10 @@
     <div class="alert alert-danger">
       A campaign with the same import ID already exists in the shared library. Exporting again is not allowed.
     </div>
+  {% elif category_quest_count == 0 %}
+    <div class="alert alert-danger">
+      You cannot export a campaign with no published quests.
+    </div>
   {% else %}
     <p>Please read and agree to the following license before sharing this campaign:</p>
 
@@ -41,8 +45,11 @@
         </label>
       </div>
 
-      <button type="submit" class="btn btn-primary">Share Campaign to Library</button>
+      <button id="confirm-export-button" type="submit" class="btn btn-primary">
+        Share Campaign to Library
+      </button>
       <a href="{% url 'quests:categories' %}" class="btn btn-info">Cancel</a>
+
     </form>
     {% if library_quest_import_ids %}
         <div class="alert alert-danger panel-top">
--- a/src/quest_manager/templates/quest_manager/category_detail.html
+++ b/src/quest_manager/templates/quest_manager/category_detail.html
@@ -19,6 +19,22 @@
                 <i class="fa fa-edit"></i>
             </a>
 
+            {% if can_export %}
+              <a class="btn btn-default" role="button"
+                {% if object.quest_count != 0 %}
+                  href="{% url 'library:export_category' object.import_id %}"
+                  title="Export this Campaign to the Library"
+                {% else %}
+                  title="Cannot export without any published quests"
+                  disabled
+                  tabindex="-1"
+                  aria-disabled="true"
+                {% endif %}
+              >
+                <i class="fa fa-fw fa-upload"></i>
+              </a>
+            {% endif %}
+
             {% if not object.published %}
                 <form method="post" action="{% url 'quests:category_publish' object.id %}" class="preview-action-form">
                     {% csrf_token %}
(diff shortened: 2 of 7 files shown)


src/library/templates/library/confirm_export_campaign.html
```diff
@@ -41,8 +41,17 @@
...
-      <button type="submit" class="btn btn-primary">Share Campaign to Library</button>
+      <button id="confirm-export-button" type="submit" class="btn btn-primary"
+              {% if category_quest_count == 0 %}disabled{% endif %}>
+        Share Campaign to Library
+      </button>
       <a href="{% url 'quests:categories' %}" class="btn btn-info">Cancel</a>
+
+      {% if category_quest_count == 0 %}
+        <p class="text-danger mt-2">
+          You cannot export a campaign with no published quests.
```
tylerecouture: Can you put this in a well at the top of the page similar how we did with the other detail pages?

src/quest_manager/views.py
```diff
@@ -89,8 +89,16 @@ def get_queryset(self):
     def get_context_data(self, *args, **kwargs):
         context_data = super().get_context_data(*args, **kwargs)
 
+        site_config = SiteConfig.get()
+        current_schema = getattr(self.request.tenant, "schema_name", None)
+        can_export = (
```
tylerecouture: instead of duplicating this can_export code (here and below), let's put this in a function on the SiteConfig model.  Then it can have a proper unit test too.

`can_export = site_config.can_user_export_to_library(self.request.user)`

tylerecouture (changes requested):

src/library/templates/library/confirm_export_campaign.html
thefoxwithin: Changed it to be the well at the top, but also removes the rest of the confirmation page just like when the campaign already exists on the library.

image

image

I did this because the red well looked awkward if it was just included with the rest of the page. Here's what I mean
image

src/quest_manager/templates/quest_manager/category_detail.html
```diff
@@ -19,6 +19,22 @@
                 <i class="fa fa-edit"></i>
             </a>
 
+            {% if can_export %}
+              <a class="btn btn-default" role="button"
+                {% if object.quest_count != 0 %}
+                  href="{% url 'library:export_category' object.import_id %}"
+                  title="Export this Campaign to the Library"
```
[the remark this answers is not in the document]
thefoxwithin: I thought about adding in something like this, but there's no way that you can get to the export campaign button without published quests in the campaign without either making your own view, or doing more complicated manipulation.

And so I thought about earlier conversations where things like that were rejected due to being unnecessary fail safes.

tylerecouture (approved):

PR #1889, opened 2025-08-24
updated can_user_export_to_library to check the current_schema and return false when on library; Closes #1888

### LINKED ISSUE #1888: Remove export buttons from the library
The export buttons are still visible when on the library in a few places. When viewing the details of a quest, and on all of the campaign pages where buttons appear. The export buttons should never be visible while on the library because they're redundant.

thefoxwithin:
### What?
Added a check to `can_user_export_to_library` to return `False` if the user is currently on the library schema.

### Why?
Export buttons should not appear on the library deck. On the library schema, these buttons are redundant and would not work correctly. This improves the UI and prevents confusion.

https://github.com/bytedeck/bytedeck/issues/1888

### How?
- Updated `can_user_export_to_library` to accept a `current_schema` argument.
- Added a check that returns `False` when `current_schema` is `"library"`.
- Updated the tests to verify behavior on both regular tenant schemas and the library schema.

### Testing
- Verified that the export button appears for the deck owner and allowed staff on regular schemas.
- Verified that the export button does **not** appear for any user on the library schema.
- Tests for anonymous and non-staff users continue to pass.

### Review request
@tylerecouture

## Summary by CodeRabbit

- New Features
  - Export permissions are now schema-aware, improving clarity and consistency when exporting to the Shared Library.
- Bug Fixes
  - Exporting is blocked when browsing the Shared Library.
  - Deck owners can export to the Shared Library only when not on the Library schema.
  - Staff exports respect the “allow staff export” setting and are blocked on the Library schema.
  - Unauthenticated and non-staff users cannot export to the Shared Library.

### DIFF
--- a/src/siteconfig/models.py
+++ b/src/siteconfig/models.py
@@ -15,6 +15,7 @@ from django.conf import settings
 from django_tenants.utils import get_public_schema_name, schema_context
 from redis import exceptions as redis_exceptions
 
+from library.utils import get_library_schema_name
 from utilities.models import RestrictedFileField
 
 User = get_user_model()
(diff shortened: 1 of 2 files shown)


tylerecouture (approved):
Nice catch.

PR #1890, opened 2025-08-25
Handling of exporting a campaign with quest conflicts now copies quests to the library; Closes #1884

### LINKED ISSUE #1884: Handle campaign export (copying quests)
When exporting a campaign individual quests can be exported already. To avoid overwriting the versions on the library make copies of the local versions then export those instead.

See here for more details https://github.com/bytedeck/bytedeck/issues/1849

thefoxwithin:
### What?
- Updated the export process to handle quests that already exist in the library by creating duplicate copies instead of overwriting or failing.
- Added `export_campaign_and_copy_quests` function to handle the export and copying of conflicting quests.
- Updated `export_campaign_to_library` to allow exports when all quests already exist in the library.
- Added `get_library_conflicting_quests` utility function to detect conflicts by `import_id`.
- Updated front-end warnings and tooltips to reflect the new behavior of creating duplicate copies.
- Implemented a safe naming mechanism for duplicate quests to avoid unique constraint violations.

### Why?
- Previously conflicting quests would be overwritten.
- With the new implementation of skipping conflicts, exporting a campaign would fail if all quests already existed in the library.
- Ensures that campaigns and their quests are always exportable, maintaining a copy in the library without overwriting existing content.
- Improves clarity for users through updated messages and icons in the UI.

https://github.com/bytedeck/bytedeck/issues/1884

### How?
- In `exporter.py`:
  - Added `export_campaign_and_copy_quests` function.
  - Modified `export_campaign_to_library` to skip conflicts and allow empty `quests` when `skip_import_ids` is not empty.
  - Used `deepcopy` and a counter-based suffix system to safely clone conflicting quests.
- In `utils.py`:
  - Added `get_library_conflicting_quests` to identify conflicting quests in the library by `import_id`.
- In `views.py`:
  - Updated the campaign export view to call `export_campaign_and_copy_quests` instead of `export_campaign_to_library`.
- In templates:
  - Updated warning message in campaign export UI.
  - Updated tooltip on the conflict icon to indicate quests will be copied.

### Testing?
- Verified exporting campaigns when some or all quests already exist in the library.
- Confirmed that duplicate quests are created with unique names and correct campaign associations.
- Checked that front-end warning messages and icons display correctly.
- Ensured the export still works for campaigns with no conflicting quests.
- Added test to make sure cloning functions correctly.
    - First export: mix of conflict + non-conflict
    - Second export: all quests conflict + second export of a conflicting quest

image

image

### Anything Else?

https://github.com/user-attachments/assets/8f1e6533-8fe5-4ac8-bd3b-70fffaa88ee2

### Review request
@tylerecouture

## Summary by CodeRabbit

- New Features
  - Campaign export is now conflict-aware: non-conflicting quests export immediately, while conflicting quests are copied as new unpublished versions linked to the campaign.
  - Ensures exported campaigns and cloned quests start unpublished.
- UI
  - Updated warning text to clarify that conflicts will be copied, not overwritten.
  - Revised tooltip to indicate conflicting quests will be copied during export.
- Tests
  - Added integration test validating cloning behavior for conflicting quests.
- Refactor
  - Centralized conflict detection logic and streamlined export flow.

### DIFF
--- a/src/library/exporter.py
+++ b/src/library/exporter.py
@@ -59,8 +59,11 @@ def export_quest_to_library(*, source_schema, quest_import_id):
 
 def export_campaign_to_library(*, source_schema, campaign_import_id, skip_import_ids=None):
     """
-    Exports a campaign (Category) and all its quests from the given tenant schema
-    into the shared library schema.
+    Exports a campaign (Category) and its published quests from the given tenant
+    schema into the shared library schema. By default, all published quests in
+    the campaign are exported. If `skip_import_ids` is provided, any quests whose
+    `import_id` is in that set are excluded (e.g., when you plan to clone
+    conflicts separately).
 
     All exported quests and the campaign will be unpublished in the library by default.
     This ensures the shared library remains a neutral, reviewable repository.
@@ -68,20 +71,25 @@ def export_campaign_to_library(*, source_schema, campaign_import_id, skip_import
     Args:
         source_schema (str): The tenant schema where the campaign currently resides.
         campaign_import_id (UUID): The import ID of the campaign to export.
+        skip_import_ids (Iterable[UUID] | None): Optional set of quest import_ids
+            to exclude from the export. When provided, the export will proceed
+            even if all quests are skipped
 
     Returns:
         ImportResult: The result of the import operation into the library schema.
 
     Raises:
         Category.DoesNotExist: If the campaign is not found in the source schema.
-        ValidationError: If no published quests exist for the campaign.
+        ValidationError: If no published quests exist for the campaign and
+            `skip_import_ids` was not provided.
     """
     with schema_context(source_schema):
         category = Category.objects.get(import_id=campaign_import_id)
         quests = Quest.objects.filter(published=True, campaign=category)
 
         # filter out conflicts (they'll be cloned and exported later)
-        quests = quests.exclude(import_id__in=skip_import_ids)
+        if skip_import_ids:
+            quests = quests.exclude(import_id__in=skip_import_ids)
 
         if not quests.exists() and not skip_import_ids:
             raise ValidationError("Cannot export a campaign without any published quests.")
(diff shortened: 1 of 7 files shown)


tylerecouture (commented):
Question (haven't looked at code yet)

When an exported quest already exists in the library, and this a duplicate is created (new uuid). Which version of the quest does the local version link to? I.e. does it get the new uuid too? Or does it maintain the old uuid (and thus a link to the original library version)

thefoxwithin:
> Question (haven't looked at code yet)
> 
> When an exported quest already exists in the library, and this a duplicate is created (new uuid). Which version of the quest does the local version link to? I.e. does it get the new uuid too? Or does it maintain the old uuid (and thus a link to the original library version)

It will continue to link to the first original uuid we don't update the local id yet.

src/library/exporter.py
```diff
@@ -1,11 +1,14 @@
 from django_tenants.utils import schema_context
 from quest_manager.admin import QuestResource
 from quest_manager.models import Quest, Category
-
-from .utils import library_schema_context
 from django.core.exceptions import ValidationError
 from django.db import IntegrityError
 
+from datetime import date
```
tylerecouture: nitpicking but these standard library imports should be at the very top.

standard lib imports
custom/pip imports, 
django imports 
project imports

src/library/templates/library/confirm_export_campaign.html
```diff
@@ -53,10 +53,9 @@
     </form>
     {% if library_quest_import_ids %}
         <div class="alert alert-danger panel-top">
-          {% comment %} TODO: Update this message to the new method of dealing with conflicts. {% endcomment %}
-            <strong>Warning:</strong> One or more quests in this campaign already exist in the library;
-            they are indicated with this icon <i class="icon-spacing fa fa-fw fa-exclamation-triangle"></i> in the quest list below.
-            Exporting this campaign will <strong>overwrite</strong> those existing quests with the local versions.
+          <strong>Warning:</strong> One or more quests in this campaign already exist in the library;
```
tylerecouture, line 56: I thought we moved all these warning/danger wells to the very top?

src/library/utils.py
```diff
@@ -24,3 +24,27 @@ def from_library_schema_first(request):
 
 
 library_schema_context = functools.partial(schema_context, get_library_schema_name())
+
+
+def get_library_conflicting_quests(local_quests):
+    """
+    Given a set of local quests, return any quests in the library
```
tylerecouture: Instead of saying "set" which is ambiguous, can be more specific?  is it a queryset? a list? any iterable?

src/library/utils.py
```diff
@@ -24,3 +24,27 @@ def from_library_schema_first(request):
...
+    that share the same import_id (i.e., conflicts).
+
+    Args:
+        local_quests (Iterable[Quest]): Local quests to check.
+
+    Returns:
+        list[Quest]: List of conflicting Quest objects from the library.
+    """
+    from quest_manager.models import Quest
+    quest_import_ids = [q.import_id for q in local_quests if q.import_id]
```
tylerecouture: why `if q.import_id`? isn't import_id a required field?

Also why are we list comprehension a queryset?  Why not just get a values_list of the import_ids?  iterating querysets is terribly inefficient.  I'm assuming we are providing a queryset here?

src/library/views.py
```diff
@@ -429,19 +427,11 @@ def get(self, request, campaign_import_id):
         campaign = get_object_or_404(Category, import_id=campaign_import_id)
         quests = list(campaign.current_quests())
```
tylerecouture, line 429: why listing/evaluating this queryset?  comment with context would be helpful

tylerecouture (changes requested):
A few comments, and this coderabbit is still open:

https://github.com/bytedeck/bytedeck/pull/1890#discussion_r2302183277

src/library/templates/library/confirm_export_campaign.html
thefoxwithin: I have it lower because if it was anywhere else it looked awkward, here's what it looks like currently.
image


Would you like me to play around with the positioning and give you screenshots of how they look?

tylerecouture: Nah, just leave it

src/library/views.py
thefoxwithin: Gotcha, added a code comment for it.

src/library/utils.py
thefoxwithin: You’re right that import_id is required on the model, so the if q.import_id check is redundant.

We’re currently always passing a list of Quest objects, not a queryset, which is why the list comprehension works. I tried switching it to accept a queryset and use .values_list(), but it consistently broke the page rendering.

Since the campaigns usually have a small number of quests, keeping it as a list is safe and ensures the warnings/UI continue to work correctly.

src/library/utils.py
thefoxwithin: Updated the docstring, we're always passing a list so it's "Given a list of local Quest objects..." now

tylerecouture (approved):

Issue #1730, opened 2025-01-25
New user form and response


tylerecouture:
Current form is a google form embedded in a Django Flatpage
https://bytedeck.com/pages/request-a-deck/

Process shoudl be something like this (but open to suggestions/improvements)
- [ ] This needs to be a django form on the public tenant, not sure where (which model) to store the data, since the new deck won't exist yet.
- [ ] Verify human (captcha or otherwise)
- [ ] Verify email address
- [ ] Then generate the new deck based on info provided
- [ ] Emails info to owner of new deck with username and temporary password (already happens when deck is manually created)
- [ ] Set deck to trial mode (max 5 users)
- [ ] Deck displays a Trial Mode banner with link to subscribe)

thefoxwithin:
Something like this as the template?

<img width="1896" height="966" alt="Image" src="https://github.com/user-attachments/assets/6a73a747-7ef7-43b4-96c6-bf50af734258" />

I made a new model in tenants called `DeckRequest` which stores
```python
    deck_name = models.CharField(max_length=255, unique=True)
    first_name = models.CharField(max_length=150)
    last_name = models.CharField(max_length=150)
    email = models.EmailField()
    verified = models.BooleanField(default=False)
    created_at = models.DateTimeField(auto_now_add=True)
```
Let me know if you think any other fields should be unique, I knew `deck_name` should be and I also thought that maybe email should be unique but I'm not sure.

I also made a new form in tenant called `DeckRequestForm` and then a `RequestNewDeck` view which renders the template above as  a form with a captcha.

The view will automatically send this Email

<img width="1045" height="529" alt="Image" src="https://github.com/user-attachments/assets/eb510856-c664-4ca5-a01c-b3ab6effd3da" />

It's all super simplistic as of right now. Let me know if there's any specific styling you want!

tylerecouture:
Looks great! Can you outline the entire workflow?  I can't see how duplicating the uniqueness/validators from the tenant form helps us.  But not sure how the whole thing will work.  Here's what I'm thinking:

1. Verify person (temporary user/deck request form?  I think the public tenant reaches into the deck/tenants to get the owner users, but that won't exist yet):
     1. Submit form with captcha
      1. Verify email address

1. Create trial deck
     1. Verification email will send a link to the deck creation form
     1. Use the existing tenant form, maybe with mods if needed?  preloaded with the verified email (can't change)
     2. Some checking like the link can't be used twice, for example.

[...]

PR #1892, opened 2025-08-30
new deck request model and form to automate process

thefoxwithin:
## What?
Refactored the deck request and tenant creation flow to use a centralized, stateless `DeckRequestService` for token generation, verification, and email handling.

- Added `DeckRequestService` class:

  - `generate_token(first_name, last_name, email)`: Creates a signed, timestamped token.
  - `decode_token(token)`: Validates and decodes token; returns user data if valid, `None` otherwise.
  - `build_verification_link(request, token)`: Builds a full verification URL.
  - `send_verification_email(first_name, email, token, request=None)`: Sends an email containing the verification token.
  - `send_welcome_email(user, tenant)`: Sends the welcome email to a newly created tenant owner.
  - `TOKEN_MAX_AGE = 3600` (1 hour expiration for tokens).

### Why?
The old deck request process relied on manual input and was less secure. Moving to a token-based system ensures that:

- Only verified requests can create new tenants.
- The email associated with the deck owner is guaranteed to be verified.
- Staff/admins can still bypass verification and created tenants normally.
- Session data doesn't persist past its intended use.

https://github.com/bytedeck/bytedeck/issues/1730

### How?
- Deck request submission now generates a signed, timestamped token with first name, last name, and email.
- The token is sent via email and must be used within 1 hour.
- When verified, the decoded payload is stored in `session["verified_deck_request"]`.
- `TenantCreateView` access is controlled via `EmailVerificationRequiredMixin`:
    - Staff can always access.
    - Non-staff must have a valid session token.
- Once the form is completed, session data is cleared.

### Testing?  
- Form validations, including email immutability when pre-verified  
- DeckRequestForm triggers verification email correctly  
- TenantCreate correctly saves owner and tenant, sets username/password, and marks email verified  
- Access restrictions enforced for public/non-public tenants and verified/non-verified users

image

Message displayed after submitting a valid deck request.
image

Clicking the verification link verifies the email and redirects to the create new deck form with pre filled in values and immutable email field.
image

Verification link only lasts an hour, if you wait longer than an hour this message gets displayed when you click the link after at least an hour has passed.
image

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Public "Request a New Deck" page with CAPTCHA, email verification link (1‑hour expiry) and welcome email containing initial login info; successful creation redirects to login and shows a "Deck created! Check your email for login credentials." alert.
  * "Deck request denied" page shown when verification is required.

* **Bug Fixes**
  * Fixed stray brace in messages alert.
  * Navbar/footer "Try it" links now use dynamic URL resolution.

* **Tests**
  * Tests updated for the new request/verification/onboarding flow.

### DIFF
--- a/src/templates/account/login.html
+++ b/src/templates/account/login.html
@@ -9,6 +9,12 @@
 {% block heading %}{% trans "Sign In" %}<i class="pull-right fa fa-sign-in"></i>{% endblock %}
 
 {% block content %}
+  {% if request.GET.created %}
+  <div class="alert alert-success" role="alert">
+    Deck created! Check your email for login credentials.
+  </div>
+  {% endif %}
+
   {% comment %}
     Conditions are setup this way because checking socialaccount_providers returns a list of strings based on the INSTALLED_APPS
     If you look at the get_providers tag, it returns all social account providers whether it is configured or not
--- a/src/tenant/forms.py
+++ b/src/tenant/forms.py
@@ -1,7 +1,11 @@
+from captcha.fields import ReCaptchaField
+from captcha.widgets import ReCaptchaV2Checkbox
+
 from django import forms
 from django.contrib.auth import get_user_model
 from django.forms import ModelForm
 
+from siteconfig.models import SiteConfig
 from .models import Tenant
 
 User = get_user_model()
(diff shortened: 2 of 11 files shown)


tylerecouture:
Only comment at this point is same one here: https://github.com/bytedeck/bytedeck/issues/1730#issuecomment-3224726612

thefoxwithin:
> Only comment at this point is same one here: [#1730 (comment)](https://github.com/bytedeck/bytedeck/issues/1730#issuecomment-3224726612)

Well if that's the case then you need not worry I did that already, I think I did that on Tuesday I must not've commented about it.

I made sure the tenant create will still work like before if you're a staff member. Also a staff member can also use the other method as well.

src/templates/account/login.html
```diff
@@ -9,6 +9,12 @@
 {% block heading %}{% trans "Sign In" %}<i class="pull-right fa fa-sign-in"></i>{% endblock %}
 
 {% block content %}
+  {% if request.GET.created %}
```
tylerecouture, line 12: Is there a reason we are adding the alert to the template instead of injecting it from the view like we do in the rest of the project?

src/tenant/forms.py
```diff
@@ -55,3 +59,91 @@ class TenantForm(TenantBaseForm):
...
+        if self.fields["email"].disabled:
+            email = (self.verified_data or {}).get("email") or self.initial.get("email")
+            if not email:
+                raise forms.ValidationError("Verified e-mail address is required.")
+            return email
+        return self.cleaned_data.get("email")
+
+    def save(self, commit=True):
+        """
+        Save the Tenant and ensure the deck owner User has the correct email and names.
```
tylerecouture, line 107: Why the commit parameter, in what workflow would we set it to false (and do we do that somewhere?)

src/tenant/templates/tenant/email/verify_deck_request.txt
```diff
@@ -0,0 +1,11 @@
+Hello {{ first_name }},
+
+Thank you for requesting a new deck.
+
+Please verify your email within 1 hour by clicking this link:
```
tylerecouture, line 5: What happens if they don't verify, what can they do about it?  Can they request again?  I've had issues in the past where something like this times out and then I have to manually go in to the django shell and clean things up so they can try again. Please confirm.

tylerecouture (changes requested):
Some basic comments.  That's all.  Nice work!

PR #1897, opened 2026-07-11
Django 5.2 prep: remove APIs dropped between 4.2 and 5.2, fix two broken unpinned dependencies

tylerecouture:
### What?
Phase 0 of the Django 4.2 → 5.2 LTS upgrade plan: remove every in-code usage of APIs that Django drops between 4.2 and 5.2, while staying fully compatible with the current Django 4.2. Also fixes two dependency breakages (unrelated to the upgrade) that currently prevent any fresh `pip install -r requirements.txt` from producing a working environment.

- Replace `STATICFILES_STORAGE` / `DEFAULT_FILE_STORAGE` (removed in Django 5.1) with the `STORAGES` dict in the `USE_S3` settings branch. The non-S3 branch keeps Django's defaults.
- Rename `assertQuerysetEqual` → `assertQuerySetEqual` (removed in Django 5.1) at all 31 call sites across 11 test files.
- Remove `USE_L10N = True` (setting removed in Django 5.0; it has been a no-op default since 4.0).
- Drop the `django-utils-six` shim and replace its 3 `text_type` usages in tests with plain `str`.
- Remove `namegenerator` from `requirements.txt` — **the package has been deleted from PyPI** (404 on the index), so every fresh install fails. Its single usage (naming fake quests/campaigns in `shell_utils.py`) is replaced with a small local generator.
- Pin `redis>=4.0.2,<8` — redis-py is an unpinned transitive dep (django-redis, celery), and version 8 negotiates RESP3 via a `HELLO` command the `redis:5.0-alpine` server in docker-compose doesn't understand, producing ~25 test errors on any fresh install.

### Why?
Django 4.2's extended support has ended, so the app is on an unpatched framework version. The upgrade plan stages all risk while still on 4.2: this PR clears the removed-API surface so later dependency bumps and the eventual `Django>=5.2,<5.3` pin flip stay small and individually green. The two dependency fixes are included because without them CI cannot build a fresh image at all — the breakage exists on `develop` today.

### How?
All changes are mechanical and backward-compatible on Django 4.2: `STORAGES` and the new assert casings have existed since 4.2, `USE_L10N=True` matches the hardcoded default, and the `str`/`text_type` swap is behavior-identical on Python 3.

### Testing?
Verified in a Python 3.10 container against the docker-compose Postgres 16 + Redis 5.0 (mirroring CI):
- `python src/manage.py check` — no issues
- `python src/manage.py makemigrations --check --dry-run` — no migration drift
- `flake8` clean on all changed files
- `manage.py test` across all nine affected apps (utilities, tags, prerequisites, comments, badges, djcytoscape, courses, profile_manager, quest_manager): **675 tests, OK**
- The redis pin was confirmed empirically: the same test errors reproduce with redis-py 8.0.1 in files this PR doesn't touch, and disappear with redis-py 7.4.1.

No front-end changes.

### Anything Else?
This is Phase 0 of a staged upgrade plan. Next phases (separate PRs): dependency cluster bumps on 4.2 (django-tenants 3.10, allauth 65.x, import-export 4.x, crispy-forms 2.x + crispy-bootstrap3, django-recaptcha 4.x, taggit/redis/celery-beat), then a deprecation-warnings-as-errors pass, then the `Django>=5.2,<5.3` flip. The Redis 5.0 server itself (EOL) should be upgraded eventually, which would let the `redis<8` pin be lifted.

### Review request
@tylerecouture

https://claude.ai/code/session_014RwTtNLXvwNkobVYP9kABW

---

## Summary by CodeRabbit

- **New Features**
  - Added Redis support for production deployments.
  - Updated cloud storage configuration for compatibility with newer Django versions.
  - Added built-in random naming for generated campaigns and quests.

- **Bug Fixes**
  - Corrected queryset comparison assertions across the test suite.
  - Updated text type checks for modern Python compatibility.
  - Replaced obsolete naming functionality and removed outdated dependencies.

### DIFF
--- a/requirements-production.txt
+++ b/requirements-production.txt
@@ -33,4 +33,5 @@ pytz # required by django-redis which uses pickle, even though django-redis chan
 
 # subdependencies
 psycopg2-binary>=2.8.6,<2.10  # since our backend is PostgreSQL, Django and django-tenants use this to interact with the db
+redis>=4.0.2,<8  # pulled in via django-redis/celery; redis-py 8+ speaks RESP3 (HELLO) which the redis 5.0 server in docker-compose doesn't support
 pillow>=10.3,<11.0  # used implicitly by django_resized (does not exist as an extra package either)
--- a/requirements-production.txt
+++ b/requirements-production.txt
@@ -19,7 +19,6 @@ django-summernote>=0.8,<0.9
 django-taggit>=2.1,<5.0
 django-tenants>=3.5,<3.6  # latest version that supports python 3.8 (but now using 3.10 so can update it?)
 django-url-or-relative-url-field>=0.2.0,<0.3.0
-django-utils-six==2.0  # only 2.0 exists. Previous version was removed back in Django 3.0
 tenant-schemas-celery>=2.2,<3.1
 
 # non django
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,5 +12,4 @@ pre-commit
 freezegun>=0.3.12
 mock>=2.0,<6.0
 model_bakery>=1.1,<2.0
-namegenerator
 names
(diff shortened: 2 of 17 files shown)


tylerecouture (commented):
Good.

PR #1902, opened 2026-07-12
Batch 3: eliminate per-badge N+1 queries on profile detail and notifications

tylerecouture:
### What?
Third batch of query-efficiency fixes, targeting the profile detail page (the worst remaining offender) and the notifications list. Measured by profiling the seeded tenant (100 students, 800 submissions, 300 badge assertions) with per-view SQL capture:

| View | Real queries before | Real queries after |
|---|---|---|
| Student profile detail (own) | 65 | **45** |
| Student profile detail (staff view) | 65 | **41** |
| Notifications list | 21 | **17** |

The changes:
1. **Profile badge popovers ran a query per badge for duplicate assertions.** The profile page renders one popover per earned badge, and each called `assertion.get_duplicate_assertions()` → one query per badge (x15 on the profiled student). `badge_assertions_dict_items` now fetches all of the user's assertions once, groups them by badge in Python, and populates a per-assertion cache that `get_duplicate_assertions()` reads.
2. **Each badge popover also ran a query for its prerequisites** (`badge.prereqs`, x10). Added `PrereqManager.prefetch_for_parents()`, which loads the prerequisites of many same-model objects in a single query and populates a `_prereqs_cache` that `HasPrereqsMixin.prereqs()` reads; `badge_assertions_dict_items` calls it for the displayed badges. Non-prefetched callers are unchanged — `prereqs()` still returns a normal queryset, so existing `.count()`/`.exists()`/`.delete()` call sites are unaffected.
3. **Notifications rendered each row's generic-FK objects one query at a time.** The list view now `prefetch_related`s `sender_object`, `target_object`, and `action_object`, collapsing the per-row sender/target/action lookups into a few grouped queries.
4. **Dead work removed on the profile page:** `badge_assertions_by_type` was computed for the context but never rendered in any template — its only real effect was the badge-granting side effect of `check_for_new_assertions()`, which is now called directly. The profile view also called `get_user_tags_and_xp()` twice (the first result was immediately overwritten); the redundant call is removed.

### Why?
The profile detail page is the most query-heavy student-facing page and scales with the number of badges a student has earned (a popover per badge, each doing 2+ queries). The notifications list scaled with the number of notifications per page. Both are frequently visited.

### How?
Profiled with Django's `CaptureQueriesContext` against the generated tenant, ranked the remaining repeated query shapes, and eliminated the per-row ones with a Python-side grouping (duplicate assertions), a batched generic-FK prefetch helper (badge prereqs), and standard `prefetch_related` (notification GFKs). The remaining profile-page prereq queries come from the preserved `check_for_new_assertions()` badge-granting path, which is correctness-critical and out of scope here.

### Testing?
- New regression tests, each asserting the batched path issues **no** per-object query (`assertNumQueries(0)`):
  - `badges`: `test_badge_assertions_dict_items_prefetches_duplicates`
  - `prerequisites`: `PrereqPrefetchTest` (prefetch serves `prereqs()` with no per-object query; unprefetched objects still return a queryset; empty input is a no-op)
- Full `badges`, `prerequisites`, `profile_manager`, and `notifications` suites pass.
- flake8 clean.
- Before/after query counts re-measured with cold caches (table above).

No template or rendered-output changes — the profile page and notifications render identically (verified by the existing view/template tests).

### Anything Else?
Remaining smaller offenders noted for a possible follow-up: the per-badge tag lists on the profile popovers (a few `taggit` queries) and the notifications list's `get_root_url()` (a tenant + domain query per notification), which would benefit from a schema-scoped cache like the ranks/rarities caches — deferred here because it touches multi-tenancy internals and the win is modest. The prerequisites recalculation engine remains the big architectural item.

### Review request
@tylerecouture

https://claude.ai/code/session_014RwTtNLXvwNkobVYP9kABW

---

## Summary by CodeRabbit

* **Performance Improvements**
  * Faster badge and prerequisite loading by reducing repeated database requests.
  * Improved notification list rendering, especially for pages with many notifications.
  * Badge duplicate information is now loaded efficiently for smoother profile and badge views.

* **Bug Fixes**
  * Profile pages continue granting newly earned badges while displaying the expected badge and submission information.

* **Tests**
  * Added coverage confirming correct badge duplicates, prerequisite results, and notification-related query efficiency.

### DIFF
--- a/src/badges/models.py
+++ b/src/badges/models.py
@@ -357,9 +357,22 @@ class BadgeAssertionManager(models.Manager):
         return sorted_qs
 
     def badge_assertions_dict_items(self, user):
-        earned_assertions = self.all_for_user_distinct(user)
+        earned_assertions = list(self.all_for_user_distinct(user))
+
+        # Group all of the user's assertions by badge in one query, so each
+        # badge's get_duplicate_assertions() (rendered in every popover) is
+        # served from memory instead of a query per badge.
+        duplicates_by_badge = defaultdict(list)
+        for assertion in self.get_queryset(False).get_user(user):
+            duplicates_by_badge[assertion.badge_id].append(assertion)
+
+        # Prefetch every displayed badge's prerequisites in one query (the
+        # popover lists them) instead of a query per badge.
+        Prereq.objects.prefetch_for_parents([a.badge for a in earned_assertions])
+
         assertion_dict = defaultdict(list)
         for assertion in earned_assertions:
+            assertion._duplicate_assertions_cache = duplicates_by_badge[assertion.badge_id]
             assertion_dict[assertion.badge.badge_type].append(assertion)
 
         return assertion_dict.items()
@@ -475,7 +488,12 @@ class BadgeAssertion(models.Model):
         return count
 
     def get_duplicate_assertions(self):
-        """A qs of all assertions of this badge for this user"""
+        """A qs (or a pre-populated list) of all assertions of this badge for this user.
+
+        badge_assertions_dict_items() can populate _duplicate_assertions_cache so
+        the profile page serves this from memory instead of a query per badge."""
+        if hasattr(self, '_duplicate_assertions_cache'):
+            return self._duplicate_assertions_cache
         return BadgeAssertion.objects.all_for_user_badge(self.user, self.badge, False)
 
 
--- a/src/badges/models.py
+++ b/src/badges/models.py
@@ -357,6 +357,17 @@ class BadgeAssertionManager(models.Manager):
         return sorted_qs
 
     def badge_assertions_dict_items(self, user):
+        """Group a user's earned badges by badge type for the profile page.
+
+        Returns one assertion per earned badge (via all_for_user_distinct),
+        grouped into dict items keyed by BadgeType. Each returned assertion
+        has its duplicate assertions and its badge's prerequisites
+        pre-populated in one query each, so the per-badge popovers render
+        without issuing a query per badge.
+
+        :param user: the user whose earned badges are grouped
+        :return: dict_items of {BadgeType: [BadgeAssertion, ...]}
+        """
         earned_assertions = list(self.all_for_user_distinct(user))
 
         # Group all of the user's assertions by badge in one query, so each
--- a/src/prerequisites/models.py
+++ b/src/prerequisites/models.py
@@ -321,6 +321,12 @@ class PrereqManager(models.Manager):
         if not parent_objects:
             return parent_objects
 
+        # all parents share one content type; a mixed list would silently drop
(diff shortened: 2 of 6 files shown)


src/badges/models.py
```diff
@@ -357,9 +357,33 @@ def all_for_user_distinct(self, user):
...
+        for assertion in self.get_queryset(False).get_user(user):
+            duplicates_by_badge[assertion.badge_id].append(assertion)
+
+        # Prefetch every displayed badge's prerequisites in one query (the
+        # popover lists them) instead of a query per badge.
+        Prereq.objects.prefetch_for_parents([a.badge for a in earned_assertions])
+
         assertion_dict = defaultdict(list)
         for assertion in earned_assertions:
+            assertion._duplicate_assertions_cache = duplicates_by_badge[assertion.badge_id]
```
tylerecouture, line 386: Add code comment explaining the purpose of this line.  Is this an ephemeral cache being created?  When is the cache released?

src/badges/models.py
```diff
@@ -475,7 +499,12 @@ def count_bootstrap_badge(self):
         return count
 
     def get_duplicate_assertions(self):
-        """A qs of all assertions of this badge for this user"""
+        """A qs (or a pre-populated list) of all assertions of this badge for this user.
+
+        badge_assertions_dict_items() can populate _duplicate_assertions_cache so
+        the profile page serves this from memory instead of a query per badge."""
+        if hasattr(self, '_duplicate_assertions_cache'):
```
tylerecouture, line 506: Why would have/not have this attribute?  Was it create ephemerally, not as part of the model? please explain

src/prerequisites/models.py
```diff
@@ -299,6 +307,39 @@ def get_queryset(self):
...
+        All parents must be instances of the same model (they share one
+        content type). Each object gets a ``_prereqs_cache`` list that
+        HasPrereqsMixin.prereqs() returns, avoiding a query per object.
+
+        :param parent_objects: an iterable of same-model objects to populate
+        :return: the parent objects as a list
+        """
+        parent_objects = list(parent_objects)
+        if not parent_objects:
+            return parent_objects
```
tylerecouture, line 322: Why?  Is this just an empty list?  Why not be explicit and return an empty list?

src/prerequisites/models.py
```diff
@@ -299,6 +307,39 @@ def get_queryset(self):
     def all_parent(self, parent_object):
         return self.get_queryset().get_all_for_parent_object(parent_object)
 
+    def prefetch_for_parents(self, parent_objects):
+        """Populate each parent object's prereqs() cache in a single query.
+
+        All parents must be instances of the same model (they share one
```
tylerecouture, line 313: What does this method do if they ARENT the same parent?  Please document.

PR #1923, opened 2026-07-13
Treat max_repeats=-1 as unlimited for repeat-per-semester quests

### LINKED ISSUE #1531: 404 Error For Returning Students when Max Repeats = -1 and Repeat per semester enabled
[This quest ](https://studiotyee.bytedeck.com/quests/1294/) was causing a 404 error for students in their 3rd repeat of the course. It was mistakenly set up to have a repeat per semester and also had max repeats set at -1. After I changed Max repeats to 0, I did not have the same issue for those students. 

Perhaps the 404 error was due to conflicting values. Would it be better to have one override the other?

tylerecouture:
### What?
Fixes `Quest.is_repeat_available()` so that `max_repeats = -1` (unlimited) works when `repeat_per_semester` is also enabled.

Closes #1531

### Why?
The issue reported 404s for students on their 3rd repeat of a course when a quest had "repeat per semester" enabled together with max repeats of -1. Root cause in `is_repeat_available()`: the per-semester branch checks `qs.count() > self.max_repeats`, and any count is greater than -1, so the quest was **never** repeat-available in that configuration — students who'd completed it in a previous semester couldn't access it again, producing the 404. The non-semester branch already special-cased `-1`; the per-semester branch didn't.

### How?
One-line change in `src/quest_manager/models.py`: the per-semester repeat check now skips the limit when `max_repeats == -1`, matching the other branch's semantics. (This answers the issue's question about which value should win: `-1` consistently means "unlimited repeats" in both modes, so no form validation conflict is needed.)

### Testing?
New test `QuestTestModel.test_is_repeat_available__infinite_repeats_per_semester` — fails before the fix (`is_repeat_available()` returns False after completion) and passes after. Full `quest_manager` suite passes (221 tests); flake8 clean.

N/A — model logic fix.

### Anything Else?
Part of a batch of bug fixes working through the open `bug`-labelled issues.

### Review request
@tylerecouture

---

### DIFF
--- a/src/quest_manager/models.py
+++ b/src/quest_manager/models.py
@@ -674,11 +674,11 @@ class Quest(IsAPrereqMixin, HasPrereqsMixin, TagsModelMixin, XPItem):
         if not time_of_last:
             return False
 
-        # if haven't maxed out repeats
+        # if haven't maxed out repeats (max_repeats = -1 means unlimited repeats)
         if self.repeat_per_semester:
             # get all completed this semester
             qs = QuestSubmission.objects.all_completed(user=user).filter(quest=self)
-            if qs.count() > self.max_repeats:
+            if qs.count() > self.max_repeats and self.max_repeats != -1:
                 return False
         elif latest_sub.ordinal > self.max_repeats and self.max_repeats != -1:
             return False
--- a/src/quest_manager/tests/test_models.py
+++ b/src/quest_manager/tests/test_models.py
@@ -370,6 +370,31 @@ class QuestTestModel(TenantTestCase):
         # No repeat left this semester
         self.assertFalse(quest_semester.is_repeat_available(student))
 
+    @freeze_time('2018-10-12 00:54:00', tz_offset=0)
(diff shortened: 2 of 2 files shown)


tylerecouture (commented):
These two settings together don't make sense.  Instead, there should be an error or warning in the quest form.  If a quest is `repeat_per_semester` but `max_repeats` is -1 (infinite), then the "repeatable_per_semester" is doing nothing.  That's probably a user error?  @claude think about this more thoroughly, remove this PR and attempt to resolve the issue anew.  Don't allow a user to set -1 on max_repeats and repeat_per_semester=True at the same time.  Provide useful feedback to the user explaining why.

tylerecouture:
Agreed — closing this in favor of form validation. Replaced by #1927, which rejects `max_repeats = -1` + `repeat_per_semester` in `QuestForm.clean()` with a message explaining the conflict and how to resolve it (set Max Repeats to a per-semester count, or turn off Repeat per semester to keep unlimited repeats).

---

Issue #1286, opened 2023-01-16
Campaign prereq doesn't ignore draft quests


tylerecouture:
Needs a test to catch the bug and a fix

To replicate in a new test deck:
- Change one of the "Orientation" quests to a draft (published = false)
- With a test student, complete the remaining orientation quests, and notice that the "Bytedeck Proficiency" badge is not granted.
- Delete the draft quest, and the badge STILL isn't granted.

Tests / bugfixes needed:

- [ ] That `Category.condition_met_as_prerequisite()` ignores quests where `Quest.published==False`
- [ ] That `Category.condition_met_as_prerequisite()` ignores quests where `Quest.archived==True`
- [ ] That `Category.condition_met_as_prerequisite()` ignores quests where `Quest.active()==True` (this checks opening and expiry dates)
- [ ] That when a Quest is removed from a category, students with the remaining quests are granted anything resulting from the remaining quests (does this need a global recalc for all current students or is there a more efficient way?)

tylerecouture:
3 of 4 completed.  @9921622 investigate the last item on the list.

PR #1925, opened 2026-07-13
Grant badges to qualifying students via a teacher-confirmed flow

### LINKED ISSUE #1157: New prereq doesn't trigger expected celery tasks
When a prereq is added to a badge or quest, celery should recalculate whether that badge should be assigned, or that quest should become available to users, but doesn't trigger the task:
Peek 2022-08-12 10-04

tylerecouture:
### What?
Lets a teacher grant a badge, on demand, to all current students who meet its prerequisites, and notifies those students' teachers of each grant.

Closes #1157

### Why?
Badges are only auto-granted lazily — `BadgeAssertion.objects.check_for_new_assertions()` runs when a student's profile page is viewed or when they complete a quest. So when a teacher adds a prereq to a badge, students who already meet the new condition get nothing until they happen to load their profile.

We deliberately do **not** grant automatically on every prereq change: a teacher building a badge may tweak its prerequisites several times (add one, then another), and auto-granting in between could grant the badge before it's ready. Instead the teacher triggers a grant-check when they're done, and confirms the count first.

### How?
- **`prerequisites/tasks.py`**: `grant_badge_assertions_for_badge(badge_id, start_from_user_id)` — grants the badge to every current-semester student who has no assertion and meets its conditions (`no_prereq_means=False`, so no-prereq badges stay manual-only), processing users in `CELERY_TASKS_BUNCH_SIZE` bunches recursively and notifying each student's current teachers. Bails if the badge was deleted/unpublished meanwhile; a short debounce guards against a double-clicked button.
- **`badges/views.py`**: `badge_grant_qualifying` view — GET shows how many current students qualify but don't have the badge yet and asks the teacher to confirm; POST queues the task and redirects. Published badges only.
- **Two entry points** (per review): a persistent "grant to qualifying students" button on the badge detail page, and a prompt (a message linking to the confirmation page) shown after a teacher saves the badge's prerequisites (`BadgePrereqsUpdate.form_valid`).
- **`badges/models.py`**: `Badge.students_who_qualify_ungranted()` powers the confirmation preview.
- **`prerequisites/signals.py`**: the `Prereq` post_save/post_delete receiver no longer queues the grant task for badge parents (quest behaviour unchanged).

### Testing?
- `test_tasks.py::GrantBadgeAssertionsForBadgeTest` — grants + teacher notification; skips users not meeting conditions; skips no-prereq badges; no duplicate assertions; unpublished badges never granted; deleted-badge and already-started guards; bunching.
- `test_signals.py::test_badge_prereq_changes_do_not_auto_grant` — badge prereq create/update/delete must NOT queue the grant task.
- `badges/tests/test_views.py` — grant-qualifying page lists qualifying students (GET), POST queues the task and redirects, unpublished-badge guard, and the post-prereq-save prompt is shown for published badges and hidden for unpublished ones.

Full `prerequisites` + `badges` suites pass; flake8 clean; no new migrations.

New: a "grant to qualifying students" button on the badge detail page, and a confirmation page that lists the qualifying students before granting.

### Anything Else?
The task still covers the campaign→badge case noted previously for #1286, but only when a badge's own prereq rows change; campaign-membership changes still need a separate signal and are left for #1286.

### Review request
@tylerecouture

---

## Summary by CodeRabbit

* **New Features**
  * Added a staff-only “grant qualifying” review and confirmation flow for published badges.
  * Added an endpoint and confirmation page to list eligible students and grant badge assertions in bulk.
  * Introduced an async background task to grant qualifying badges and notify affected students’ teachers.
  * After saving prerequisites on a published badge, staff are prompted to review eligible students.

* **Bug Fixes**
  * Badge grant/review actions are hidden and blocked for unpublished badges.
  * Prevented duplicate badge grants and ensured prerequisites changes do not auto-grant badges.

### DIFF
--- a/src/badges/models.py
+++ b/src/badges/models.py
@@ -239,24 +239,36 @@ class Badge(IsAPrereqMixin, HasPrereqsMixin, TagsModelMixin, models.Model):
     def get_absolute_url(self):
         return reverse('badges:badge_detail', kwargs={'badge_id': self.id})
 
+    def student_qualifies_ungranted(self, user):
+        """Return True if `user` meets this badge's prerequisites but hasn't been granted it
+        yet; otherwise False.
+
+        Badges with no prerequisites are manual-grant-only, hence `no_prereq_means=False`
+        (matching BadgeManager.get_conditions_met). Shared by the grant-check preview
+        (students_who_qualify_ungranted) and the grant task, so both apply the same rule.
+        """
+        from prerequisites.models import Prereq
+
+        return (
+            not BadgeAssertion.objects.all_for_user_badge(user, self, False).exists()
+            and Prereq.objects.all_conditions_met(self, user, no_prereq_means=False)
+        )
+
     def students_who_qualify_ungranted(self):
-        """Return current-semester students who meet this badge's prerequisites but have
-        not been granted it yet.
+        """Return the current-semester students who meet this badge's prerequisites but have
+        not been granted it yet, as a list of User objects (empty when none qualify).
 
         Used to preview and confirm a teacher-initiated grant-check (issue #1157): rather
         than auto-granting on every prereq edit (which risks granting a badge before it is
-        finished), the teacher runs this check on demand. Badges with no prerequisites are
-        manual-grant-only, hence `no_prereq_means=False` (matching BadgeManager.get_conditions_met).
+        finished), the teacher runs this check on demand.
+
+        Returns:
+            list[User]: e.g. ``[<User: ada>, <User: grace>]``
         """
         from courses.models import CourseStudent
-        from prerequisites.models import Prereq
 
         students = CourseStudent.objects.all_users_for_active_semester()
-        return [
-            user for user in students
-            if not BadgeAssertion.objects.all_for_user_badge(user, self, False).exists()
-            and Prereq.objects.all_conditions_met(self, user, no_prereq_means=False)
-        ]
+        return [user for user in students if self.student_qualifies_ungranted(user)]
 
     def get_icon_url(self):
         if self.icon and hasattr(self.icon, 'url'):
--- a/src/badges/models.py
+++ b/src/badges/models.py
@@ -239,6 +239,25 @@ class Badge(IsAPrereqMixin, HasPrereqsMixin, TagsModelMixin, models.Model):
     def get_absolute_url(self):
         return reverse('badges:badge_detail', kwargs={'badge_id': self.id})
 
+    def students_who_qualify_ungranted(self):
+        """Return current-semester students who meet this badge's prerequisites but have
+        not been granted it yet.
+
+        Used to preview and confirm a teacher-initiated grant-check (issue #1157): rather
+        than auto-granting on every prereq edit (which risks granting a badge before it is
+        finished), the teacher runs this check on demand. Badges with no prerequisites are
+        manual-grant-only, hence `no_prereq_means=False` (matching BadgeManager.get_conditions_met).
+        """
+        from courses.models import CourseStudent
+        from prerequisites.models import Prereq
+
+        students = CourseStudent.objects.all_users_for_active_semester()
+        return [
+            user for user in students
+            if not BadgeAssertion.objects.all_for_user_badge(user, self, False).exists()
+            and Prereq.objects.all_conditions_met(self, user, no_prereq_means=False)
+        ]
+
     def get_icon_url(self):
         if self.icon and hasattr(self.icon, 'url'):
             return self.icon.url
--- a/src/badges/templates/badges/detail.html
+++ b/src/badges/templates/badges/detail.html
@@ -43,7 +43,7 @@
             {% if badge.published %}
             <a class="btn btn-success" title="Grant to all students who meet this {{ config.custom_name_for_badge|lower }}'s prerequisites"
                href="{% url 'badges:grant_qualifying' badge.id %}">
-              <i class="fa fa-check-circle-o"></i>
+              <i class="fa fa-fw fa-users"></i>
             </a>
             {% endif %}
           </div>
--- a/src/badges/templates/badges/detail.html
+++ b/src/badges/templates/badges/detail.html
@@ -40,6 +40,12 @@
             <a class="btn btn-success" title="Bulk grant to multiple student" href="{% url 'badges:bulk_grant_badge' badge.id %}">
               <i class="fa fa-users"></i>
             </a>
+            {% if badge.published %}
+            <a class="btn btn-success" title="Grant to all students who meet this {{ config.custom_name_for_badge|lower }}'s prerequisites"
+               href="{% url 'badges:grant_qualifying' badge.id %}">
+              <i class="fa fa-check-circle-o"></i>
+            </a>
+            {% endif %}
           </div>
         </div>
         {% endif %}
--- a/src/badges/views.py
+++ b/src/badges/views.py
@@ -7,8 +7,9 @@ from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.shortcuts import get_object_or_404, redirect, render
 from django.http import JsonResponse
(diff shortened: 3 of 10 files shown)


tylerecouture (commented):
We should think about this more.  When a teacher is creating a batch, they may tweak the prerequisites a few times, or add one first, then add another.  We don't want to accidentally grant the badge before it is ready.  Perhaps instead, after a badge's prerequisites are change, we prompt the editor to run a check and grant badges to any qualifying students.

tylerecouture:
Good call — reworked to be teacher-initiated instead of automatic (pushed in d2fa228).

- Prereq changes no longer grant anything, so tweaking a badge's prereqs mid-build can't grant it early.
- The teacher runs the grant-check on demand and **confirms the count first**: the page shows how many current students meet the prereqs but don't have the badge yet, then grants on confirm.
- Two entry points, as discussed: a persistent "grant to qualifying students" button on the badge detail page, plus a prompt (a message linking to that confirmation page) shown right after they save the badge's prerequisites.

The grant task itself is unchanged (still grants in bunches and notifies each student's teachers); only the trigger moved from a signal to the confirmation view. PR description updated to match.

- Claude Code

---

src/badges/templates/badges/detail.html
```diff
@@ -40,6 +40,12 @@
             <a class="btn btn-success" title="Bulk grant to multiple student" href="{% url 'badges:bulk_grant_badge' badge.id %}">
               <i class="fa fa-users"></i>
             </a>
+            {% if badge.published %}
+            <a class="btn btn-success" title="Grant to all students who meet this {{ config.custom_name_for_badge|lower }}'s prerequisites"
+               href="{% url 'badges:grant_qualifying' badge.id %}">
+              <i class="fa fa-check-circle-o"></i>
```
tylerecouture: Check the badge buttons elsewhere in the app, what icon do we use for granting assertions of a badge?

src/badges/models.py
```diff
@@ -239,6 +239,25 @@ def __str__(self):
...
         return reverse('badges:badge_detail', kwargs={'badge_id': self.id})
 
+    def students_who_qualify_ungranted(self):
+        """Return current-semester students who meet this badge's prerequisites but have
+        not been granted it yet.
+
+        Used to preview and confirm a teacher-initiated grant-check (issue #1157): rather
+        than auto-granting on every prereq edit (which risks granting a badge before it is
+        finished), the teacher runs this check on demand. Badges with no prerequisites are
+        manual-grant-only, hence `no_prereq_means=False` (matching BadgeManager.get_conditions_met).
```
tylerecouture: clarify what data type is being returned by this method and provide an example in the docstring

src/badges/views.py
```diff
@@ -67,6 +70,61 @@ def badge_list(request):
...
+                    f'Prerequisites changed. '
+                    f'<a href="{grant_url}">Check and grant this {badge_name} to qualifying students?</a>'
+                )
+            )
+        return response
+
+
+@non_public_only_view
+@staff_member_required
+def badge_grant_qualifying(request, badge_id):
```
tylerecouture: Can we make this a CBV?  New views should all be CBV where possible, unless there is a compelling reason why to use a FBV in this case.

src/prerequisites/tasks.py
```diff
@@ -108,6 +108,74 @@ def update_conditions_for_quest(self, quest_id, start_from_user_id):
...
+    cache_key = f'grant_badge_assertions_for_badge_{badge_id}_wait'
+    if start_from_user_id == 1 and cache.get(cache_key):
+        return f"Skipping task for badge {badge_id}, already started."
+    cache.set(cache_key, True, 1)
+
+    users = CourseStudent.objects.all_users_for_active_semester()
+    users = users.order_by('id').filter(id__gte=start_from_user_id)[:settings.CELERY_TASKS_BUNCH_SIZE]
+
+    user = None
+    for user in users:
```
tylerecouture, line 144: How much of this is a duplicate of `Badge.students_who_qualify_ungranted`?  Is there a reason not to use that method instead of duplicating code here?

tylerecouture:
Thanks — all four addressed in 2c17a92:

- **CBV** (`views.py:95`): `badge_grant_qualifying` is now a class-based view, `BadgeGrantQualifying(NonPublicOnlyViewMixin, View)`, with `get`/`post`.
- **Duplication** (`tasks.py:144`): you were right, the grant condition duplicated the preview. Extracted `Badge.student_qualifies_ungranted(user)` as the single source of the "meets prereqs but not yet granted" rule; both the task and `students_who_qualify_ungranted()` now call it.
- **Docstring / return type** (`models.py:249`): `students_who_qualify_ungranted()` now documents that it returns `list[User]` (empty when none qualify) with an example, and the new per-user helper is documented too.
- **Icon** (`detail.html:46`): switched from `fa-check-circle-o` to `fa-users` to match the grant convention (`badges:grant` uses `fa-user`, `bulk_grant_badge` uses `fa-users`). One heads-up: it now looks identical to the manual bulk-grant button beside it (both green `fa-users`) — distinguished only by tooltip. Happy to add a modifier (e.g. a stacked check) if you'd prefer them visually distinct.

- Claude Code

---

tylerecouture:
Addressed CodeRabbit's review in a7f8b0c:

- **XSS / `format_html`** (`views.py`) — switched the prereq-change prompt from `mark_safe(f"…")` to `format_html(...)` so interpolated values are escaped.
- **`full_clean()` before `save()`** — added in the two `test_views.py` unpublished-badge tests, `test_signals.py`, and `test_tasks.py`, per the repo path instruction.
- **Explicit GFK target** (`test_signals.py`) — the baker-made `Prereq` now gets a `prereq_object`.
- **`test_tasks.py` `setUp`** — added a docstring and `TenantClient`, matching the suite convention.

Two pre-merge check notes:
- **Title check** — updated the PR title to reflect the teacher-confirmed flow.
- **Linked Issues check** — CodeRabbit suggests restoring the automatic prereq-change trigger. That's intentionally **not** applied: per @tylerecouture's review, auto-granting on every prereq edit could grant a badge mid-build before it's ready, which is exactly why this was reworked to a teacher-confirmed grant. The task still satisfies #1157 (grants qualifying students + notifies teachers); only its trigger is now explicit rather than automatic.

- Claude Code

---

src/badges/views.py
```diff
@@ -67,6 +70,74 @@ def badge_list(request):
...
+        if self.object.published:
+            badge_name = SiteConfig.get().custom_name_for_badge.lower()
+            grant_url = reverse('badges:grant_qualifying', args=[self.object.id])
+            messages.info(
+                self.request,
+                mark_safe(
+                    f'Prerequisites changed. '
+                    f'<a href="{grant_url}">Check and grant this {badge_name} to qualifying students?</a>'
+                )
+            )
```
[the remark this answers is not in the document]
tylerecouture: @claude explain why this is unsafe and we are changing to "format_html" which we don't use elsewhere. Let's be consistent and we can consider a project-wode update if we think the code introduces XSS vulnerabilities. I don't think it does here, because it's only showing our code, and teacher's info, correct? Nothing a student or random user could inject?

tylerecouture:
To answer the question on `views.py`: no real XSS here — the two interpolated values are a reversed URL and the staff-set badge label (`SiteConfig.custom_name_for_badge`), and the message only renders on the staff-only prereqs form, so there's no student/anonymous injection path. Ruff's S308 just flags every `mark_safe`-on-f-string categorically.

Since `format_html` isn't our convention (elsewhere we pass plain strings to `messages.*` and rely on `messages-snippet.html`'s `{{ message|safe }}`), I've switched it to a plain f-string in 209117f — consistent, and a plain f-string doesn't trip S308 either.

Opened #1961 to standardize message-HTML escaping project-wide (there are a couple of spots that interpolate usernames unescaped — a bigger surface than this one), so we can handle that deliberately rather than one-off here.

- Claude Code

---

PR #1950, opened 2026-07-15
nginx: template the domain so production and staging share one config

tylerecouture:
### What?
Replaces the static `nginx/bytedeck.aws.conf` with a single templated config, `nginx/bytedeck.conf.template`, that both production and staging render from. The domain (`server_name`, the `Host`-check regex, and the Let's Encrypt cert paths) is now a `${DOMAIN}` placeholder instead of a hardcoded `bytedeck.com`.

### Why?
Staging (`bytedeck-staging.com`) was maintaining a **hand-edited copy** of the nginx config whose only difference from production was the domain — in `server_name` (two blocks), the `Host`-validation regex, and the three `/etc/letsencrypt/live/<domain>/…` cert paths. That's config drift waiting to bite: any real nginx change had to be applied to both copies by hand. This makes the repo the single source of truth for both environments.

### How?
- **`nginx/bytedeck.conf.template`** — the former config with the five domain occurrences replaced by `${DOMAIN}`.
- **`nginx/Dockerfile`** — renders the template at image build time with `envsubst '${DOMAIN}'`. The shell-format is restricted to `${DOMAIN}`, so nginx's own `$host`/`$scheme`/`$request_uri`/`$1` variables are left untouched (`envsubst` ships in the `nginx` image).
- **`docker-compose.prod.aws.yml`** — adds a `DOMAIN` build arg wired to the existing `ROOT_DOMAIN` from `.env`. Because `server-update.sh` builds per-host, each environment bakes in its own domain from the `.env` it already has.

Result: staging needs **no** modified nginx file — it just keeps `ROOT_DOMAIN=bytedeck-staging.com` in its `.env` (and its wildcard cert at `/etc/letsencrypt/live/bytedeck-staging.com/`, which it already has).

The legacy `/media/` → CloudFront redirect stays hardcoded, since staging uses the same CloudFront distribution.

### Testing?
- Rendered the template with `DOMAIN=bytedeck.com` and diffed against the previous `nginx/bytedeck.aws.conf`: **the config body is byte-identical** (63 directive lines) — production behaviour is unchanged; only the header comment differs.
- Rendered with `DOMAIN=bytedeck-staging.com` and verified all five domain sites (2× `server_name`, the host regex, 3× cert paths) substitute correctly while `$host`/`$scheme`/`$request_uri`/`$1` and the CloudFront redirect are preserved verbatim.
- Validated `docker-compose.prod.aws.yml` YAML and confirmed no remaining references to the old filename.

n/a — infrastructure only.

### Anything Else?
- No app code changes; nothing to deploy differently other than the `DOMAIN` build arg (auto-derived from the existing `ROOT_DOMAIN`).
- Follow-up (out of scope): the old legacy `nginx/bytedeck.conf` (non-AWS) is still present and unused; can be removed in a later cleanup.

### Review request
@tylerecouture

- Claude Code

---

## Summary by CodeRabbit

- **New Features**
  - Nginx configurations now support customizable domains for production and staging environments.
  - TLS certificate paths and hostnames are generated automatically from the configured domain.

- **Documentation**
  - Updated deployment documentation to explain shared domain-based configuration and forwarded request scheme handling.

### DIFF
--- a/docker-compose.prod.aws.yml
+++ b/docker-compose.prod.aws.yml
@@ -34,7 +34,7 @@ services:
         # Domain baked into the nginx config template (server_name, host check,
         # and cert paths). Reuses ROOT_DOMAIN from .env so production and
         # staging build the same image from one template.
-        DOMAIN: ${ROOT_DOMAIN}
+        DOMAIN: $ROOT_DOMAIN
     volumes:
         - /etc/letsencrypt:/etc/letsencrypt
     #   - static_data:/var/www/bytedeck/static
--- a/docker-compose.prod.aws.yml
+++ b/docker-compose.prod.aws.yml
@@ -31,6 +31,10 @@ services:
       args:
         WUID: $WUID
         WGID: $WGID
+        # Domain baked into the nginx config template (server_name, host check,
+        # and cert paths). Reuses ROOT_DOMAIN from .env so production and
+        # staging build the same image from one template.
+        DOMAIN: ${ROOT_DOMAIN}
     volumes:
         - /etc/letsencrypt:/etc/letsencrypt
     #   - static_data:/var/www/bytedeck/static
--- a/nginx/Dockerfile
+++ b/nginx/Dockerfile
@@ -5,13 +5,17 @@ COPY uwsgi_params /etc/nginx/uwsgi_params
 COPY nginx.conf /etc/nginx/nginx.conf
 
 # Render the site config from a single template shared by production and
-# staging. envsubst is restricted to '${DOMAIN}' so it substitutes only the
-# domain and leaves nginx's own $host/$scheme/$request_uri/$1 variables intact.
-# DOMAIN is a build arg wired to ROOT_DOMAIN (see docker-compose.prod.aws.yml);
-# the build happens per-host, so each environment bakes in its own domain.
-ARG DOMAIN=bytedeck.com
+# staging. DOMAIN is REQUIRED and has no default on purpose: it is supplied per
+# environment from ROOT_DOMAIN via docker-compose.prod.aws.yml (bytedeck.com for
+# production, bytedeck-staging.com for staging), so hardcoding a default here
+# would be misleading and could silently ship the wrong domain. The guard below
+# fails the build loudly if it is missing. envsubst is restricted to '${DOMAIN}'
+# so it substitutes only the domain and leaves nginx's own
+# $host/$scheme/$request_uri/$1 variables intact.
+ARG DOMAIN
 COPY bytedeck.conf.template /tmp/bytedeck.conf.template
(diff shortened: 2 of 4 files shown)


nginx/Dockerfile
```diff
@@ -3,7 +3,17 @@ FROM nginx:stable
...
 COPY uwsgi_params /etc/nginx/uwsgi_params
 COPY nginx.conf /etc/nginx/nginx.conf
-COPY bytedeck.aws.conf /etc/nginx/sites-available/bytedeck.conf
+
+# Render the site config from a single template shared by production and
+# staging. envsubst is restricted to '${DOMAIN}' so it substitutes only the
+# domain and leaves nginx's own $host/$scheme/$request_uri/$1 variables intact.
+# DOMAIN is a build arg wired to ROOT_DOMAIN (see docker-compose.prod.aws.yml);
+# the build happens per-host, so each environment bakes in its own domain.
+ARG DOMAIN=bytedeck.com
```
tylerecouture: Why is it set to "bytedeck.com" here?  Is this a default value?  It's confusing here, because what about bytedeck-staging.com?  Either needs removed or a thorough comment explaining.

docker-compose.prod.aws.yml
```diff
@@ -31,6 +31,10 @@ services:
       args:
         WUID: $WUID
         WGID: $WGID
+        # Domain baked into the nginx config template (server_name, host check,
+        # and cert paths). Reuses ROOT_DOMAIN from .env so production and
+        # staging build the same image from one template.
+        DOMAIN: ${ROOT_DOMAIN}
```
tylerecouture: Why does the $DOMAIN have braces around it, but the $WUID and $WGID don't?

nginx/Dockerfile
tylerecouture: Good catch — it was a build-arg *default* (fallback) that was never actually used, since `docker-compose.prod.aws.yml` always passes `DOMAIN`. It read as if prod were hardcoded, which is exactly the confusion you hit. Fixed in `1ffe9e1b`: removed the default (`ARG DOMAIN`) and added a build-time guard that fails the build loudly if `DOMAIN` isn't supplied, plus a comment explaining it comes from `ROOT_DOMAIN` per environment.

- Claude Code

---

docker-compose.prod.aws.yml
tylerecouture: No good reason — in Compose `$VAR` and `${VAR}` are equivalent (braces only matter when the name butts against other characters, e.g. `${VAR}_x`). Pure inconsistency on my part; changed to `$ROOT_DOMAIN` in `1ffe9e1b` to match the `$WUID`/`$WGID` style.

- Claude Code

---

PR #1954, opened 2026-07-15
Self-host Redis (drop ElastiCache) + production restart policies & healthchecks

tylerecouture:
### What?
Two related production-reliability changes (the second was added after review started — see the last commits):

1. **Self-hosted Redis**: replaces AWS ElastiCache with a `redis:7-alpine` container on the EC2 host, for both production and staging. Redis remains the Celery **broker** (db 0) and Django **cache** (dbs 1/2).
2. **Resilience batch**: `restart: unless-stopped` + healthchecks for **every** production service (web, celery, celery-beat, redis, nginx).

### Why?
- ElastiCache is comparatively expensive for this workload; a locked-down Redis container on the existing app host costs nothing extra. Also retires the **EOL Redis 5.0** image dev was on.
- Production previously had **no restart policies and no healthchecks** on the app services — a crashed container stayed down until someone noticed. A dead celery-beat is especially silent: all periodic tasks just stop.

### How?
**Redis (consolidated per review feedback):**
- Shared config (image `redis:7-alpine`, healthcheck, sysctls, network, and the web/celery/celery-beat `depends_on redis healthy` wiring) now lives in the base **`docker-compose.yml`** — no more duplication between dev and prod.
- **`docker-compose.override.yml`** (dev) adds only the published port (`6379`) for local `runserver`/tests.
- **`docker-compose.prod.aws.yml`** adds only production concerns: `restart: unless-stopped` and the hardened command — memory capped via `--maxmemory` (`REDIS_MAXMEMORY`, default `256mb`), `--maxmemory-policy noeviction` (queued tasks never silently dropped), persistence disabled (broker/cache are disposable). **Internal-only**: no published port, `backend-network` only → no password needed in this single-host setup.
- `.env.example.aws`: `REDIS_HOST=redis`, documented `REDIS_MAXMEMORY`.

**Resilience (`docker-compose.prod.aws.yml`):**
- `restart: unless-stopped` on web, celery, celery-beat, redis, nginx — crashed containers come back automatically, including after host reboot.
- Healthchecks:
  - `web`: TCP connect to the uwsgi socket (`:8000`), `start_period: 10m` to cover the `migrate_schemas`/`collectstatic` startup run.
  - `celery`: `celery inspect ping` against the worker over the broker (catches hung/disconnected workers, not just dead processes), 60s interval.
  - `celery-beat`: `pgrep` process check (beat has no ping command; `procps` is in the app image).
  - `redis`: `redis-cli ping` (shared base config).
- Compose restarts *crashed* containers but does **not** auto-restart ones that merely turn unhealthy — the healthcheck is a monitoring signal (`docker compose ps` shows `(healthy)`/`(unhealthy)`); README documents this.

### Testing?
- Validated **both** merged configs with `docker compose config` after each change:
  - prod: redis resolves internal-only with the hardened command; all five services show `restart: unless-stopped`; all healthchecks present; `web`/`celery`/`celery-beat` depend on healthy redis.
  - dev: same shared redis base + published port; `depends_on` (db + redis) intact.
- Confirmed the resolved prod/dev behaviour is unchanged by the consolidation (same effective config, just de-duplicated).

n/a — infrastructure only.

### Anything Else?
**Deploy step (per host):** set `REDIS_HOST=redis` in the server's `.env`, redeploy with `./production/server-update.sh`, then enable the host tuning once: `systemctl enable --now redis-host-setup` (THP off / `vm.overcommit_memory=1`). No data migration — broker/cache are disposable — then decommission ElastiCache.

Tunables flagged for review: `REDIS_MAXMEMORY` default (256mb) and the `noeviction` policy; happy to adjust.

### Review request
@tylerecouture

- Claude Code

### DIFF
--- a/docker-compose.prod.aws.yml
+++ b/docker-compose.prod.aws.yml
@@ -21,42 +21,17 @@ services:
       # - media_data:/app/media
       # - socket_data:/bytedeck-volume
     user: $WUID:$WGID
+    # Restart policies are production/staging-only on purpose: in development a
+    # crashed process should stay down and visible, not quietly loop-restart.
+    # (Healthchecks are shared and live in docker-compose.yml.)
     restart: unless-stopped
-    # Liveness: uwsgi listening on its socket. TCP connect is enough (it speaks
-    # the uwsgi protocol, not HTTP). The long start_period covers the
-    # migrate_schemas + collectstatic run before uwsgi binds -- failures during
-    # start_period don't count against the container.
-    healthcheck:
-      test: ["CMD", "python", "-c", "import socket; socket.create_connection(('127.0.0.1', 8000), timeout=5).close()"]
-      interval: 30s
-      timeout: 10s
-      retries: 5
-      start_period: 10m
 
   celery:
     user: $WUID:$WGID
     restart: unless-stopped
-    # Liveness: ask this worker to answer a ping over the broker. Catches a
-    # worker that is running but disconnected/hung, not just a dead process.
-    # Deliberately infrequent -- each check runs a full celery client.
-    healthcheck:
-      test: ["CMD-SHELL", "cd /app/src && celery -A hackerspace_online inspect ping -d celery@$$HOSTNAME -t 10 | grep -q pong"]
-      interval: 60s
-      timeout: 30s
-      retries: 3
-      start_period: 2m
 
   celery-beat:
     restart: unless-stopped
-    # Liveness: beat has no ping command, so check the scheduler process exists
-    # (procps/pgrep is installed in the app image). A dead beat silently stops
-    # all periodic tasks, so even a coarse signal is worth surfacing.
-    healthcheck:
-      test: ["CMD-SHELL", "pgrep -f 'celery.*beat' > /dev/null"]
-      interval: 60s
-      timeout: 10s
-      retries: 3
-      start_period: 2m
 
   # Production/staging additions to the shared redis service (image, healthcheck,
   # network — see docker-compose.yml). Self-hosted on this EC2 host, replacing
--- a/docker-compose.prod.aws.yml
+++ b/docker-compose.prod.aws.yml
@@ -21,9 +21,54 @@ services:
       # - media_data:/app/media
       # - socket_data:/bytedeck-volume
     user: $WUID:$WGID
+    depends_on:
+      redis:
+        condition: service_healthy
 
   celery:
     user: $WUID:$WGID
+    depends_on:
+      redis:
+        condition: service_healthy
+
+  celery-beat:
+    depends_on:
+      redis:
+        condition: service_healthy
+
+  # Self-hosted Redis (Celery broker on db 0, Django cache on db 1/2), replacing
+  # AWS ElastiCache. It is reachable ONLY on the internal backend-network — no
+  # published port — so in this single-host setup it needs no password.
+  redis:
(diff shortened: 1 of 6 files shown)


tylerecouture (commented):
@claude Are we duplicating redis info between our development environment (in override file) and in production (aws file)?  Now that both will be using a redis container, we should consolidate the common config in the main yml file, then any overrides in the override (development) and aws (production/staging) files.

tylerecouture:
Good call — done in `cf4299a2`. The shared redis config (image `redis:7-alpine`, healthcheck, sysctls, network, and the web/celery/celery-beat `depends_on redis healthy` wiring) now lives in the base `docker-compose.yml`; the override (dev) adds only the published `6379` port, and the AWS file adds only production concerns (`restart: unless-stopped` + the maxmemory/noeviction/no-persistence command). Verified both merged configs with `docker compose config` — resolved behaviour is unchanged, just de-duplicated.

Heads-up that I also stacked the resilience batch onto this PR in `3c332d1c` (restart policies + healthchecks for web/celery/celery-beat/nginx) since it edits the same service blocks — the PR title/description are updated to cover both. Happy to split it out to its own PR if you'd rather keep this one Redis-only.

- Claude Code

---

docker-compose.prod.aws.yml
```diff
@@ -21,9 +21,66 @@ services:
       # - media_data:/app/media
       # - socket_data:/bytedeck-volume
     user: $WUID:$WGID
+    restart: unless-stopped
```
tylerecouture, line 27: @claude Are these restarts and healthchecks on all the apps not relevant in development too?  Also, we only set depends-on to redis, but what about the other dependancies?  Do any container depends on nginx, or celery, or web, etc?

tylerecouture: Both good questions:

**Healthchecks/restarts in dev?** Half yes — done in `9c75886c`. **Healthchecks** are observability and are just as useful in development, so the web/celery/celery-beat checks moved into the base `docker-compose.yml` (dev now shows `(healthy)`/`(unhealthy)` in `docker compose ps` too). **Restart policies** stay production-only on purpose: in dev you want a crashed `runserver`/worker to stay down and visible while you debug, not quietly loop-restart.

**Other depends_on?** The full dependency graph is already wired:
- `nginx → web` (exists in the AWS file) — nginx's upstream is `web:8000`, so web must exist for DNS to resolve.
- `web`/`celery`/`celery-beat` → `redis` healthy (base file) — all three need the broker/cache.
- `web`/`celery`/`celery-beat` → `db` healthy in **dev only** (override file). In prod the database is **RDS**, which isn't a compose service, so there's nothing to `depends_on` — the app just connects over the network.
- Nothing depends **on** celery/celery-beat (fire-and-forget consumers), and nothing depends on nginx (it's the front door).

The one candidate I deliberately left out: making `celery-beat` wait for `web` to be *healthy* (i.e. migrations finished) before starting. It would guarantee ordering on a completely fresh database, but `docker compose up -d` blocks on health conditions — so every deploy would wait out web's startup, and a temporarily-unhealthy web would fail the whole `up` (and the systemd unit) even though web itself was running. Since `restart: unless-stopped` already self-heals the fresh-DB case (beat crashes, retries, succeeds once migrations land), the looser coupling seemed safer. Happy to add it if you'd rather have strict ordering.

- Claude Code

---

PR #1964, opened 2026-07-15
Prevent concurrent double-start of quests from granting XP twice (partial unique constraint)

### LINKED ISSUE #1345: Quests are at times repeatable when they shouldn't be
A student completed a non-repeatable quest, but it reappeared in the available page. She was able to complete it again and get the XP for it twice.

https://studiotyee.bytedeck.com/quests/submission/40128/ 

https://studiotyee.bytedeck.com/quests/submission/40127/

tylerecouture:
### What?
Stop a concurrent "double start" (e.g. a student opening two browser tabs and starting the same quest twice before either commits) from creating two in-progress submissions of a non-repeatable quest — which could then both be completed and approved, granting XP twice. Closes #1345.

### Why?
`start()` (`quest_manager/views.py`) gates on `quest.is_available()` → `QuestSubmissionManager.not_submitted_or_inprogress()`, which only *counts* existing submissions. There was no lock, no `transaction.atomic`, and no DB uniqueness on `QuestSubmission`, so two near-simultaneous requests both passed the guard before either row was committed, and both created a submission. This matches the report exactly: both samples were auto-approve quests where ordinal 1 was approved *after* ordinal 2, on a high-traffic first day of classes — and the issue thread shows two submission URLs plus double XP.

I reproduced the race deterministically (test in the issue comments) before implementing this fix.

### How?
* **DB-level guard**: a partial `UniqueConstraint` on `QuestSubmission` — `(user, quest, semester)` where `is_completed=False AND first_time_completed IS NULL` — i.e. at most one **never-yet-completed** in-progress submission of a quest per semester. Both rows created by a double start are in that state, so the race is blocked. The condition deliberately exempts:
  * **completed** submissions (`is_completed=True`), so repeatable quests still accumulate completed submissions; and
  * **returned** submissions — `mark_returned()` flips `is_completed` back to `False`, but `first_time_completed` stays set from the first completion — so a teacher returning an old submission of a repeatable quest while the student has a new one in progress remains a legal state (review catch by @tylerecouture).
* **Graceful race resolution**: `create_submission()` wraps the insert in `transaction.atomic()` and, if it loses the race (`IntegrityError`), returns the in-progress submission that won instead of creating a duplicate — so `start()` just redirects the student to it, same as the existing "already in progress" path.
* **Safe migration**: `0049` first deletes any pre-existing duplicate never-completed in-progress rows (keeping the earliest per `(user, quest, semester)`; completed, returned, and NULL-semester rows untouched) so adding the constraint can't fail on existing decks.

### Testing?
New `quest_manager/tests/test_regression_1345.py`:
* the constraint rejects a second fresh in-progress submission of the same quest/semester (`IntegrityError`);
* calling `create_submission()` twice (the double-start race) leaves exactly **one** submission and the second call returns the first;
* the reported symptom is dead: a double start of a non-repeatable quest yields one approved submission and single XP (`calculate_xp == quest.xp`);
* regression guard: a repeatable quest can still be started again after completing the previous submission (ordinal 2 created normally);
* **return workflow guard**: a teacher can return an old submission of a repeatable quest while a new one is in progress — two `is_completed=False` rows, no `IntegrityError`;
* migration dedup coverage: keeps the earliest fresh duplicate, deletes the rest, and spares completed, returned, and NULL-semester rows (migration at 100% branch coverage).

Also fixed three pre-existing test fixtures that relied on states the constraint now forbids — multiple fresh in-progress rows of one quest, or `is_approved=True` without `is_completed=True`, an impossible state in real flows since `mark_approved()` always sets `is_completed=True` (two `calculate_xp` tests in `test_managers.py`, and the ajax-count fixture in `test_views.py`, whose in-progress distractor rows now use distinct quests).

Full suite run locally (`test src`, 1030 tests): the only failures are the known order-dependent flakes (`QuestPrereqsUpdate.test_post_save_button__*` documented in CLAUDE.md, plus two djcytoscape view tests) — all four pass standalone and none involve the new constraint. `flake8 src` clean; `makemigrations --check` clean.

N/A — model/manager + migration only; no UI change. The losing tab is transparently redirected to the surviving submission.

### Anything Else?
* The constraint scopes to `semester` (matching how submissions live per-semester); NULL-semester rows are unaffected (Postgres treats NULLs as distinct) and are skipped by the migration's dedup pass.
* An additional belt-and-braces option discussed in the issue — `select_for_update()` around the availability check in `start()` — wasn't needed once the DB constraint exists; the constraint is the authoritative guard and is race-proof by construction.

### Review request
@tylerecouture

- Claude Code

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate “never-yet-completed” in-progress quest submissions for the same user/quest/semester during concurrent starts.
  * Made quest submission creation race-safe by recovering the already-created in-progress record when the database rejects a duplicate.
  * Ensured XP/approval is effectively applied only once for the duplicate-start race, while preserving repeatable quest restart/return behavior.
  * Added a migration to deduplicate existing duplicates while exempting non-applicable rows.

* **Tests**
  * Added regression and migration-level coverage for the double-start race and deduplication.
  * Updated XP and submission-count tests to reflect correct completion states and counting rules.


tylerecouture:
Re the (pending) review question on `mark_returned()` — *"What happens if an old submission of a repeatable quest is returned, which they already had a new one in progress?"*

Great catch — as originally written the constraint **would** have broken that workflow: `mark_returned()` flips `is_completed` back to `False`, so returning an old submission while a new one was in progress would have violated the unique constraint and crashed the return action with an `IntegrityError`.

Fixed in 7cdb812 by narrowing the constraint's condition to `is_completed=False AND first_time_completed IS NULL` ("never yet completed"):

* Both rows created by a concurrent **double start** are in that state, so the #1345 race is still blocked.
* A **returned** submission keeps its `first_time_completed` from the first completion, so it's exempt — two "in progress" rows (one fresh, one returned) are a legal state again.

The migration's dedup pass and `create_submission()`'s lost-race lookup were narrowed to the same scope, and there's a new regression test for exactly this scenario (`test_returning_old_submission_while_new_one_in_progress_is_allowed`), plus dedup coverage that returned rows survive. All regression tests pass; migration still at 100% branch coverage.

(Replying here because the review comment is still in a pending/unsubmitted review, which the API won't let me thread onto — it also never reached this session as a webhook, for the same reason.)

- Claude Code

---

PR #1972, opened 2026-07-16
Self-heal missing allauth account tables before account.0003 (develop twin of #1971)

tylerecouture:
### What?
Identical cherry-pick of the staging hotfix **#1971**: adds `tenant.0015_ensure_allauth_account_tables`, a self-healing data migration that recreates django-allauth `account` tables in schemas whose migration history records them as applied but where they were never created, guaranteed (`run_before`) to run before allauth's `account.0003`.

### Why?
The allauth `>=65` upgrade (#1916) crash-loops any long-lived database with the phantom-history state — staging's public schema is broken right now (see #1971), and older development databases will hit the same wall. Landing the same migration in `develop` self-heals dev environments and keeps `develop`/`staging` consistent (the file is byte-identical, so future `develop → staging` merges are clean).

### How? / Testing?
See #1971 — same commit, same migration, same `TenantTestCase` tests covering the repair path (tables dropped → recreated) and the no-op path (healthy schema untouched).

n/a.

### Anything Else?
Merge whichever of this and #1971 first; the other stays clean since the content is identical.

### Review request
@tylerecouture

- Claude Code

---

### DIFF
--- a/src/tenant/migrations/0015_ensure_allauth_account_tables.py
+++ b/src/tenant/migrations/0015_ensure_allauth_account_tables.py
@@ -1,0 +1,59 @@
+# Self-healing repair for schemas whose django-allauth `account` tables are
+# missing even though the migration history records them as applied.
+#
+# Background: on long-lived databases (production/staging public schema, some
+# development environments) the early `account` migrations were recorded as
+# applied without the tables ever being created (faked history from when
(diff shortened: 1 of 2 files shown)


tylerecouture:
I merged the hotfix in staging back into develop instead of a parallel commit of this PR

PR #2018, opened 2026-07-18
Show a spinner while bootstrap-tables load, no unstyled flash (#1981)

### LINKED ISSUE #1981: Bootstrap Tables slow to load, show in place animation?
Currently, when the tables load, they show the raw table, preformatting first for a second or two (on large tables):

Image

Then after the formatting and bootstrap-table loads up, it looks like this:

Image

How to prevent the format from jumping between the two?  Can we render "holding" images/adnimations that get replaced with the proper formatting ones it loads?  Similar to, for example, Facebook does this:

Image

tylerecouture:
### What?
Stops data tables from flashing their raw, unstyled HTML and then "jumping" into the formatted bootstrap-table a second or two later. Instead the table area shows a **loading spinner** until bootstrap-table has finished, then the finished table appears in place.

Closes #1981

### Why?
Every data table is server-rendered as a plain `<table data-toggle="table">`, and bootstrap-table reformats it (pagination / search / fixed layout) on load. On large tables the browser paints the raw table first and reflows into the styled one after — the visible jump in the issue. It affects ~15 table pages (quests, approvals, submissions, badges, portfolios, campaigns, tags, profiles, library…).

### How?
Two small changes to already-global assets, so every table is covered uniformly (bootstrap-table itself is loaded per-section at a few different versions):

- **`custom_common.css`** (loaded render-blocking in `<head>`): hides `table[data-toggle="table"]` from first paint so the raw table is never shown, and takes it out of flow while hidden so a large table doesn't reserve a tall blank gap. Once bootstrap-table wraps the table in `.bootstrap-table` it's revealed again — a version-independent, JS-free reveal. Also styles the spinner.
- **`custom.js`** (loaded site-wide): drops a spinner into the table's place while it formats and removes it once the table is wrapped. It **polls** for the wrapper, so it works regardless of script load order or bootstrap-table version, and has a timeout fallback that reveals the raw table if bootstrap-table never runs (e.g. a CDN hiccup) — a script failure can't hide the data permanently.

### Testing?
No server-side code changed, so there's no Django test to add here. I verified the behavior in **headless Chromium** against the bundled bootstrap-table assets, simulating a delayed init:

- **During load:** the raw table is not painted (no unstyled flash), the content below it doesn't get pushed down (hidden table is out of flow), and the spinner shows in the gap.
- **After init:** the table is revealed and fully functional — search box, sortable columns, "Showing 1 to 10 of 180 rows", page-size selector, and pagination all render.
- **Fallback:** with bootstrap-table never loading, after the timeout the raw table is revealed (all rows visible, spinner removed) — so the data is never stuck hidden.

Front end is affected. Behavior change is temporal (no single-frame diff): the raw-table flash/jump is replaced by a centered spinner that swaps to the finished table. **Worth a quick eyeball on a real deck** with a large table (e.g. the student list) to confirm it reads well with the site's Bootstrap 3 + Font Awesome styling loaded.

### Anything Else?
This is the lightweight spinner/fade approach. If you ever want the fancier Facebook-style skeleton rows, that'd be a follow-up (per-table row/column shapes).

### Review request
@tylerecouture

- Claude Code

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented an unstyled table from briefly appearing while enhanced tables load.
  * Added a loading indicator for tables awaiting initialization.
  * Ensured tables remain visible if enhanced formatting is delayed or unavailable.

### DIFF
--- a/src/static/css/custom_common.css
+++ b/src/static/css/custom_common.css
@@ -1108,3 +1108,37 @@ div.btn-group.action-btn-group {
 
 
 
+
+/* === #1981: prevent the bootstrap-table "flash of unstyled table" ================= */
+/* Data tables (data-toggle="table") are server-rendered as a plain <table>, so the
+   browser paints the raw, unstyled table first and then — a second or two later on large
+   tables — bootstrap-table tears it down and rebuilds it (pagination, search, fixed layout).
+   That rebuild is the visible "jump". This file is loaded in the render-blocking <head>, so
+   hiding the raw table here stops it from ever being painted; custom.js drops a spinner into
+   its place and the rules below reveal the finished table. */
+table[data-toggle="table"] {
+    /* take the raw table out of flow while hidden so a big table doesn't reserve a tall
+       blank gap under the spinner */
+    position: absolute;
+    visibility: hidden;
+}
+
+/* Once bootstrap-table wraps the table in .bootstrap-table it is fully formatted — put it
+   back in flow and show it. Version-independent and JS-free, so the table always appears. */
+.bootstrap-table table[data-toggle="table"],
+table[data-toggle="table"].bt-reveal {
+    position: static;
+    visibility: visible;
+}
+
+/* Spinner shown in the table's place while it formats (inserted by custom.js). */
+.bt-loading {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    min-height: 120px;
+    color: #999;
+}
+.bt-loading .fa {
+    opacity: 0.6;
+}
--- a/src/static/js/custom.js
+++ b/src/static/js/custom.js
@@ -36,4 +36,40 @@ $(document).ready(function() {
        window.location.href = $(this).attr("href");
      });
 
+    // #1981: bootstrap-table reformats server-rendered tables on load, briefly showing the raw
+    // table before it "jumps" into the styled/paginated version. The head CSS (custom_common.css)
+    // hides those tables until they're formatted; here we show a spinner in the gap and make sure
+    // the table is revealed even if bootstrap-table's script never runs.
+    $('table[data-toggle="table"]').each(function() {
+        var $table = $(this);
+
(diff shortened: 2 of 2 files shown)


tylerecouture (commented):
Is this the correct place for this code?  Check the codebase and the areas we are using bootstrap table, is there a more rpecise way of injecting this into only the pages that use bootstrap table?  If not, or it doesn't make sense, then leave it here.

tylerecouture:
Good question — I checked the codebase. Short version: it's in the right place, and here's the reasoning.

**CSS (`custom_common.css`)** — this has to be render-blocking `<head>` CSS. If it loaded later (per section), the raw table would already have painted before it could be hidden, and that first paint *is* the flash we're fixing. It also belongs right next to bootstrap-table's own stylesheet — which is *already global*: `bootstrap-table.min.css` is loaded site-wide in `head_css.html` (included by `base.html` on every page). So the styling half of bootstrap-table is already global; putting the loading CSS there is consistent, not new sprawl.

**JS (`custom.js`)** — it's a no-op on any page without a `table[data-toggle="table"]` (just one empty jQuery lookup). bootstrap-table's *JS* is indeed included per section — badges, profiles, tags, library, and ~8 `quest_manager` templates, at three different versions (1.18.3 / 1.20.2 / the bundled static one). I could add a small `bootstrap-table-loading.js` after each of those includes, but that means duplicating it across 8+ templates, keeping it in sync with those mixed versions, and remembering it for every future table page. Since `custom.js` is already global and the code self-scopes to `data-toggle="table"`, global is simpler, DRY, and automatically covers new table pages.

So my recommendation is to leave it where it is — it mirrors how bootstrap-table's CSS is already wired up, and the render-blocking requirement rules out per-section CSS anyway. If you'd still prefer the tighter scoping for the JS despite the duplication, I'm happy to move it into a per-section include — just say the word.

- Claude Code

---

PR #2025, opened 2026-07-19
Test suite cleanup: baker GFK determinism + a naming/docstring convention guard

tylerecouture:
### What?

First installment of a broader test-suite cleanup. This PR lands the **correctness fix** and the **infrastructure** to drive consistency; the bulk per-app naming/docstring burn-down follows (see below), gated on your feedback about scope.

**1. Correctness — `baker.make(CytoScape)` GFK determinism**
`CytoScape.initial_content_type` is a non-nullable `ForeignKey(ContentType)` (uniquely constrained with `initial_object_id`), so bare `baker.make(CytoScape)` pointed the map's initial object at a *random, non-existent* content type — I verified it differs on every call (12 distinct models over 12 makes). Added a `bake_scape()` helper pinning it to a real `Quest` and used it at the three affected sites (the `_quantity=4` case cycles distinct Quest ids to satisfy the uniqueness constraint).

**2. Infrastructure — a convention guard (`hackerspace_online/tests/test_conventions.py`)**
AST-scans every `test_*.py` and asserts each `test_*` method follows `test_subject__case` and has a docstring, and that `setUp`/`setUpTestData`/`tearDown` have docstrings. Modules predating the conventions are grandfathered in a `LEGACY_UNCONVENTIONAL` denylist (seeded with the 79 currently-violating modules) that is **meant to shrink**. New/cleaned modules are enforced and can't regress; a second test keeps the denylist honest (fails if an entry no longer exists or already conforms).

**3. First burn-down — 9 modules cleaned** (the admin/template-tag/model tests from the recent coverage sweeps): renamed to `test_subject__case`, added missing method/`setUp` docstrings, removed from the denylist. Denylist: 79 → 70.

Also removed a stray debug `print()` in `test_management_commands.py`.

### Why?

From a full review of the ~24k-LOC suite. Notably, several things that *looked* like debt turned out to be fine on inspection, and I deliberately did **not** "fix" them:
- `baker.make(Comment)` without a target is **safe** — its GFK is nullable, so baker leaves it `None` (unlike Prereq/CytoScape's non-nullable ones).
- The `args=[1]` / `pk=1` view tests target **migration-seeded** objects (default Rank/Semester/Block/Course/MenuItem) whose pks are stable in the reused schema — green and correct, not fragile.
- Most "raw" `assertEqual(response.status_code, …)` checks operate on **already-fetched POST responses** the `ViewTestUtilsMixin` helpers can't express — legitimate, not inconsistency.

So the real, durable wins are the GFK determinism fix and a guard that stops new naming/docstring debt while enabling an incremental cleanup.

### How?

See per-commit breakdown. The guard runs in the normal `manage.py test` suite (no CI/workflow changes needed) and is `SimpleTestCase` (AST only, no DB).

### Testing?

- New/changed modules pass, including the guard (43 tests in the affected set); full djcytoscape app green.
- `ruff check src` clean.

### Anything Else?

**Question on scope for the remaining 70 modules.** Only ~29% of the suite's 1,165 test methods currently follow `test_subject__case`, so a full burn-down means renaming ~800 existing methods + adding ~500 docstrings — a large, git-blame-churning change. Two ways to proceed, and I'd like your call:
1. **Incremental (my default):** keep the guard; clean modules opportunistically as they're touched. No big-bang rename.
2. **Full sweep now:** I continue burning the denylist to empty across all apps in follow-up commits on this branch.

Happy to go either way — the guard makes both safe.

### Review request
@tylerecouture

- Claude Code

---

## Summary by CodeRabbit

- **Tests**
  - Added automated checks enforcing descriptive test names and docstrings across the test suite.
  - Improved test names and documentation for template tags, models, admin actions, and media behavior.
  - Strengthened test fixtures to use valid related records.
  - Simplified database initialization test output handling.


tylerecouture:
Addressed the CodeRabbit review in `fdb0c67`:
- Added docstrings to `test_save__sets_first_scape_as_primary` / `test_save__changes_primary_scape`.
- Passed `encoding="utf-8"` to the guard's `read_text()`.

Skipped the `full_clean()`-before-`save()` suggestion in `test_save__changes_primary_scape`: that `save()` *is* the behavior under test (saving a map flagged primary demotes the old primary), and per CLAUDE.md `full_clean()` is "preferred in new code … but match the surrounding style rather than adding drive-by `full_clean()` calls" — same call we made (and you agreed with) on #2020. Adding it here would also risk unrelated validation failures on fields the test doesn't touch.

Note: `djcytoscape/tests/test_models.py` is still in the guard's `LEGACY_UNCONVENTIONAL` list, so its remaining tests aren't yet held to the convention — that module's full cleanup is part of the pending burn-down.

- Claude Code

---

PR #2026, opened 2026-07-19
Bring the whole test suite up to the naming + docstring conventions (empty the guard denylist)

tylerecouture:
### What?

Completes the convention burn-down started in #2025. **Every** `test_*.py` module now conforms to the project's test conventions, so the guard's `LEGACY_UNCONVENTIONAL` denylist is **emptied** — naming + docstrings are now enforced across the entire suite, with no grandfathered files.

Across all 17 apps (71 modules) this:
- **renames ~800 test methods** to `test_<subject>__<specific_case>`,
- **adds ~500 missing docstrings** (test methods + `setUp`/`setUpTestData`/`tearDown`), and
- **prunes a handful of dead commented-out test blocks**.

No test logic, assertions, fixtures, imports, or behavior were changed.

### Why?

You asked to do the full burn at once. Only ~29% of the suite's 1,165 test methods followed the convention before this; now 100% do, and the guard (`hackerspace_online/tests/test_conventions.py`) locks it in so it can't regress.

### How?

The work was fanned out across the apps and then verified centrally. The mechanical edits were name/docstring/dead-code only.

**One case needed real care.** Renaming reorders methods (they run alphabetically), which surfaced a *pre-existing* order dependence in `TenantAdminViewPermissionsTest`: `test_delete_view` opens with an anonymous request through the public `TenantClient`, which leaves the DB connection on the public schema so the sibling tests' `force_login()` (their first DB touch) can resolve the public-schema users. When a rename pushed it to sort after a `force_login`-first sibling, that sibling ran first and hit `Save with update_fields did not affect any rows` in `update_last_login`. I renamed it to `test_delete_view__enforces_permissions` so it keeps sorting first, and added a comment documenting the dependency. (The root fragility is pre-existing; this PR preserves develop's passing behavior rather than rewriting the tenant-admin tests.)

### Testing?

- **Full suite green: 1,168 tests pass** (serial, run twice — once to find the ordering issue, once to confirm the fix).
- `ruff check src` clean.
- The convention guard passes with an **empty** `LEGACY_UNCONVENTIONAL`, i.e. it now validates every module.

N/A — tests only.

### Anything Else?

This is a large diff by line count, but it's almost entirely method renames and one-line docstrings; there are no logic changes. Reviewing per-file (or trusting the green suite + guard) should be straightforward.

### Review request
@tylerecouture

- Claude Code

---

## Summary by CodeRabbit

* **Tests**
  * Standardized test names across announcements, badges, courses, quests, notifications, tenants, and other areas to clearly describe scenarios and expected outcomes.
  * Added explanatory test documentation and streamlined shared setup descriptions.
  * Expanded coverage for configuration caching, HTML sanitization, widget behavior, and quest formatting.
  * Confirmed the test suite now conforms to naming conventions without legacy exceptions.

### DIFF
--- a/src/announcements/tests/test_models.py
+++ b/src/announcements/tests/test_models.py
@@ -7,13 +7,15 @@ from hackerspace_online.tests.utils import ByteDeckTenantTestCase
 class AnnouncementTestModel(ByteDeckTenantTestCase):
     @classmethod
     def setUpTestData(cls):
+        """Create an announcement shared across the test methods."""
         cls.announcement = baker.make(Announcement)
 
-    def test_creation(self):
+    def test_creation__str_returns_title(self):
+        """An announcement is created and its string representation is its title."""
         self.assertIsInstance(self.announcement, Announcement)
         self.assertEqual(str(self.announcement), self.announcement.title)
 
-    def test_get_comments(self):
+    def test_get_comments__returns_only_connected_comments(self):
         """Test that an announcement returns comments connected to it"""
         comment1 = baker.make("comments.Comment", target_object=self.announcement)
         comment2 = baker.make("comments.Comment", target_object=self.announcement)
--- a/src/announcements/tests/test_signals.py
+++ b/src/announcements/tests/test_signals.py
@@ -16,10 +16,11 @@ class AnnouncementsSignalsTest(ByteDeckTenantTestCase):
 
     @classmethod
     def setUpTestData(cls):
+        """Create a teacher and student shared across the test methods."""
         cls.teacher = baker.make(User, username='teacher', is_staff=True)
         cls.student = baker.make(User, username='student', is_staff=False)
 
-    def test_save_announcement_signal(self):
+    def test_save_announcement_signal__creates_and_deletes_publish_task(self):
         """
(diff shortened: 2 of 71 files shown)


tylerecouture:
Addressed the actionable review comments in `1ff3886`:
- Removed the two commented-out lines I missed in `test_published__returns_only_published_quests`.
- Fixed the comment typo + three docstring typos/grammar in `courses/tests/test_models.py`.
- Added docstrings to the `_ct_pk` GFK-choice helper in the three `utilities` test modules.

**Skipping the `full_clean()`-before-`save()` suggestions** (djcytoscape `test_signals` tearDown, `hackerspace_online/test_utils`, `notifications/test_tasks`, `prerequisites/test_models`, `prerequisites/test_signals`). Each of those `save()`s just mutates a fixture in test setup/teardown, not the behavior under test, and CLAUDE.md frames `full_clean()` as "preferred in new code … but match the surrounding style rather than adding drive-by `full_clean()` calls." This is the same call we made (and you agreed with) on #2020/#2025 — this PR is a naming/docstring cleanup that deliberately changes no test behavior, and adding `full_clean()` would risk unrelated validation failures on fields these tests don't exercise. Happy to revisit if you'd prefer them added.

- Claude Code

---

PR #2039, opened 2026-07-21
Notify Library staff on push and tell the sharer it's pending review (#1949)

### LINKED ISSUE #1949: Notify Library admin on campaign/quest push, and tell the sharer it's pending review
### Summary
When content is shared to the Library it needs to go through a review step before it's published, but right now nobody is told. Two gaps:

1. **Notify the Library admin.** Send an email to the Library's admin when a new **campaign** is pushed to the Library, so they know there's something to review and publish. (This is the notification called for in #1850: *"After sharing, a notification should be sent to Library staff so they can review and publish the imported campaign and quests manually."*)
2. **Set expectations for the person sharing.** After someone pushes a campaign to the Library, tell them that a Library admin will be notified and that the campaign will only appear once it has been reviewed/published — so they don't expect it to show up immediately.

**The same applies to quests** — pushing an individual quest to the Library should notify the Library admin and give the sharer the same "pending review" messaging.

### Why
Shared campaigns/quests land in the Library unpublished and require manual review before they're visible. Without a notification the admin has no signal that new content is waiting, and without the confirmation messaging the sharer is left wondering why their campaign/quest didn't appear.

### Acceptance criteria
- [ ] Pushing a **campaign** to the Library sends an email to the Library admin (whoever should review submissions) identifying the campaign, who shared it, and where to review/publish it.
- [ ] Pushing a **quest** to the Library does the same.
- [ ] After a successful campaign push, the sharer sees a message stating an admin has been notified and the campaign will appear once reviewed/published.
- [ ] After a successful quest push, the sharer sees the equivalent message.
- [ ] Emails are sent asynchronously (via the existing Celery email task) so the share request doesn't block on SMTP.
- [ ] Tests cover both the notification being sent and the confirmation messaging, for campaigns and quests.

### Context / related
- Campaign share button + confirmation page: #1850 (the review-notification step was specced there)
- Campaign export/share feature: #1849, #1884
- Quest export to Library: #1848, #1860

> Open question for whoever picks this up: who is the "Library admin" recipient — the Library tenant's `SiteConfig.deck_owner`, a configured address, or Library staff users? Worth confirming before implementing.

- Claude Code

tylerecouture:
### What?

When a teacher pushes a quest or campaign to the Shared Library, this PR:

1. **Emails the active Library staff** that new content is awaiting review, with a direct link to review and publish it.
2. **Tells the sharer** — via the success message — that the content won't appear in the Library until a Library admin reviews and publishes it, so they aren't left expecting it to show up immediately.

Closes #1949.

### Why?

Shared content lands in the Library *unpublished* and is invisible to other decks until a Library admin manually reviews and publishes it. Until now there was no email nudge to the reviewers (only an in-app notification, which is easy to miss on a schema they rarely visit), and the sharer got a plain "has been shared" message with no hint that a manual review gate stood between their push and the content actually being available. Both sides were left guessing.

### How?

- Added `email_library_staff_of_push(content_type, content_name, exported_obj, sharer)` in `library/views.py`. It runs inside the library schema context, collects active `is_staff` users who have an email address, builds the review/publish URL from the Library tenant's root URL + the exported object's `get_absolute_url()`, renders a plain-text template, and dispatches asynchronously via the existing `tenant.tasks.send_email_message` Celery task (queued on `default`) so the share request doesn't block on SMTP. It's a no-op when no staff member has an email (e.g. only the `deck_ai` bot user is staff).
- Wired it into `ExportQuestView.post` and `ExportCampaignView.post`, right after the existing in-app `notify.send(...)`, and updated the success message on both to explain the pending-review step.
- Added `library/templates/library/email/content_pushed.txt` for the email body.

### Testing?

- New `test_export_post__emails_library_staff_and_messages_sharer` on both `QuestLibraryTestsCase` and `CampaignLibraryTestCases`. Each patches `send_email_message.apply_async`, creates a Library staff user *with* an email (the default library staff — deck_owner/deck_ai — have no email), enables `allow_staff_export`, POSTs the export, and asserts: the task is dispatched once; the subject and body contain the content name; the sharer and the review link (with the Library host) appear in the body; and the sharer sees the pending-review success message.
- Full suite green: **1172 tests OK** (Python 3.12 / Django 5.2). `manage.py check` clean, `ruff check src/library` clean.

N/A — email + flash message only, no template/UI changes beyond the message text.

### Anything Else?

The recipient set is *all active Library staff with an email address*. Library staff whose accounts carry no email (the seeded `deck_owner`/`deck_ai` service users) are silently skipped, so a fresh Library with only service-account staff sends nothing rather than erroring.

### Review request
@tylerecouture

- Claude Code

---

### DIFF
--- a/src/library/templates/library/email/content_pushed.txt
+++ b/src/library/templates/library/email/content_pushed.txt
@@ -1,0 +1,10 @@
+{{ sharer }} shared a {{ content_type }} to the ByteDeck Library:
+
+    {{ content_name }}
+
+It has been added to the Library but is not visible to other decks until a Library admin reviews and publishes it.
+
+Review and publish it here:
+{{ review_url }}
+
+— ByteDeck Library
--- a/src/library/templates/library/email/content_pushed.txt
+++ b/src/library/templates/library/email/content_pushed.txt
@@ -1,4 +1,4 @@
-{{ sharer }} shared a {{ content_type }} to the ByteDeck Library:
+{{ sharer }} shared a {{ content_type }} from <a href="{{ source_deck_url }}">{{ source_deck_url }}</a> to the ByteDeck Library:
 
     {{ content_name }}
 
--- a/src/library/templates/library/email/content_pushed.txt
+++ b/src/library/templates/library/email/content_pushed.txt
@@ -5,6 +5,6 @@
 It has been added to the Library but is not visible to other decks until a Library admin reviews and publishes it.
 
 Review and publish it here:
-{{ review_url }}
+<a href="{{ review_url }}">{{ review_url }}</a>
 
 — ByteDeck Library
--- a/src/library/tests/test_views.py
+++ b/src/library/tests/test_views.py
@@ -1,5 +1,6 @@
 import uuid
 from copy import deepcopy
+from unittest.mock import patch
 from django.contrib.auth import get_user_model
 from django.contrib.messages import get_messages
 from django.db import connection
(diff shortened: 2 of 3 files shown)


src/library/templates/library/email/content_pushed.txt
```diff
@@ -0,0 +1,10 @@
+{{ sharer }} shared a {{ content_type }} to the ByteDeck Library:
```
tylerecouture: Can you add a "from {{ deck_name }} to the Bytedeck Library:"  where deck name is their deck url, and have the name link to their deck?

tylerecouture: Done in af94c0c0. The email now opens with the source deck named and linked:

> {sharer} shared a {content_type} from [https://theirdeck.bytedeck.com](https://theirdeck.bytedeck.com) to the ByteDeck Library:

The deck URL is resolved on the source deck (`request.tenant.get_root_url()`) before the export switches into the library schema, then rendered as an `<a>` tag — so the HTML email shows a clickable link, and the plain-text alternative (via html2text) renders it as `[url](url)`. Both export tests now assert the source deck link appears in the email body.

- Claude Code

---

src/library/templates/library/email/content_pushed.txt
```diff
@@ -0,0 +1,10 @@
+{{ sharer }} shared a {{ content_type }} from <a href="{{ source_deck_url }}">{{ source_deck_url }}</a> to the ByteDeck Library:
+
+    {{ content_name }}
+
+It has been added to the Library but is not visible to other decks until a Library admin reviews and publishes it.
+
+Review and publish it here:
+{{ review_url }}
```
tylerecouture: Should this not be a real URL href too?

tylerecouture: Good catch — fixed in 1cbd369b. The review/publish URL is now wrapped in an `<a href>` too, so it's a real clickable link in the HTML email (and renders as `[url](url)` in the plain-text alternative), consistent with the source-deck link. Both export tests now also assert the review link is a real href pointing at the Library deck.

- Claude Code

---

PR #2049, opened 2026-07-21
Cap deck name at the URL/schema limit, not 20 chars; show the limit — Closes #1975

### LINKED ISSUE #1975: Deck form 20 character warning
20 is too short.  It doesn't have to match the "short name" or the "long name" but it does have to match the longest the url can be: `this-is-too-long-for-the-tenants-app-maybe.bytedeck.com`

also the character lmit should be included in the text here: 

> The name of your deck, for example the name `example` would give you the site: `example.bytedeck.com` The name may only include lowercase letters, numbers, and dashes. It must start with a letter, and may not end in a dash nor include consecutive dashes

tylerecouture:
### What?

Deck names are no longer capped at 20 characters. The limit is now the real one — how long the deck's subdomain/schema can be (the `Tenant.name` field's `max_length`) — and that limit is stated in the field's help text.

Closes #1975.

### Why?

The 20-char cap existed only because the deck name was reused verbatim as `SiteConfig.site_name_short` (`max_length=20`). But as the issue notes, 20 is too short — the name doesn't need to match the short/long display names, it only needs to fit the URL (`<name>.bytedeck.com`). The display fields are already truncated from the name at creation time (`tenant/initialization.py` truncates into `site_name`/`site_name_short`), so a longer name is safe. The help text also didn't mention any character limit, so requesters only found out by hitting a validation error.

### How?

- **`src/tenant/forms.py`**: removed the `MAX_DECK_NAME_LENGTH = 20` cap and its length check in `TenantBaseForm.clean_name`. Length (and format) are already enforced by the model field's `max_length` (62, the Postgres schema/URL limit) and `check_tenant_name` validator, so the extra 20-cap was the only thing holding names down. `clean_name` still rejects the reserved `public` name and orphaned-schema collisions.
- Added a `TenantBaseForm.__init__` that appends the limit to the name field's help text (e.g. "… The name can be at most 62 characters long."), derived from the model field so it stays correct if the field ever changes. No model change → no migration.

### Testing?

- Rewrote `test_clean_name__deck_name_length_capped_at_url_limit`: a name one over the field limit is rejected; a name **longer than the old 20-char cap** but within the field limit is now accepted (would have failed before this change).
- Added `test_name_help_text_includes_length_limit`: the name field's help text contains the limit.

`ruff check src` passes.

The only rendered change is the deck-name field's help text on the deck-creation form gaining a trailing sentence stating the character limit (e.g. "… The name can be at most 62 characters long."). No layout/visual change beyond that added text, so I haven't attached a screenshot — happy to add one if you'd like to see it rendered.

### Anything Else?

`site_name_short` (20) and `site_name` (50) are still truncated from the deck name at creation, which is already covered by the initialization tests — so a long deck name produces a truncated short name rather than an error.

### Review request
@tylerecouture

- Claude Code

---

### DIFF
--- a/src/tenant/forms.py
+++ b/src/tenant/forms.py
@@ -9,14 +9,6 @@ from .models import Tenant
 
 User = get_user_model()
 
-# The deck name is used verbatim as SiteConfig.site_name_short (max_length 20)
-# and also seeds the other name-derived display fields (site_name, the Site
-# name). Capping the name here — at the shortest of those limits — means the
-# user gets a clear error at form time instead of a valid-looking long name
-# being silently truncated when the deck is created. (The creation code still
-# truncates defensively for non-form paths like the admin / management commands.)
-MAX_DECK_NAME_LENGTH = 20
-
 
 class TenantBaseForm(ModelForm):
     """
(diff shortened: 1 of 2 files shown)


tylerecouture (commented):
The original problem was the result of ".bytedeck.com" or ".bytedeck-staging.com" being added to the tenant/dekc name.  So if we have a 62 character name, then add .butedeck.com or .bytedeck-staging.com somewhere, does that cause an error?

Please review the old issue that initiated these changes.

tylerecouture:
Good question — I went back and re-read #1975. Short answer: **no, a 62-char name + the domain suffix does not cause an error.** The suffix is appended *after a dot*, so it becomes separate DNS labels and lands in a separate DB field — it doesn't get concatenated onto the name into one over-long string. Here's the reasoning with the actual numbers:

The two things the name has to fit into, and both apply to the **name alone** (not name + suffix):

1. **Postgres schema name** — derived in `views.py` as `slugify(name).replace("-", "_")[:63]`, stored in `schema_name` (`max_length=63`). A 62-char name → 62-char schema. ✓
2. **DNS label** — the `<name>` in `<name>.bytedeck.com` is a single DNS label, and the per-label limit is 63 octets. 62 < 63. ✓ The `.bytedeck.com` / `.bytedeck-staging.com` parts are *their own* labels, so they don't count against the name's label length.

Where the name **and** suffix live together is the `TenantDomain.domain` field (`domain_url = f"{name_slug}.{current_domain}"`), and there's lots of headroom there:

| suffix | full domain | field limit |
|---|---|---|
| `.bytedeck.com` | 75 chars | 253 |
| `.bytedeck-staging.com` | 83 chars | 253 |

So the worst case (62-char name + the longer staging suffix) is 83 chars against a 253-char field, and the total FQDN limit is also 253 — no overflow either way.

Net: 62 is a safe cap because the binding constraints (Postgres schema = 63, DNS label = 63) apply to the name by itself and 62 sits one under both; the domain suffix never eats into that budget. If you'd prefer, I can add an explicit test asserting that a max-length (62-char) name creates a tenant with a valid schema and a `<name>.<suffix>` domain, to lock this in.

- Claude Code

---

PR #2051, opened 2026-07-21
Fix content list bullets rendering as level-2 circles inside Bootstrap list-groups — Closes #811

### LINKED ISSUE #811: General Info/Common Data list margins
The Common Data /General info sections on the left side of Quest Detail pages is narrow, and the big padding on the left takes up too much space.

Change the padding-left to 15px (from default 40?) for UL and OL items in the General info section

Guess at selector:
`#common-data panel-body ul`

But make sure it's only the first list depth (check how nested lists look)

tylerecouture:
### What?

Fixes list bullets in quest/submission/announcement content rendering as **level-2 hollow circles** instead of **level-1 solid discs** (and squares one level deeper).

Closes #811.

### Why?

The real problem in #811 isn't the left padding (my first pass mis-read it as that) — it's the **bullet style**, as clarified on the issue. Rendered rich-text content (Summernote HTML) is sometimes shown inside a Bootstrap **`.list-group`, which is itself a `<ul>`**. That makes a content list a *nested* `<ul>`, so the browser applies the level-2 marker (`circle`) to its top level and level-3 (`square`) to the next — the content's bullets look like sub-bullets. It shows up in the views that render content inside a `.list-group` (the quest submission/approval previews and the announcements list).

### How?

Per review feedback, the fix is **scoped to a content container**, not to the *absence* of a class:

- Each rendered rich-text block is wrapped in a `<div class="user-content">` at the affected render sites (`preview_content_submissions.html`, `preview_content_approvals.html`). The announcements list already wrapped content in `.announcement-content`, so it just gains the `user-content` class too.
- One scoped block in `src/static/css/custom_common.css` re-establishes depth-correct markers **counting from the content list itself**:

```css
.user-content ul       { list-style-type: disc; }
.user-content ul ul    { list-style-type: circle; }
.user-content ul ul ul { list-style-type: square; }
```

Because the rule only applies *inside* `.user-content` — which by construction holds nothing but the rendered Summernote HTML — it fixes content lists **regardless of whether they carry a class**, and can **never** touch Bootstrap's own component lists (`.list-unstyled`, `.dropdown-menu`, `.nav`), which live outside the wrapper.

> Replaces the earlier `.list-group-item ul:not([class])` selector, which keyed off class-absence and would break if a content list ever gained a class (per @tylerecouture's review).

### Testing?

CSS + template-only. Verified headless (Playwright + Chromium) with the real DOM shape — content inside `ul.list-group > li > .user-content`, a sibling Bootstrap `.list-unstyled`, and Bootstrap's `list-style:none` rule loaded:

- content `ul` → `disc`, `ul ul` → `circle`, `ul ul ul` → `square`
- sibling `.list-unstyled` → `none` (confirming Bootstrap lists are unaffected)

Headless before/after of a content list inside a `.list-group` — hollow level-2 circles become proper solid discs, with nesting still correct (visible result is unchanged from the earlier approach; only the mechanism is now container-scoped):

#811 before/after — content list bullets

### Anything Else?

`<ol>` isn't affected — Bootstrap's `.list-group` is a `<ul>`, so content `<ol>`s keep their decimal numbering.

### Review request
@tylerecouture

- Claude Code


tylerecouture (commented):
"not([class])" seems too easy to break, what if a class is added to the list for some other reason?  Instead, it should be scoped to inside whatever container this appears in (where the problem exists)

tylerecouture:
Good call — keying off the *absence* of a class (`:not([class])`) is fragile exactly as you say. I've reworked it to scope by container instead.

**What changed:**
- Each rendered rich-text block is now wrapped in a `<div class="user-content">` at the render sites where content lands inside a `.list-group` (the quest submission/approval previews). Announcements already had a `.announcement-content` wrapper, so I just added `user-content` to it.
- The CSS is now scoped to that wrapper:
  ```css
  .user-content ul     { list-style-type: disc; }
  .user-content ul ul  { list-style-type: circle; }
  .user-content ul ul ul { list-style-type: square; }
  ```

Because the rule only applies *inside* `.user-content` — which by construction holds nothing but the Summernote HTML — it (a) fixes content lists regardless of whether they carry a class, and (b) can never touch Bootstrap's own component lists (`.list-unstyled`, `.dropdown-menu`, `.nav`, …), which live outside the wrapper. No more reliance on class-absence.

**Verified headless (Playwright + Chromium)** with the real DOM shape — content inside `ul.list-group > li > .user-content`, plus a sibling Bootstrap `.list-unstyled`, and Bootstrap's `list-style:none` rule loaded:

| element | computed `list-style-type` |
|---|---|
| content `ul` (level 1) | `disc` ✓ |
| content `ul ul` (level 2) | `circle` ✓ |
| content `ul ul ul` (level 3) | `square` ✓ |
| sibling `.list-unstyled` | `none` ✓ (untouched) |

The visible result is unchanged from the earlier screenshot (hollow level-2 circles → solid discs); only the mechanism is now container-scoped.

- Claude Code

---

PR #2052, opened 2026-07-21
Target Python 3.12 in the pyupgrade hook; pin local interpreter

tylerecouture:
### What?
Finish the Python 3.12 migration's loose ends: point the `pyupgrade` pre-commit hook at 3.12, and add a `.python-version` pin.

### Why?
`develop` already runs on Python 3.12 everywhere that matters — `Dockerfile` (`FROM python:3.12-slim`), the CI lint job, `pyproject.toml` (`target-version = "py312"`), and the docs. An audit for stray `3.10` references turned up only two real gaps:

1. **`pyupgrade` was still targeting Python 3.6–3.8.** The hook passed `--py36-plus --py37-plus --py38-plus` (and was pinned at `v2.34.0`), so it applied *none* of the 3.9–3.12 modernizations the project could now use.
2. **No `.python-version`.** A contributor using pyenv/uv could silently create a venv on a different interpreter than the 3.12 the container enforces — the exact drift the migration was meant to remove.

### How?
- Bump `pyupgrade` `v2.34.0 → v3.21.2` and set `args: ["--py312-plus"]`. The rev bump is required because `--py312-plus` was only added in pyupgrade v3.15. (The hook runs only on files staged in a commit, so there's no mass rewrite — it just modernizes touched files to 3.12 idioms going forward.)
- Add `.python-version` containing `3.12` so pyenv/uv auto-select the right interpreter for local venvs, matching the Docker base image.

### Testing?
Config-only; no application code. `.pre-commit-config.yaml` validated as YAML and the pyupgrade rev/args parse correctly. Verified via `git ls-remote` that `v3.21.2` is a real pyupgrade tag that supports `--py312-plus`.

For completeness, a full-tree sweep found **no remaining Python-3.10 references** outside `CHANGELOG.md` (historical entries, correctly left untouched). The `django-tenants>=3.10.2` pin in `requirements-production.txt` is a package version, not a Python version.

N/A — tooling only.

### Anything Else?
Related but intentionally **not** in this PR (each its own decision): bumping the other pinned pre-commit hooks (`pre-commit-hooks v4.3.0`, current is v6.0.0) is Dependabot territory; adding `requires-python` to `pyproject.toml` was skipped because the project isn't pip-installed as a package (it installs via `requirements.txt`), so it would have no enforcing effect.

Separately: the "double CI runs" concern is already resolved on `develop` — both `build_and_test.yml` and `lint.yml` now scope the `push` trigger to long-lived branches and use `concurrency: cancel-in-progress`. No change needed there.

### Review request
@tylerecouture

- Claude Code

---

## Summary by CodeRabbit

* **Chores**
  * Updated the project’s supported Python version to 3.12.
  * Updated code quality tooling to align with Python 3.12 syntax and features.


.pre-commit-config.yaml
```diff
@@ -10,10 +10,13 @@ repos:
         exclude: ^requirements/generated.txt
 
   - repo: https://github.com/asottile/pyupgrade
-    rev: v2.34.0
+    rev: v3.21.2
     hooks:
       - id: pyupgrade
-        args: [ "--py36-plus", "--py37-plus", "--py38-plus" ]
+        # Target the project's actual runtime (Python 3.12). The old stacked
```
tylerecouture: These kind of code comments are not helpful.  They are helpful in the PR description, but in the code, they mean nothing because the line they are referring to has been removed (if it was just commented out, that would make sense to refer to it).  So, remove the old line, and remove these comments that no longer have context.

tylerecouture: Good call — removed in 8a693ad. `args: ["--py312-plus"]` speaks for itself, and the rationale for the bump lives in the PR description where it has context. I'll keep this in mind going forward: code comments should explain the current code, not narrate what a removed line used to be.

- Claude Code

---

PR #2057, opened 2026-07-21
CLAUDE.md: sign GitHub posts with the session id

tylerecouture:
### What?
Update the Claude Code sign-off convention in `CLAUDE.md`: every GitHub post (PR description, comment, review comment, issue) should be signed **"- Claude Code (`<session-id>`)"** — with the session id, not just "- Claude Code".

### Why?
When more than one Claude session is active, a bare "- Claude Code" gives no way to tell *which* session wrote a comment, so you can't reply in the right one. Embedding the session id (the same `session_…` value already in the auto-added `claude.ai/code` PR link) lets you reply in the exact session that made the post.

### How?
One-line change to the existing sign-off bullet under "Code Style & PR Conventions". This PR's own description and commit are already signed the new way as a demonstration.

### Testing?
Docs-only.

N/A.

### Anything Else?
This is the doc/convention half of a broader "sessions stay current" thread. The companion question — *how to make long-running sessions actually pick up CLAUDE.md edits like this one instead of holding a stale copy* — needs a harness-level hook (a CLAUDE.md instruction can't refresh itself); I'm handling that separately.

### Review request
@tylerecouture

- Claude Code (session_01BMPt54dgApiXjk6TXqD2fi)

---


CLAUDE.md
```diff
@@ -100,7 +100,7 @@ Celery (with `tenant-schemas-celery` for schema awareness) handles background ta
...
 * Claude Code (web sessions): after pushing a branch with an open PR, subscribe to the PR's activity (`subscribe_pr_activity`) and follow up on CI failures and review comments until the PR is merged or closed.
 * Claude Code: after addressing a PR review comment (pushing the fix and/or answering the question in a reply), mark that review thread as resolved (`resolve_review_thread`) so the PR reflects its true state — don't leave addressed threads dangling for the reviewer to close.
-* Claude Code: sign off everything you post to GitHub (PR descriptions, comments, review comments, issues) with "- Claude Code".
+* Claude Code: sign off everything you post to GitHub (PR descriptions, comments, review comments, issues) with "- Claude Code (`<session-id>`)", where `<session-id>` is this session's id — the `session_…` value (e.g. `session_01BMPt54dgApiXjk6TXqD2fi`), the same id in the `https://claude.ai/code/<session-id>` link auto-added to PR descriptions. Including the session id lets the maintainer reply in the exact session that made the post.
```
tylerecouture: Instead of Session ID, can you use the session name?  human readable and useful, unlike session id.  For example your session name is "bytedeck - integrations and automations"

tylerecouture: Done in 64a9947 — the convention now signs with the human-readable session name (e.g. `bytedeck - integrations and automations`), with the `session_…` id kept only in the auto-added `claude.ai/code` link for machine reference. Agreed the name is far more useful for spotting which session to reply in.

One practical wrinkle worth flagging so this actually works reliably: a session needs to *know* its own name to sign with it, and the name isn't automatically in a session's context — I only know mine because you just told me here (I'd been going off the `session_…` id, which is what does show up). So the convention works when the session knows its name, but a fresh session that hasn't been told its name can't produce it on its own. If Claude Code exposes the session name to the model in your setup, great; if not, the reliable fallback is that a session uses whatever name it's been given (or asks) and always keeps the id-link in PR descriptions as the unambiguous pointer. Happy to adjust the wording if you'd like it to spell out that fallback.

- Claude Code (bytedeck - integrations and automations)

---

Issue #2061, opened 2026-07-21
Badge auto granting tweaks


tylerecouture:
1. This banner should be colored with a bootstrap style. Info blue?

<img width="880" height="398" alt="Image" src="https://github.com/user-attachments/assets/85269fcf-ba0e-4394-96f9-d4ed0438ef3c" />

2. The two messages at the top are redundant:

<img width="658" height="135" alt="Image" src="https://github.com/user-attachments/assets/aaa22a96-4850-47cb-9243-4ef92b2db056" />

3. Make it clear that the scan only apply to Students Currently Enrolled in a Course (is this tested?  Does it actually do this?)

4. The Badge detail page, the buttons are bad, they don't fit and break the line awkwardly.  Should we group them into two groups?  Or don't group them?  Make them look better, be consistent with buttons through out the app.

PR #2066, opened 2026-07-21
Badge grant-qualifying tweaks: info banner, dedupe copy, students-only scope, button layout (#2061)

tylerecouture:
### What?

Addresses the four points in #2061 for the "grant to everyone who qualifies" flow (from #1925/#1157).

1. **Banner colour** — the confirm page's grey `.well` becomes a Bootstrap **`alert alert-info`** (the app's info-blue), matching the messages style used elsewhere.
2. **Redundant messages** — the page had two blocks both restating the count and the grant action. Collapsed to one: the alert lists who qualifies; the form line just notes teachers are notified and that only enrolled students are checked. (Also removed a dead `heading` value the view computed but the template never used.)
3. **"Only currently-enrolled students"** — made explicit in the copy, **and** actually enforced: the scan already restricted to `all_users_for_active_semester()`, but didn't pass `students_only=True`, so a staff/test account enrolled in a course could be swept in. Now both the preview (`Badge.students_who_qualify_ungranted`) and the grant task pass `students_only=True`.
4. **Badge detail buttons** — the crowded 6-button row is split into two labelled groups (**Manage**: edit/copy/delete, **Grant**: one/bulk/all-who-qualify) inside a `btn-toolbar`, so they wrap as tidy units instead of breaking mid-group.

### Why?

Per the issue: the banner should use a Bootstrap style; the two top messages were redundant; it needed to be clear (and true) that the scan only covers students currently enrolled in a course; and the detail-page buttons didn't fit and broke awkwardly.

### How?

- `badge_grant_qualifying_confirm.html`: `.well` → `alert alert-info`; merged the two `<p>` messages; clarified the enrolled-only scope in both the populated and empty states.
- `badges/models.py` + `prerequisites/tasks.py`: `all_users_for_active_semester(students_only=True)` (which filters out `is_staff` and test accounts) in the preview and the grant task.
- `badges/detail.html`: one `.btn-group` of six → a `.btn-toolbar` with two `.btn-group`s.
- `badges/views.py`: dropped the unused `heading` context key.

### Testing?

The enrollment/students-only restriction had **no negative test** before (every existing test used an enrolled non-staff student). Added, and each fails without this change:
- `prerequisites/tests/test_tasks.py`: the grant task skips a qualifying-but-unenrolled user and an enrolled staff member, while still granting the enrolled student.
- `badges/tests/test_models.py`: `Badge.students_who_qualify_ungranted()` excludes the unenrolled user and the enrolled staff member, includes the enrolled student.

`ruff check` passes.

Confirm page — grey `.well` + duplicated message → blue `alert-info` + single message with the enrolled-only note:

confirm before/after

Badge detail action buttons — one crowded group that breaks mid-row → two clean groups (Manage / Grant):

| Before | After |
|---|---|
| before | after |

> The button shots are headless reproductions using the app's Slate theme CSS (not the live app, which was unavailable here). Icons are shown as text labels since Font Awesome loads from a CDN — the point they illustrate is the grouping/wrapping, which is faithful.

### Anything Else?

`students_only=True` changes one edge case: a **staff or test account enrolled as a CourseStudent** is no longer eligible for this bulk grant-check (they never should have been — it's the "grant to qualifying **students**" action). Individual and bulk manual grants are unaffected.

### Review request
@tylerecouture

- Claude Code

---

### DIFF
--- a/src/badges/models.py
+++ b/src/badges/models.py
@@ -267,7 +267,9 @@ class Badge(IsAPrereqMixin, HasPrereqsMixin, TagsModelMixin, models.Model):
         """
         from courses.models import CourseStudent
 
-        students = CourseStudent.objects.all_users_for_active_semester()
+        # students_only=True: the grant-check is for students, so exclude staff/test accounts that
+        # happen to be enrolled in a course (issue #2061). The grant task filters the same way.
+        students = CourseStudent.objects.all_users_for_active_semester(students_only=True)
         return [user for user in students if self.student_qualifies_ungranted(user)]
 
     def get_icon_url(self):
--- a/src/badges/templates/badges/badge_grant_qualifying_confirm.html
+++ b/src/badges/templates/badges/badge_grant_qualifying_confirm.html
@@ -14,19 +14,22 @@
   <p></p>
 
   {% if qualifying_students %}
-    <div class="alert alert-info">
-      <p>
-        <strong>{{ qualifying_students|length }}</strong>
-        student{{ qualifying_students|length|pluralize }} currently enrolled in a course
-        {{ qualifying_students|length|pluralize:"meets,meet" }} this
-        {{ config.custom_name_for_badge|lower }}'s prerequisites but
-        {{ qualifying_students|length|pluralize:"has,have" }} not been granted it yet:
-      </p>
-      <ul>
-        {% for student in qualifying_students %}
-          <li><a href="{{ student.profile.get_absolute_url }}">{{ student.profile }}</a></li>
-        {% endfor %}
-      </ul>
+    <div class="panel panel-info">
+      <div class="panel-heading"><h3 class="panel-title">Students who qualify</h3></div>
+      <div class="panel-body">
+        <p>
+          <strong>{{ qualifying_students|length }}</strong>
+          student{{ qualifying_students|length|pluralize }} currently enrolled in a course
+          {{ qualifying_students|length|pluralize:"meets,meet" }} this
+          {{ config.custom_name_for_badge|lower }}'s prerequisites but
+          {{ qualifying_students|length|pluralize:"has,have" }} not been granted it yet:
+        </p>
+        <ul>
+          {% for student in qualifying_students %}
+            <li><a href="{{ student.profile.get_absolute_url }}">{{ student.profile }}</a></li>
+          {% endfor %}
+        </ul>
+      </div>
     </div>
     <form action="" method="post">{% csrf_token %}
       <p>Granting will also notify each student's teacher. Only students enrolled in a course this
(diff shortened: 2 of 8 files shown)


tylerecouture (commented):
1. Why is the edit button a different height?

2. Reverts the blue well back to how it was before, but add an info heading to the panel

tylerecouture:
Both addressed in `a487cb7`:

**1. Edit button height.** The uneven height came from my splitting the actions into two `btn-group`s inside a `btn-toolbar` — when the groups wrapped, the rows didn't line up. Reverted to a **single `.btn-group`** (Bootstrap keeps every button the same height and connected there, like the Current Semester / All Active toggle just below it), and widened its column to `col-sm-5`. The six icon-only buttons now sit on one row without wrapping/breaking:

buttons after

**2. Grant page well → info panel.** Reverted the flat blue `alert-info` back to a neutral panel body, with an info-styled heading — `panel panel-info` (blue "info" heading over a neutral body) rather than the whole box being blue:

confirm panel

(Both are headless reproductions in the Slate theme — the live app wasn't available in this environment. The button shot uses placeholder glyphs at `fa-fw` width to show the real icon-only layout; the earlier text-label shot was just to illustrate the grouping.)

If "add an info heading to the panel" meant keeping the grey `.well` with a heading rather than converting to `panel panel-info`, say the word and I'll switch it.

---

tylerecouture:
Add an hr separating the Bade content and the list info below it.  Below the heading, make it an H3 for "Students who qualify for this badge"

Then the Cancel/Grant buttons, then the plain well with students (link students to their profiles) then another set of the same buttons at the bottom of the well (for long lists, in case user has scrolled to the bottom). Also, mention how much XP will be granted to students somewhere in the text.

tylerecouture:
Reworked the confirm page to match your layout (pushed in `b841b69`):

1. **`<hr>` separator** between the badge content (icon + description) and the qualifying-students section.
2. **H3 heading** — *"Students who qualify for this badge"* (uses the deck's custom badge name).
3. **Cancel / Grant buttons above the list**, then
4. the **plain `.well`** listing the qualifying students, each **linked to their profile**, then
5. **a second Cancel / Grant set at the bottom** of the well, so on a long list you don't have to scroll back up to grant.
6. **XP is now stated** in the intro line: *"Granting it will award **N XP** to each of them, and their teachers will be notified."*

Both submit buttons post the same form, so either one grants; Cancel returns to the badge.

confirm page v3

> Headless reproduction using the app's Slate theme CSS (live app unavailable here); markup and copy are the real ones.

Also added assertions to the existing grant-qualifying view test locking in the new H3 heading and the XP amount.

---

Issue #2060, opened 2026-07-21
Students can join a course when there is no semester open!


tylerecouture:
All semesters are closed, even though the last one used is still the active semester:

<img width="888" height="423" alt="Image" src="https://github.com/user-attachments/assets/263237ed-9ba6-4a0b-913a-cb1039cbf578" />

But students can still register for it!  This page should say "No semesters are currently open.  You teacher needs to open a semester before you can join a course."  Or something similar.  Check if there are similar messages in the app for other scenarios first and maintain consistancy.

<img width="729" height="419" alt="Image" src="https://github.com/user-attachments/assets/bc97edc2-b5a9-4b55-b3e0-a051ab2ab8f0" />

Also, when a teacher signs in to their deck, and there are no active semesters, they should get a message/banner with a warning that the deck has no open semester, so student will no be able to sign up for a course within their deck.  Then link to the semesters page and advise them to create a new semester and activate it.

tylerecouture:
Both halves of this are already in `develop`, so I think this can close. It was fixed the same day it was filed, in `1448d78c` ("Block course registration when no semester is open; warn teachers (#2060)", PR #2067), and the later semester refactor (#2157 Phases 2 and 3) kept it working when `active_semester` became nullable and a deck could run several semesters at once.

**Student side.** `CourseStudentCreate` refuses to show the form, on GET *and* on POST:

```python
# src/courses/views.py
@staticmethod
def _no_open_semester():
    return SiteConfig.get().has_no_open_semester()
```

`SiteConfig.has_no_open_semester()` is `not Semester.objects.open().exists()`, i.e. deck-wide rather than the `active_semester` pointer, so a deck whose pointer still names a closed semester (exactly your screenshot) counts as having none. `coursestudent_form.html` then renders, in place of the form:

> No semesters are currently open. Your teacher needs to open a semester before you can join a course.

which is the wording you asked for. The "Join a Course" buttons on the profile page and the quests page are swapped for a short note under the same condition.

**Teacher side.** `semester_status_banner.html`, included from `base.html` for staff on every page:

> **No semester is open:** students can't join a course until you create and activate a new semester. [Manage semesters](#)

[...]

PR #2067, opened 2026-07-21
Block course registration when no semester is open; warn teachers (#2060)

tylerecouture:
### What?

Two parts, both from #2060:

1. **Students can no longer join a course when no semester is open.** The student "Join a course" page now shows *"No semesters are currently open. Your teacher needs to open a semester before you can join a course."* instead of the registration form, and creates nothing.
2. **Teachers get a heads-up.** A staff-only warning banner appears (on every page) whenever the deck's active semester is closed, telling them students can't join a course and linking to the semesters page to create and activate one.

### Why?

Closing a semester (`Semester.complete_active_semester`) sets `closed=True` but leaves it as `SiteConfig.active_semester`. Nothing in the join flow checked semester state, so a student could still register — and get attached to a **closed** semester. And a teacher whose semester had lapsed had no signal that their students were locked out.

### How?

- **`courses/views.py`** — `CourseStudentCreate` gains a `_no_open_semester()` check (active semester missing or `closed`); GET and POST short-circuit to a message instead of the form, so nothing is created. Scoped to student self-registration (`courses:create`); staff "add student" (`CourseAddStudent`, a separate class) is untouched.
- **`courses/coursestudent_form.html`** — shows the message when `no_open_semester` is set.
- **`templates/base.html`** — staff-only `alert-warning` banner gated on `config.active_semester.closed`, linking to `courses:semester_list`. Mirrors the existing `config.hs_closed` banner; no view changes (`config` is globally available via context processor).

**Note on "open":** the guard keys off the semester's stored `closed` flag (the state the report describes — "all semesters are closed"). There's an unused date-based `Semester.is_open()`; I kept to `closed` so a future-dated semester a teacher is setting up isn't treated as "not open". Happy to switch to date-based if you'd prefer.

### Testing?

- Registration is blocked when the active semester is closed: GET shows the message and hides the form; POST with valid data creates no `CourseStudent`.
- The banner shows for staff when the semester is closed, is hidden when a semester is open, and is never shown to students.

`ruff check` passes.

The teacher banner and the student join-page message (Slate theme):

no open semester

> Headless reproduction using the app's Slate theme CSS (the live app was unavailable in this environment) — the alert markup and copy are the real ones.

### Review request
@tylerecouture

- Claude Code

---

### DIFF
--- a/src/courses/templates/courses/coursestudent_form.html
+++ b/src/courses/templates/courses/coursestudent_form.html
@@ -5,6 +5,12 @@
 
 {% block content %}
 
+{% if no_open_semester %}
+  <div class="alert alert-warning">
+    <p>No semesters are currently open. Your teacher needs to open a semester before you can join a course.</p>
+  </div>
+{% else %}
+
 {% if submit_btn_value != 'Update' %}
   <p>
   {% if form.instance.user and form.instance.user != user %}
@@ -23,5 +29,7 @@
     </form>
   </div>
 </div>
+
+{% endif %}
 
 {% endblock %}
--- a/src/courses/tests/test_views.py
+++ b/src/courses/tests/test_views.py
@@ -533,6 +533,65 @@ class CourseStudentViewTests(ViewTestUtilsMixin, ByteDeckTenantTestCase):
         response = self.client.get(reverse('courses:create'))
         self.assertNotEqual(response.status_code, 403)
 
+    def _close_active_semester(self):
+        """Close the deck's active semester (leaving it as the active semester), reproducing the
+        'no semester is open' state from issue #2060. Saving fires the SiteConfig cache invalidation.
+        """
+        active = SiteConfig.get().active_semester
+        active.closed = True
+        active.save()
+
+    def test_CourseStudentCreate_view__blocked_when_no_open_semester__get(self):
+        """When the active semester is closed, the join page shows a 'no semesters open' message
+        instead of the registration form (issue #2060)."""
+        self.client.force_login(self.test_student1)
+        self._close_active_semester()
+
+        response = self.client.get(reverse('courses:create'))
+
+        self.assertEqual(response.status_code, 200)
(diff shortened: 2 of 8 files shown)


tylerecouture:
Can you clarify where these messages appear?

For staff it should be a dismissable message (django message framework)

For students, it should show in place of the "join a course" button, and the join a course form (if no open semesters. In case student accesses directly via Url, or some other prevention of direct access to the form (form clean method or otherwise)

tylerecouture:
Reworked this per your review — the static banner is gone, and staff vs. student now get distinct treatments:

**Staff → dismissable Django message.** When a staff member lands on the home page and there's no open semester, they get a `messages.warning(...)` (the standard framework message, with the × dismiss) linking straight to the semesters page:

> ⚠ This deck has no open semester, so students can't join a course. **Create and activate a semester** to let them register.

It's emitted in `home()` (in `hackerspace_online/views.py`) right before the staff redirect to the approvals queue, so it shows on their first landing rather than being a permanent fixture.

**Students → the "Join a Course" button is replaced in place.** Everywhere a student would otherwise see a **Join a Course** link (profile page and the quests "Available" tab), when there's no open semester the button is swapped for an inline note:

> *No semester is open yet, so you can't join a course — ask your teacher to open one.*

**Direct URL access is still blocked.** The join view (`CourseStudentCreate`) guards both `GET` and `POST` — so even if a student navigates straight to the form URL (or POSTs to it), they get bounced with the same message rather than reaching the form.

All three paths key off a single new helper, `SiteConfig.has_no_open_semester()` (true when there's no active semester or it's closed), so the logic lives in one place.

Screenshot of the staff message + the student note:

staff message and student note

One design choice worth flagging: the staff message currently appears on the **home landing** (dismissable, one-shot). If you'd rather it persist site-wide (on every page until a semester is opened), I can move it into middleware instead — let me know which you prefer.

---

PR #2105, opened 2026-07-23
Consolidate bootstrap-table on one vendored version (1.20.2)

tylerecouture:
### What?

Second step of the Bootstrap cleanup series (follows #2095). The LMS was loading **bootstrap-table at three different versions from three sources**:

| Version | Source | Used by |
|---|---|---|
| 1.16.0 | local | global CSS (`head_css.html`), profile list JS |
| 1.18.3 | cdnjs | quest tables (`quest_manager/base.html`) |
| 1.20.2 | unpkg | badge detail, tags list |

This standardizes everything on a **single locally-vendored 1.20.2**, and (per review) loads both the CSS and JS **per-page, only where the widget is used**.

### Why?

- **Consistency + no CDN dependency** (same rationale as #2095 — the cdnjs/unpkg copies don't load in CI/offline dev).
- **1.20.2 is the right target**, not throwaway: it's the newest already in use, and bootstrap-table only gained **Bootstrap 5 support around 1.19** — so this is a **prerequisite** for the eventual BS5 migration.

### How?

- Vendored bootstrap-table **1.20.2** dist over the old local 1.16.0 (`css/bootstrap-table.min.css`, `js/bootstrap-table.min.js`), and added the matching **1.20.2 export extension** (`js/bootstrap-table-export.min.js`), removing the old standalone `js/bootstrap-table-export.js` it supersedes.
- `quest_manager/base.html`, `badges/detail.html`, `tags/list.html` now load the local JS instead of cdnjs/unpkg.
- **Loading is now per-page and consistent**: the bootstrap-table CSS was previously loaded *globally* in `head_css.html` (on every page) while the JS was per-page. Since the CSS only styles the widget's JS-generated markup (`.bootstrap-table` / `.fixed-table-*`), it did nothing on pages without the widget. Moved it to the four templates that load the JS (`quest_manager/base`, `profile_list`, `badges/detail`, `tags/list`) so CSS + JS are co-located, and dropped it from the global `head_css`.
- `profile_list.html` points at the vendored export extension (and a stale commented-out 1.8.1 CDN line was removed).

JS is loaded per-page (only where tables exist) rather than globally, to avoid shipping ~130 KB on table-less pages — and now the CSS matches.

**Pages covered by `quest_manager/base`:** the Library (`library/library_overview.html`) **extends `quest_manager/base.html`**, so its Quests and Campaigns tabs inherit *both* the table JS (as they always have) and — via this PR — the CSS added to that template's `{% block head %}`. So Library is fully covered by the per-page move. (`portfolios/list`, `portfolios/edit` carry a `data-toggle="table"` attribute but extend `base_sidebar` and never load the table JS, so those render as plain HTML tables — the widget CSS never applied to them, and this change leaves them unaffected.)

### Testing?

- `python src/manage.py check` — clean; `ruff check src` — clean; `badges` + `tags` + `profile_manager` view tests (117) — OK. Full suite in CI.
- **Functionally verified on a running `hackerspace` deck** (the version bump touches the custom quest-table accordion and the profile export, so this was validated, not just rendered):
  - **Quest tables**: 168 rows loaded; clicking a row expands the accordion detail-view (see screenshot).
  - **Profile list**: renders with the export control present.
  - **Badge detail / Tags list**: tables render correctly.
  - **Library Quests + Campaigns tabs** (shared library enabled): both initialize the widget (`.fixed-table-container` present) and render styled — confirming the per-page CSS covers Library via `quest_manager/base` inheritance.
  - After moving the CSS per-page: re-verified each widget page still initializes and renders styled; **no console/page errors** anywhere.

No visual change is intended (it's a version/loading consolidation) — these confirm the table features still work under the single vendored 1.20.2:

**Quest-table accordion expands** (the custom `bootstrap-table-accordion.js` integration):

quest accordion

**Profile list + export control:**

profile export

**Library Quests tab** (extends `quest_manager/base`; widget styled via the per-page CSS):

library quests

(Dark panel on the right is the dev-only django-debug-toolbar.)

### Anything Else?

Remaining in the series: PR3 dedupe the `select2-bootstrap` theme (local vs cdnjs), PR4 delete dead files (`bootstrap_theme_examples.html`, `bootstrap-theme.min.css`, the dead `bootstrap-table-filter-control.js`). Then the LMS BS3 → BS5 migration. Public tenant stays on Bootstrap 5 (intended).

### Review request
@tylerecouture


src/badges/templates/badges/detail.html
```diff
@@ -1,10 +1,11 @@
 {% extends "badges/base.html" %}
+{% load static %}
 {% load crispy_forms_tags %}
 {% load utility_tags %}
 
 {% block head %}
-<link href="https://unpkg.com/bootstrap-table@1.20.2/dist/bootstrap-table.min.css" rel="stylesheet">
-<script src="https://unpkg.com/bootstrap-table@1.20.2/dist/bootstrap-table.min.js"></script>
+{# bootstrap-table CSS is loaded globally in head_css.html; only the JS is needed here #}
```
tylerecouture: Why have this comment here and not on other pages that load the table.js?

Also, why load the table css globally? Why no only in the bs table pages?

tylerecouture: Both good points — fixed in 9a321c52.

You're right that it was inconsistent. The CSS was already global in `head_css.html` before this PR and I left it that way, but there's no good reason for it: the bootstrap-table CSS only styles the widget's JS-generated markup (`.bootstrap-table` / `.fixed-table-*`), so it does nothing on pages without the widget. I've moved it to the four pages that load the table JS (`quest_manager/base`, `profile_list`, `badges/detail`, `tags/list`) so CSS and JS are co-located, dropped it from the global `head_css`, and removed the explanatory comments (no longer needed now that it's consistent).

While checking this I found that `portfolios/list`, `portfolios/edit`, and `library/tab_library_quests` carry a `data-toggle="table"` attribute but **never load the bootstrap-table JS** (no base template loads it either), so those tables render as plain HTML tables — the widget CSS was never applying to them, and this change leaves them unaffected. (Whether those were meant to be real widgets is a separate, pre-existing question — happy to file it if you'd like.)

Verified on a running deck: the quest/profile/badge/tags tables still initialize and render correctly with the per-page CSS.

---

tylerecouture:
Before continuing, can you confirm that the library tab doesn't use it?  I do see bs table live on https://hackerspace.bytedeck.com/library/quests-list/  but you say "library/tab_library_quests" doesn't use it.

Both the Quests and Campaign tabs use it:

image

tylerecouture:
You're right, and my earlier note was wrong — the Library tabs **do** use bootstrap-table. I missed that `library/library_overview.html` **extends `quest_manager/base.html`**, so it inherits that template's table JS (which is why it's always worked) — my grep only looked for a direct `<script>` in the library templates and missed the inheritance. Apologies for the incorrect claim.

The good news is that this PR **doesn't break it**: I added the bootstrap-table CSS to `quest_manager/base.html`'s `{% block head %}` (alongside the JS in its `{% block js %}`), and since `library_overview` doesn't override those blocks, the Library page inherits **both** the CSS and JS. So it keeps working — the per-page move happens to cover Library correctly through the same inheritance.

I verified it on a running deck (shared library enabled) on this branch — both the **Quests** and **Campaigns** tabs initialize the widget (`.fixed-table-container` present) and render styled, with no console errors:

library quests tab

To be thorough I re-checked the others I'd lumped in: **`portfolios/list` and `portfolios/edit`** really don't load the table JS (they extend `portfolios/base` → `base_sidebar`, not `quest_manager/base`), so those tables *do* render as plain HTML tables — the widget never initializes there (`.fixed-table-container` = 0), confirmed on the running deck. So the only page I got wrong was Library, and it's fine. I've corrected the note in the PR description.

---

PR #2106, opened 2026-07-23
Warn before leaving a form with unsaved changes (#192)

### LINKED ISSUE #192: Unsaved changes protection when editing a quest or announcement etc.
Should have a confirmation before leaving unsaved changes.

tylerecouture:
### What?

Closes #192. If you edit a content form (a quest or an announcement) and then try to leave the page — close the tab, hit back/refresh, or follow a link — without saving, the browser now asks you to confirm first, so you don't lose unsaved work. Submitting the form saves normally and never warns.

### Why?

These edit forms are long (Summernote WYSIWYG bodies, prerequisites, tags…). It's easy to click away or hit back and silently lose everything you typed. #192 asks for *"a confirmation before leaving unsaved changes."*

### How?

- **`static/js/warn-unsaved-changes.js`** — a small reusable guard. Opt a form in by adding **`data-warn-unsaved`** to its `<form>` tag. The script marks the form "dirty" on any field edit — native inputs/selects/checkboxes (`input`/`change`) **and** Summernote WYSIWYG edits (its bubbling `summernote.change` event) — and, while dirty, cancels `beforeunload` so the browser shows its native *"Leave site? Changes you made may not be saved."* dialog. A `submit` stands the guard down, because saving is intentional.
- Loaded globally (in `templates/base.html`), and the **quest** create/edit form (`quest_form.html`) and **announcement** create/edit form (`announcements/form.html`) opt in via `data-warn-unsaved`. Because it's opt-in, any other content-edit form (badges, campaigns, …) can adopt it later by adding the one attribute.

The confirmation is the **browser's own** dialog (the modern, cross-browser way to guard tab-close/refresh/back as well as in-app link clicks); browsers ignore custom message text and show their own.

### Testing?

- `quest_manager/tests/test_views.py` — the quest **create and edit** pages include `data-warn-unsaved` on the form and load the guard script (`test_quest_form__has_unsaved_changes_guard`).
- `announcements/tests/test_views.py` — same for the announcement **create and edit** pages (`test_announcement_form__has_unsaved_changes_guard`).

Both fail without the template wiring. There's no server-side logic to cover (pure front-end behaviour).

**Behavioural verification** (Playwright against the running `hackerspace` deck, dispatching a real `beforeunload` on the quest edit form and checking whether the guard cancels it):

| State | Leave confirmation shown? |
| --- | --- |
| Form untouched | **No** |
| After editing a field | **Yes** |
| After editing, then submitting | **No** |

The confirmation itself is the browser-native "Leave site?" dialog (browser chrome — it can't be rendered/screenshotted in a headless capture, hence the behavioural table above). For context, here's the **Update Quest** form the guard now protects, from the real running app (seeded `hackerspace` deck, dark theme):

quest edit form

### Anything Else?

- Opt-in by design: only forms tagged `data-warn-unsaved` are guarded, so search/filter forms and other quick forms are unaffected.
- No new dependencies — plain JS using the already-loaded jQuery only to hook Summernote's change event.

### Review request
@tylerecouture

---

### DIFF
--- a/src/announcements/templates/announcements/form.html
+++ b/src/announcements/templates/announcements/form.html
@@ -11,7 +11,7 @@
 {% block heading_inner %} {{ heading }}{% endblock %}
 
 {% block content %}
-<form action="" enctype="multipart/form-data" method="post">{% csrf_token %}
+<form action="" enctype="multipart/form-data" method="post" data-warn-unsaved>{% csrf_token %}
   <div class="announcement-action-btns">
     <a href="{% url 'announcements:list' %}" role="button" class="btn btn-danger">Cancel</a>
     <input type="submit" value="{{ submit_btn_value }}" class="btn btn-success" />
--- a/src/announcements/tests/test_views.py
+++ b/src/announcements/tests/test_views.py
@@ -36,6 +36,15 @@ class AnnouncementViewTests(ViewTestUtilsMixin, ByteDeckTenantTestCase):
         """Set up a tenant test client for each test."""
         self.client = TenantClient(self.tenant)
 
+    def test_announcement_form__has_unsaved_changes_guard(self):
+        """The announcement create and edit forms opt into the unsaved-changes warning, and the guard script is loaded (#192)."""
+        self.client.force_login(self.test_teacher)
+
+        for url in (reverse('announcements:create'), reverse('announcements:update', args=[self.ann_pk])):
+            response = self.client.get(url)
+            self.assertContains(response, 'data-warn-unsaved')
+            self.assertContains(response, 'warn-unsaved-changes.js')
+
     def test_all_announcement_page_status_codes__anonymous(self):
         ''' If not logged in then all views should redirect to login page '''
 
--- a/src/quest_manager/templates/quest_manager/quest_form.html
+++ b/src/quest_manager/templates/quest_manager/quest_form.html
@@ -11,7 +11,7 @@
 {% block heading_inner %} {{ heading }}{% endblock %}
 
 {% block content %}
-<form action="" enctype="multipart/form-data" method="post"> {% csrf_token %}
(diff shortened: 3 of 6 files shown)


tylerecouture (commented):
How does this affect student submissions, since they can save a draft submission comment

tylerecouture:
Good question — short answer: **student submissions are unaffected.**

The guard is strictly opt-in: `warn-unsaved-changes.js` only attaches to a `<form data-warn-unsaved>`, and the only two forms carrying that attribute are the **quest** create/edit form (`quest_form.html`) and the **announcement** create/edit form (`announcements/form.html`). The student **submission** form (`submission.html`) does **not** have the attribute, so its behaviour is exactly as before — including its existing "Save Draft" button and the 60‑second autosave.

I deliberately left the submission form out, and it would actually be **wrong** to add it there as‑is, because that form saves drafts over **AJAX** rather than a real form `submit`:

- The guard only stands down on a `submit` event. "Save Draft" (and the 60s autosave) fire an AJAX POST, **not** a submit — so the form's `dirty` flag would never clear after a draft was safely saved.
- Result: a student who typed a comment, let it autosave, and then navigated away would get a "you have unsaved changes" warning **even though the draft is saved** — a false alarm on the one form that already protects against loss.

So for the submission form the autosave already covers the "don't lose my work" case, and bolting the browser dialog on top would just nag. If you'd like leave‑protection there too, I can wire it up to *cooperate* with the autosave (clear the dirty flag on a successful draft save / mark it dirty again on further edits) — that's a small, deliberate follow‑up rather than something this PR should silently change.

Separately: I resolved the merge conflict with `develop` (two other PRs had since added tests to `quest_manager`/`announcements` `test_views.py`) by rebasing onto the latest `develop` and keeping both sets of tests. Force‑pushed as `db9e290`; the three affected tests pass locally.

---

Issue #1343, opened 2023-04-21
Quest quick reply comment doesn't escape HTML and runs scripts


tylerecouture:
HTML should be escaped completely from "quick reply form".  Currently you can enter comments in the 
![Screenshot-16](https://user-images.githubusercontent.com/10604391/233675822-b6d841aa-3d76-4f62-956a-c586400c34c2.png)

Issue #2113, opened 2026-07-23
Comments are getting wrongly escaped.


tylerecouture:
Why are comments getting their html stripped/escaped?

<img width="1310" height="1000" alt="Image" src="https://github.com/user-attachments/assets/02f04ee0-fce1-4c23-a8bf-4fdba24a952d" />

<img width="1411" height="885" alt="Image" src="https://github.com/user-attachments/assets/bf9a1b2b-b485-4998-bf2a-14aa13848701" />

PR #2123, opened 2026-07-23
Sanitize comment HTML instead of escaping it all (#2113)

tylerecouture:
## What

Comments were showing their **literal HTML tags** as text (e.g. a comment displaying `<p>giggity</p>` instead of the paragraph) — issue #2113. This fixes the regression by **sanitizing** comment HTML with an allow-list instead of escaping all of it, so legitimate formatting renders while scripts are still stripped.

## Why

The stored-XSS fix for #1343 added a blanket `escape()` on the `comment_text` field of the plain-text / quick-reply comment forms. In practice those fields don't receive plain text — they carry **rich HTML from the Summernote editor** (a student's submission comment posts `<p>…</p>`, `<b>`, lists, links, etc.). Escaping all of it turned that legitimate formatting into literal text on the page.

The forms are still rendered with `|safe`, so we can't simply stop sanitizing — that would reopen the #1343 XSS hole. The correct fix is an **allow-list sanitizer**: keep safe formatting, strip anything that can execute.

## How

- New `comments/sanitize.py` → `sanitize_comment_html()`. It runs `bleach.clean()` with the **same tag/style allow-list as the "safe" Summernote widget** (`bytedeck_summernote.settings.ALLOWED_TAGS` / `STYLES`), but with a **stricter attribute policy** that drops inline event handlers (`on*`) and a small deny-list (`srcdoc`, `formaction`, …). Result:
  - kept: `<p>`, `<b>`, `<a href="http…">`, `<ul>/<li>`, styles, etc. → comments render
  - stripped: `<script>` (escaped), `onerror`/`onclick`/… handlers, `javascript:` URLs → nothing can execute
  - This is actually **stricter than the current Summernote box**, which keeps `on*` attributes.
- `comments/forms.py` (`CommentForm`, non-wysiwyg branch) and `quest_manager/forms.py` (renamed `EscapeCommentTextMixin` → `SanitizeCommentTextMixin`, used by the reply / quick-reply forms) now call `sanitize_comment_html()` instead of `escape()`.

Sanitizing at the form level covers every save path, including the student draft path that writes `Comment.text` directly (bypassing `clean_html()`).

## Testing

- The #1343 regression tests (in `comments`, `quest_manager`, `announcements`) are updated to assert **sanitization** — `<script>` neutralized, `onerror` stripped — rather than full escaping.
- New tests assert **legitimate formatting survives** (`<b>bold</b>`, links preserved; not escaped) across all the reply/comment forms and through the full submission-complete save path.
- `python manage.py test comments quest_manager.tests.test_forms quest_manager.tests.test_views.SubmissionCompleteViewTest announcements.tests.test_views.AnnouncementViewTests` — all pass. `ruff` clean.

## Screenshots

Captured from a real `hackerspace` dev deck. Same comment content (a paragraph with bold text, a link, and a bulleted list) on a real quest submission, rendered by the real template:

**Before** (the bug — tags shown as literal text):

Before: comment shows literal HTML tags

**After** (this PR — formatting renders):

After: comment renders formatting

---

## Summary by CodeRabbit

* **Security**
  * Improved comment and reply protection by removing scripts, JavaScript URLs, and inline event handlers.
* **Formatting**
  * Preserved safe HTML formatting, including bold text, paragraphs, and links, in comments and replies.
* **Bug Fixes**
  * Updated comment rendering and submission flows to safely support legitimate rich-text formatting without exposing users to malicious content.

### DIFF
--- a/src/announcements/tests/test_views.py
+++ b/src/announcements/tests/test_views.py
@@ -214,10 +214,10 @@ class AnnouncementViewTests(ViewTestUtilsMixin, ByteDeckTenantTestCase):
         self.assertRedirects(response, self.test_announcement.get_absolute_url())
         self.assertTrue(Comment.objects.all_with_target_object(self.test_announcement).exists())
 
-    def test_comment_on_announcement__escapes_html(self):
-        """HTML entered in the announcement comment form is completely escaped when
-        the comment is saved, so scripts can't execute when the comment is rendered.
-        Regression test for issue #1343.
+    def test_comment_on_announcement__sanitizes_html(self):
+        """HTML entered in the announcement comment form is sanitized when the comment
+        is saved: scripts and event handlers are stripped so nothing executes when the
+        comment is rendered. Regression test for issue #1343.
         """
         self.client.force_login(self.test_student1)
 
@@ -230,9 +230,29 @@ class AnnouncementViewTests(ViewTestUtilsMixin, ByteDeckTenantTestCase):
 
         comment = Comment.objects.all_with_target_object(self.test_announcement).first()
         self.assertIsNotNone(comment)
+        # the <script> tag is neutralized (escaped, not a live tag) and the onerror
+        # event handler is stripped, so no JavaScript can run
         self.assertNotIn('<script', comment.text)
-        self.assertNotIn('<img', comment.text)
         self.assertIn('&lt;script&gt;', comment.text)
+        self.assertNotIn('onerror', comment.text)
+
+    def test_comment_on_announcement__preserves_formatting(self):
+        """Legitimate formatting HTML entered in an announcement comment survives so it
+        renders instead of showing its literal tags. Regression test for issue #2113.
+        """
+        self.client.force_login(self.test_student1)
+
+        payload = '<p>hello <b>bold</b> <a href="http://example.com">link</a></p>'
+        response = self.client.post(
+            reverse('announcements:comment', args=[self.test_announcement.id]),
+            data={'comment_text': payload, 'comment_button': True}
+        )
+        self.assertRedirects(response, self.test_announcement.get_absolute_url())
+
+        comment = Comment.objects.all_with_target_object(self.test_announcement).first()
+        self.assertIsNotNone(comment)
+        self.assertIn('<b>bold</b>', comment.text)
+        self.assertIn('href="http://example.com"', comment.text)
 
     def test_copy_announcement__creates_new_announcement(self):
         """Copying an announcement via POST creates a new announcement and redirects to it."""
--- a/src/comments/forms.py
+++ b/src/comments/forms.py
@@ -1,7 +1,7 @@
 from django import forms
-from django.utils.html import escape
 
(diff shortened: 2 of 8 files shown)


tylerecouture:
@claude I would like to know why this broke.  There were tests that prevented this from happening, was there a PR that modifies the tests and broke this behaviour? Please investigate and report back.

tylerecouture:
Investigated. Here's the root cause and the answer to your test question.

## What broke it

Commit [`4ef218e`](https://github.com/bytedeck/bytedeck/commit/4ef218edb4ca5f696b8a160cda749d7a84b9a4d0) — **"Escape all HTML in plain-text comment forms to prevent stored XSS; Closes #1343 (#1921)"** (2026-07-13).

That fix added a form-level `escape()` on `comment_text` in the plain-text / quick-reply comment forms (`CommentForm`, `SubmissionQuickReplyForm`, `SubmissionQuickReplyFormStudent`, `SubmissionReplyForm`). The intent was sound — `clean_html()` only stripped `<script>` but kept things like ``, so those fields were a stored-XSS hole. But those forms actually receive **rich HTML from the Summernote editor** (a student's submission comment posts `<p>…</p>`), so escaping *everything* turned legitimate formatting into literal text — the `<p>giggity</p>` you saw.

## Was a protective test modified/removed?

**No.** I diffed the entire #1921 commit against the test suite. It did **not** modify or delete any existing test — it only **added** tests, and those tests asserted the *new* escaping behavior:
- `comments/tests/test_forms.py` → `test_comment_form__escapes_html_when_not_wysiwyg` (new file)
- `quest_manager/tests/test_forms.py` → `QuickReplyFormsEscapeHTMLTest` (3 tests asserting `escaped_payload`)
- `quest_manager/tests/test_views.py` → `test_complete_quick_reply_form__escapes_html`
- `announcements/tests/test_views.py` → `test_comment_on_announcement__escapes_html`

So the tests were written to match the code change — they locked the regression in rather than catching it.

## Why nothing caught it

There was **never a test asserting that legitimate HTML typed into a comment renders (survives).** The one pre-existing test touching this path, `SubmissionCompleteViewTest.test_complete__quick_reply_form`, posts the **plain string** `"test submission comment"` and asserts the stored value is `<p>test submission comment</p>`. Since `escape()` leaves plain text unchanged, that test kept passing before *and* after #1921 — it had no HTML in its input, so it couldn't detect the over-escaping. The coverage gap was a missing "formatting survives" assertion.

## What this PR does about the gap

Besides fixing the behavior (allow-list sanitize instead of blanket escape), it **adds the missing coverage** so this can't silently regress again:
- `test_comment_form__preserves_formatting_when_not_wysiwyg`, `test_reply_forms__preserve_legitimate_formatting`, `test_complete_quick_reply_form__preserves_formatting`, `test_comment_on_announcement__preserves_formatting` — all assert `<b>`, links, `<p>` **survive** (these would fail under the old `escape()`).
- The existing XSS tests are kept (now asserting `<script>` neutralized + `onerror` stripped), so the #1343 protection is still enforced — in fact more strictly than before (the old code kept `onerror` inside the Summernote path; the new sanitizer strips it).

---

PR #2129, opened 2026-07-23
Serve the select2-bootstrap theme locally, not from cdnjs

tylerecouture:
### What?

Third step of the Bootstrap cleanup series (follows #2095, #2105). Six form templates loaded the **select2-bootstrap-theme** CSS from `cdnjs` (`0.1.0-beta.10`), while a **byte-identical copy already sat unused** at `src/static/css/select2-bootstrap.min.css`. This points them at the local file and drops the cdnjs `<link>`.

### Why?

- **Consistency + no CDN dependency** (same rationale as the earlier cleanup PRs — the cdnjs copy doesn't load in CI/offline dev).
- The local file is **already there and already the right version** — the cdnjs `<link>` was a redundant external dependency on a file we ship.

### How?

- Repointed the theme `<link>` in all six templates to `{% static 'css/select2-bootstrap.min.css' %}`:
  - `badges/assertion_form.html`
  - `quest_manager/submission.html`
  - `quest_manager/quest_form.html`
  - `prerequisites/advanced_prereqs_form.html`
  - `djcytoscape/generate_new_form.html`
  - `djcytoscape/cytoscape_form.html`
- Added `{% load static %}` to the three that didn't already have it (`submission`, `generate_new_form`, `cytoscape_form`).
- **No change to the CSS file itself** — I confirmed the existing local `select2-bootstrap.min.css` is **byte-for-byte identical** to the cdnjs `0.1.0-beta.10` build (both 16,792 bytes, `diff` clean), so there's no version or visual change.

### Testing?

- `python src/manage.py check` — clean; `ruff check src` — clean; `badges` + `djcytoscape` view tests (110) — OK. Full suite in CI.
- **Verified on a running `hackerspace` deck**: the quest create form's select2 fields (Campaign, Common Quest Info, Tags) render with the bootstrap theme — the Tags dropdown opens with the themed blue-highlighted option — the local `select2-bootstrap.min.css` loads (HTTP 200), and there are **no console errors**.

No visual change (byte-identical file) — this confirms the local theme renders the select2 dropdowns correctly:

quest form select2

(Dark panel on the right is the dev-only django-debug-toolbar.)

### Anything Else?

Last in the cleanup series: **PR4** deletes the dead files (`bootstrap_theme_examples.html`, `bootstrap-theme.min.css`, the dead `bootstrap-table-filter-control.js`). After that, the LMS BS3 → BS5 migration. Public tenant stays on Bootstrap 5 (intended).

### Review request
@tylerecouture

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated form and submission pages to load Select2 styling from local application assets.
  * Improved reliability and consistency of page styling when external CDN resources are unavailable.

### DIFF
--- a/src/badges/templates/badges/assertion_form.html
+++ b/src/badges/templates/badges/assertion_form.html
@@ -4,7 +4,7 @@
 
 {% block head %}
     {{ form.media.css }}
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/select2-bootstrap-theme/0.1.0-beta.10/select2-bootstrap.min.css">
+    <link rel="stylesheet" href="{% static 'css/select2-bootstrap.min.css' %}">
 {% endblock %}
 
 {% block heading_inner %} {{ heading }}{% endblock %}
--- a/src/djcytoscape/templates/djcytoscape/cytoscape_form.html
+++ b/src/djcytoscape/templates/djcytoscape/cytoscape_form.html
@@ -1,10 +1,11 @@
 {% extends "djcytoscape/base.html" %}
+{% load static %}
 {% load crispy_forms_tags %}
 {% block heading_inner %}Update Quest Map{% endblock%}
 
 {% block head %}
     {{ form.media.css }}
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/select2-bootstrap-theme/0.1.0-beta.10/select2-bootstrap.min.css">
+    <link rel="stylesheet" href="{% static 'css/select2-bootstrap.min.css' %}">
 {% endblock %}
 
 {% block content %}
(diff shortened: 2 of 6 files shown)


tylerecouture:
Should we update the version of select2-bootstrap while we are in here?

tylerecouture:
Good instinct, but we're already on the newest one for Bootstrap 3 — so there's nothing to bump here.

`select2-bootstrap-theme` (the theme this app uses) tops out at **`0.1.0-beta.10`**, which is exactly the version we vendor. That package only ever shipped betas (`beta.4` → `beta.10`) and was never taken to a stable release, so `0.1.0-beta.10` is its final version. It targets select2 4.0.x, which is what `django-select2` (8.4.8) bundles, so it's the right match for our current setup.

The only genuinely "newer" select2 themes target **different** Bootstrap majors:
- `select2-bootstrap4-theme` (1.0.0) — Bootstrap 4
- `select2-bootstrap-5-theme` (1.3.0) — Bootstrap 5

Those aren't a like-for-like bump; they're a Bootstrap-version change. So the right time to swap this is the **LMS BS3 → BS5 migration**, where we'd move to `select2-bootstrap-5-theme` (1.3.0) alongside the rest of the BS5 work. I'll add that to the migration plan. For this PR (which is just "stop pulling the file from a CDN"), keeping `0.1.0-beta.10` is correct.

---

PR #2131, opened 2026-07-23
Cache-bust map scripts so the campaign name-wrap fix reaches users (#1289)

### LINKED ISSUE #1289: Prevent line breaks in Campaign name in Quest maps
image

tylerecouture:
### What?

Closes #1289. Long campaign names still appear to wrap onto stacked lines down the side of a campaign column on the quest map — even though #1937 already fixed this. The fix is correct and on `develop`; it just **wasn't reaching browsers**. This PR delivers it by cache-busting the map scripts.

### Why? (root cause)

#1937 fixed the wrapping by setting `text-wrap: none` on the compound campaign label in `maps.js`. But `maps.js` is served with a **plain, unhashed filename and no `?v=` query**:

```
<script src="{% static 'djcytoscape/js/maps.js' %}"></script>
```

So a browser that had already cached the pre-#1937 `maps.js` keeps serving the **old** file — which still has `text-wrap: wrap` and breaks long campaign names onto stacked lines. That's why #1289 looked unsolved even after #1937 merged: the deployed code is right, but the cached JS is stale. (The project already cache-busts other changed assets this way — `jquery-double-submission.js?v=1.1`, `custom_common.css?v=1.7` — `maps.js` was just missing it.)

I verified there is **no remaining rendering bug** on current `develop`: the only campaign node that appears on a real map is the compound parent (`classes="campaign"`), and its label already renders single-line with `text-wrap: none`. (A campaign can't be a downstream/reliant node — `Category` is `IsAPrereqMixin` but not `HasPrereqsMixin` — so no other campaign-name node type exists to wrap.) The wrapping only reproduces with the **old** `maps.js`, i.e. from cache.

### How?

- **`djcytoscape/templates/djcytoscape/quest_map.html`** — add a `?v=1.1` cache-buster to the `maps.js` and `maps-dark.js` includes (with a `{% comment %}` noting to bump it whenever those files change), matching the existing `?v=` convention for changed JS/CSS. A browser holding the old `maps.js` now fetches the new one, which carries the #1937 fix.

### Testing?

- `djcytoscape/tests/test_views.py::ViewTests::test_quest_map__map_scripts_are_cache_busted` — asserts the map page serves `maps.js?v=` (and `maps-dark.js?v=` for dark-theme users). Fails before this change, passes after. Full `djcytoscape` view suite green; `ruff` clean.

Same real map on the running `hackerspace` deck (dark theme), campaign renamed to a long multi-word name to force the break. **Before** = what a browser with the stale cached `maps.js` renders (old `text-wrap: wrap`); **After** = current / cache-busted `maps.js` (`text-wrap: none`).

| Before (stale cached maps.js — wraps) | After (cache-busted — single line) |
| --- | --- |
| before | after |

### Anything Else?

- If you (or a user) still see wrapping after this ships, a one-time hard refresh (Ctrl+Shift+R) on the map page clears the stale `maps.js`; the `?v=` bump makes that automatic going forward.
- Longer term, hashed static filenames (a manifest static storage) would cache-bust everything automatically and remove the need to bump `?v=` by hand — out of scope here, flagging as a follow-up option.

### Review request
@tylerecouture — heads up, this turned out to be a caching-delivery issue rather than a new CSS bug; the before/after above shows the stale-cache wrapping vs. the fixed render. If your screenshot was something other than a compound campaign column wrapping, point me at the specific campaign and I'll dig further.

---

### DIFF
--- a/src/djcytoscape/templates/djcytoscape/quest_map.html
+++ b/src/djcytoscape/templates/djcytoscape/quest_map.html
@@ -85,10 +85,17 @@
 
 </script>
 
-<script src="{% static 'djcytoscape/js/maps.js' %}"></script>
+{% comment %}
+  Cache-bust the map scripts with a ?v= query (same convention as the versioned CSS in
+  head_css.html and other JS like jquery-double-submission.js). maps.js has no hashed
+  filename, so without this a browser keeps serving a stale copy — which is why the campaign
+  name-wrap fix (#1289 / PR #1937) never reached users who had already cached the old maps.js.
+  Bump this version whenever maps.js / maps-dark.js changes.
+{% endcomment %}
+<script src="{% static 'djcytoscape/js/maps.js' %}?v=1.1"></script>
 
 {% if request.user.profile.dark_theme %}
- <script src="{% static 'djcytoscape/js/maps-dark.js' %}"></script>
+ <script src="{% static 'djcytoscape/js/maps-dark.js' %}?v=1.1"></script>
 {% endif %}
 
 
--- a/src/djcytoscape/tests/test_views.py
+++ b/src/djcytoscape/tests/test_views.py
@@ -107,6 +107,19 @@ class ViewTests(ViewTestUtilsMixin, ByteDeckTenantTestCase):
         # self.assert200('djcytoscape:regenerate', args=[self.map.id])
         # self.assert200('djcytoscape:regenerate_all')
 
+    def test_quest_map__map_scripts_are_cache_busted(self):
+        """maps.js (and maps-dark.js) are served with a ?v= cache-buster so browsers don't keep
+        serving a stale copy that misses map fixes like the campaign name-wrap fix (#1289 / #1937)."""
+        self.client.force_login(self.test_student1)
+        response = self.client.get(reverse('djcytoscape:quest_map', args=[self.map.id]))
+        self.assertContains(response, 'js/maps.js?v=')
+
+        # dark-theme users load an extra script that must be cache-busted too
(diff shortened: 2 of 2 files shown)


tylerecouture:
In your after shot, notice how the text is not centered on the campaign.  Is there a way to center the text on that campaign outline?

tylerecouture:
Good eye — but that off-center look is mostly an artifact of my screenshot, not the real render. To force the wrap for the before/after I'd temporarily renamed that campaign to an oversized name ("Digital Hackerspace Orientation") that's far longer than its 4-quest box, so the label hangs past the bottom.

With the **real** name it sits basically centered on the outline (same campaign, name restored to "Orientation (12 XP)"):

realistic name centered

The reason I can't make it *pixel-perfect* for every campaign: the label is a caption rotated −90° and anchored to the campaign's left edge with a **fixed** vertical offset (`text-margin-y`), while campaigns vary in both height (number of quests) and name length. A single fixed offset centers the common cases well (tall campaigns fill their column; normal names sit centered like above), but a short box with an unusually long name will always overhang somewhat. Truly size-aware centering would need a computed per-node margin (label length vs. box height) rather than one constant — a bigger change than this cache-bust PR.

So two options:
1. **Leave positioning as-is** here (it's centered for real names; this PR's job was to actually deliver the #1937 wrap fix), and I open a separate small issue for size-aware label centering if you want to pursue it.
2. I **nudge the constant** a bit in this PR so short campaigns center a touch higher on average — quick, but still a fixed compromise, not perfect for the extreme case.

Which do you prefer? Happy to do either.

---

Issue #2044, opened 2026-07-21
Retire unpaid, unused decks: reminder sequence, then removal or recoverable archival


tylerecouture:
Part of #1729.

## Problem

Abandoned decks (never subscribed, trial long expired, no staff activity) accumulate forever and are a real resource drain: every Postgres schema multiplies `migrate_schemas` time on each deploy, bloats backups, and adds rows to every all-tenants fan-out task. Today nothing removes them — and the admin delete deliberately keeps the schema (`auto_drop_schema=False`, `delete(force_drop=False)`), so even "deleted" decks leave orphaned schemas behind (which also block re-registering the same deck name — see `TenantBaseForm.clean_name`).

## Proposed work

- [ ] **Define "retirable"**: e.g. suspended (trial and/or paid period lapsed per the #1729 status model) **and** no staff login for N months (`Tenant.last_staff_login` is already cached) **and** not admin-flagged as exempt (comped decks, the library, etc.).
- [ ] **Reminder series before any action**, reusing the daily status-check + notice-ledger machinery from the #1729 plan (e.g. T-60/T-30/T-14/T-7 days emails to the deck owner: "your deck will be archived on DATE — log in or subscribe to keep it").
- [ ] **Decide: delete vs recoverable archive** (see open question below), then implement it as an explicit admin-approved action (a review queue/report of retirable decks, not fully automatic destruction — at least initially).
- [ ] **Also sweep existing orphaned schemas** left by past admin deletes (schemas with no matching `Tenant` row).
- [ ] Grace path: a retired deck's owner asking for restoration within the retention window gets their deck back.

## Open question: delete or archive?

Options, roughly in increasing safety / decreasing savings:

1. **Hard delete** (`delete(force_drop=True)` after final notice): maximum savings, zero recoverability. Probably too risky for schools that may return next school year.
2. **Archive then delete**: `pg_dump --schema=<name>` to S3 (or the media storage bucket), verify the dump, then drop the schema and keep the `Tenant` row flagged `archived`. Recoverable by `pg_restore` within a retention window (e.g. 1–2 years), after which dumps are purged. Gets the migration/backup savings immediately while staying recoverable.
3. **Detach only**: keep the schema but exclude archived decks from `migrate_schemas` and fan-out tasks. Least work, but schemas drift from the migration head, making later recovery progressively harder — probably a trap.

Option 2 looks like the right trade-off, but needs a decision on retention length and where dumps live.

## Acceptance

[...]

PR #2146, opened 2026-07-24
Admin tenant deletion: year-of-inactivity guard + schema drop (#2044)

tylerecouture:
### What?
The tenant admin can now truly delete abandoned decks — behind **four** independent protections (maintainer-requested retirement policy, #2044, tightened per review):

* **Explicit arming**: a new `Tenant.can_delete` boolean (default **False**, editable in the tenant admin) — deletion is refused until an admin deliberately turns it on for that specific deck.
* **Billing-state guard**: only a **suspended** deck qualifies. An active subscription, a running trial, or a managed-manually deck (both dates blank) is never deletable, regardless of arming or inactivity.
* **Inactivity guard**: `last_staff_login` must be recorded and **more than 365 days ago** (`INACTIVE_DELETE_DAYS`); a blank value can't prove a year of silence, so unknown decks stay protected.
* **Keyword confirmation** (existing flow, unchanged): type `owner/deckname` to confirm.

And the deletion is now real: **the deck's Postgres schema is dropped with the row** (`obj.delete(force_drop=True)`) instead of leaving an orphaned schema silently holding all the deck's data. `auto_drop_schema` stays `False` model-wide — only this deliberate admin flow drops.

### Why?
Deleting a deck previously removed only the `Tenant` row — the schema (and every quest, submission, and student record in it) stayed in Postgres invisibly. And nothing stopped an admin from deleting a deck a school was actively using. A real deletion now takes two separate admin decisions (arm, then delete), on a deck that is both billing-lapsed and staff-silent for a year, plus the typed confirmation.

### How?
* `Tenant.is_deletable` property: not public schema AND `can_delete` AND `is_suspended` AND year-old recorded `last_staff_login`. (`is_suspended` alone covers all three protected billing states.)
* `TenantAdmin.has_delete_permission(request, obj)` requires the Django permission **and** `obj.is_deletable` — hides the change form's Delete button and 403s the delete view (GET and POST). `delete_model` re-checks the guard (`PermissionDenied` — schema drops are unrecoverable, any future path fails closed) then deletes with `force_drop=True`.
* Migration `0022_tenant_can_delete`. ⚠️ Note: #2110 (open) also adds a `0022_*` tenant migration — whichever merges second needs a trivial renumber/merge-migration; I'll handle it as soon as one lands.

### Testing?
* `test_is_deletable__requires_arming_suspension_and_a_year_of_staff_silence` (pure unit: the armed+suspended+silent deck is deletable; each protection individually refuses — unarmed, subscribed, on trial, managed-manually, recent login, blank login, public schema).
* View tests: `test_delete_view__refused_when_not_armed_via_can_delete`, `test_delete_view__refused_while_billing_state_is_active` (all three billing states), `test_delete_view__refused_for_recently_active_deck`, `test_delete_view__refused_without_last_staff_login_on_record`, `test_delete_model__guard_fails_closed_even_if_reached_directly`.
* `test_delete_view__uses_delete_model_and_drops_schema`: a real schema is verified **gone** after the admin delete.
* Existing permission/keyword tests unchanged in behavior — the fixture deck is seeded armed + suspended + 366 days silent so they keep exercising the Django perms and confirmation mechanics.
* Full serial suite: **1559 tests, OK**; `ruff` and `makemigrations --check` clean.

Real dev admin (`localhost:8000/admin/`), superuser.

**The arming switch** (default off), with its help text:
checkbox

**Unarmed (or otherwise protected) deck** — no Delete button:
protected

**Armed + suspended + 400 days silent** — Delete available:
allowed

**Keyword confirmation** (unchanged flow) before anything is dropped:
confirm

### Anything Else?
* `last_staff_login` is the nightly-cached field — worth running the "Refresh deck stats" action before a deletion spree so the guard judges fresh data.
* This implements the delete side of the #2044 retirement decision.

### Review request
@tylerecouture

https://claude.ai/code/session_01UAo12812TYt3GJgpZi7Pmn


tylerecouture:
Lets add one more protection.  Can't delete a tenant that has an active subscription or trial (or both are blank), and add a new boolean field on the tenant model can_delete that is false by default.  So admin must first turn that on before deleting it.

tylerecouture (commented):
Lets add a couple more protections. Can't delete a tenant that has an active subscription or trial (or both are blank), and add a new boolean field on the tenant model can_delete that is false by default. So admin must first turn that on before deleting it.

tylerecouture:
Both protections added in `689d7cb`:

* **`can_delete` arming field** — new boolean on `Tenant` (default **False**, editable in the tenant admin change form with help text; migration `0022_tenant_can_delete`). Deletion is refused until it's deliberately turned on, so a real deletion now takes two separate admin decisions plus the typed keyword.
* **Billing-state guard** — implemented as "must be **suspended**": `is_suspended` is exactly *not subscribed/in grace AND not on trial AND not both-dates-blank*, so an active subscription, a running trial, and a managed-manually deck are all protected in one check, matching your list.

`is_deletable` is now: not public schema **and** `can_delete` **and** `is_suspended` **and** recorded `last_staff_login` &gt; 365 days ago. New tests cover every refusal individually (unarmed / subscribed / trial / comped / recent login / blank login / public) plus the armed-and-eligible happy path; full suite 1559 OK. The PR description and screenshots are updated — including a shot of the new checkbox with its help text.

One heads-up: this adds tenant migration `0022_tenant_can_delete`, and open #2110 also carries a `0022_*` tenant migration — whichever merges second gets a trivial renumber from me right away.

---

Issue #1977, opened 2026-07-16
Map generation inconsistancy


tylerecouture:
Each time I generate this map, it toggles between the two layouts:

<img width="879" height="939" alt="Image" src="https://github.com/user-attachments/assets/b823e214-84a0-4746-a32a-709797878675" />

<img width="862" height="925" alt="Image" src="https://github.com/user-attachments/assets/8b88945e-2c4a-478f-8440-ea0173e0478f" />

Why?  Is there a way to make this deterministic?  Also, is there a way we can allow the user to SET the order of the campaigns, from left to right?  For example we could we add a new "map sort order" to the campaign model?

tylerecouture:
Clsoed in #2016

tylerecouture:
This doesn't work.  I set the "Blender Donut Project Level 1" campaign sort value to 1000, and the other is at 0 (default), but they still ignore it.  Blender Basics 1 appears on the left always.

tylerecouture:
The blender basics Category/Campaign still is not vertically straight (there is a bend between Blender Basics 2. and 3. Where the Blender Donut Project branches off.

Also, the Category/Campaign sort order does nothing in this example see the settings in the screenshot, Donut should be to the left with a lower value.  This is on current staging as of 23 July 2026:

<img width="307" height="523" alt="Image" src="https://github.com/user-attachments/assets/22b51b89-413b-496a-8682-3cd11b47a1f4" />

<img width="319" height="554" alt="Image" src="https://github.com/user-attachments/assets/4741e56e-2402-4535-a23d-d775f324ddf1" />

<img width="823" height="792" alt="Image" src="https://github.com/user-attachments/assets/d94ef29f-7e48-4fb9-803f-ad5fad9ace72" />

Issue #1787, opened 2025-05-26
Quests in Campaign in Map Should Remain Vertically Stacked


fjbagley:
If a quest is a prerequisite for multiple quests, then the subsequent quest within the same campaign should remain vertically stacked instead of getting shifted sideways unless both subsequent quests are within the same campaign. 

In this example, Blender Basics 3 should remain right below Blender Basics 2.

![Image](https://github.com/user-attachments/assets/1e9e9a03-2ace-486e-b43a-b1fb87986cf4)

PR #2152, opened 2026-07-24
Make campaign map_order work for connected campaigns, compactly (#1977)

tylerecouture:
### What?

Reopened #1977: setting a campaign's **map_order** had no effect when one campaign branches off another — e.g. a *Blender Donut* campaign whose first quest continues from a *Blender Basics* quest. No matter the value, Donut stayed on the right. This fixes that, does it **without spreading the map out**, and keeps campaign-less "connector" nodes (the intro quest, a shared badge) sitting where they belong after a reorder.

### Why?

The left-to-right campaign ordering grouped map nodes into columns with a **union-find over connected components**. A cross-campaign prerequisite edge (Donut branching off Blender Basics) joined the two campaigns into a **single column**, so map_order could never separate them — setting one lower did nothing.

My first attempt (each campaign its own column, repacked left-to-right) fixed the ordering but **spread complex maps out**: campaign-less "bridge" nodes — like the shared *ByteDeck Proficiency* badge on the default map — each became their own column, widening every deck's main map.

### How?

Instead of repacking, **permute the campaigns among the x-slots dagre already produced**:

- Sort campaigns by `(map_order, smallest quest id)` and drop the i-th into the i-th slot from the left.
- The **set of x positions is unchanged**, so the map stays exactly as compact as dagre made it, and only the campaigns swap places.
- Each campaign's quests shift together as a rigid block, so vertical stacking (#1787) is preserved.
- Sorted slots + a deterministic order key also stop the layout **toggling between generations** (the original #1977 complaint).

Then **re-center campaign-less connector nodes**: moving a campaign can strand a campaign-less node — the intro quest that leads into a campaign, a shared badge — over the campaign's *old* slot (e.g. the "Start Here" intro quest left sitting above where a campaign used to be). After the permutation, each campaign-less node is re-centered over the mean x of its neighbours so it follows the campaign it connects to. **Both steps are skipped when nothing moved** (every campaign at the default map_order), so those maps render **byte-for-byte as before**.

### Testing?

`djcytoscape/tests/test_map_order_render.py` — the existing headless render test (real vendored cytoscape/dagre + real `maps.js` in Chromium) now covers, in addition to the three prior cases (lower→left, swap, equal→id-tiebreak):

- **Connected campaigns** — B branches off the side of A, B has the lower map_order, and must render left of A (checked both directions). Fails on the old grouping; passes now.
- **Intro node follows its campaign** — a campaign-less intro node that leads into a campaign must sit over that campaign after a reorder, not be stranded over the reordered-away one (checked both directions). Fails without the re-centering; passes with it.

Full `djcytoscape` suite (81 tests) green; `ruff` clean. I also verified on the running `hackerspace` deck that the **default 6-campaign map is positioned identically** to before this change (same campaign x-coordinates, no overlaps).

Two connected campaigns on the real running app: *Blender Donut* (map_order **0**) branches off *Blender Basics* (map_order **10**), so Donut should be on the left.

| Before (map_order ignored — Donut stuck right) | After (Donut ordered left, same compact spacing) |
| --- | --- |
| before | after |

And the connector-node re-centering, on a campaign with an internal branch-and-rejoin plus an intro quest — the "Start Here" intro sits centered above the quest it leads into, the spine stays straight, and the two branch quests spread symmetrically and rejoin:

diamond

### Anything Else?

- The other half of the reopened report — *Blender Basics not vertically straight (a bend where Donut branches off)* — I could **not** reproduce on `develop`, including with the internal branch-and-rejoin shown above: the spine stays straight (the #1787 / #2023 straightening). That screenshot was taken on **staging**, which likely hasn't merged those fixes yet, so it should resolve once staging catches up.

### Review request
@tylerecouture

### DIFF
--- a/src/djcytoscape/static/djcytoscape/js/maps.js
+++ b/src/djcytoscape/static/djcytoscape/js/maps.js
@@ -179,73 +179,51 @@ addedCampaignLayoutEdges.remove();
 /***************************************
  * #1977: order the campaign columns left-to-right by Category.map_order.
  *
- * dagre decides the order of same-rank nodes with a crossing-minimization heuristic and ignores
- * the order elements were fed in, so the campaign order can't be set before layout — it has to be
- * imposed on the finished layout. Each campaign (a compound node + its member quests) and each
- * connected clump of campaign-less quests is one connected "column"; we sort the columns by their
- * campaign map_order (ties, and campaign-less columns, fall back to the smallest node id — the
- * deterministic #2012 order) and repack them left-to-right, shifting each column horizontally as a
- * rigid block so its internal shape and every node's vertical position are untouched.
+ * dagre decides the order of same-rank nodes with a crossing-minimization heuristic and ignores the
+ * order the elements were fed in, so campaign order has to be imposed on the finished layout.
+ *
+ * Rather than repack every column left-to-right (which spreads the map out whenever campaign-less
+ * "bridge" nodes — e.g. a shared badge, or the intro quest — sit between campaigns, and which merged
+ * two campaigns into one un-orderable column when one branches off the other), we PERMUTE the
+ * campaigns among the x-slots dagre already gave them: sort the campaigns by (map_order, smallest
+ * quest id) and drop the i-th into the i-th slot from the left. The set of x positions is unchanged,
+ * so the map stays exactly as compact as dagre made it and campaign-less nodes don't move — only the
+ * campaigns swap places. Each campaign's quests shift together as a rigid block, so their internal
+ * shape and every node's vertical position (the #1787 vertical stacking) are untouched. Because both
+ * the sorted slots and the desired order are deterministic, the result no longer toggles between
+ * generations even though dagre's raw order does (the original #1977 complaint).
  ***************************************/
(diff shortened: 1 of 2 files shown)


tylerecouture:
Thanks @coderabbitai — addressed one, declining the other:

1. **Mirrored connected-campaign assertion** — done (`1a736b9`). The connected test now also checks the swapped order (A lower → A left), so the fix is proven direction-independent for the connected case too.

2. **Extract a pure `planCampaignShifts` helper for unit testing** — skipping. The stated benefit is "fast, deterministic unit tests instead of the slow Playwright test," but this repo has **no JavaScript test runner** (no Jest/Vitest/node harness) — `maps.js` is only ever exercised through the headless render test, which already runs in CI and now covers the connected case in both directions. Extracting the logic wouldn't buy a faster test without standing up JS test infrastructure, and to make the helper reachable by a test I'd have to hang it off the global scope of the map page purely for tests, which I'd rather not do. If a JS unit-test harness lands later (e.g. alongside the #2053/#503 front-end-testing work), this is a good candidate to revisit.

---

PR #2178, opened 2026-07-24
HOTFIX: Admin cap is always authoritative (suspension resets it to 5 once) + Maintenance status

tylerecouture:
### What?
Two maintainer requests from live testing (2026-07-24), reworked per review feedback on this PR:

* **The admin-set cap is always the enforced cap.** Lowering a suspended deck's **Max active users to 1** used to change nothing — the code overrode every suspended deck to the trial constant (5) at read time. Now `effective_max_active_users` is simply the field, in every state. **"Revert to trial limits" happens as a one-time WRITE instead**: when a deck's suspension episode begins, the nightly `deck_status_check` writes the trial default (5) into `max_active_users` exactly once (`reset_cap_on_new_suspension`), after which the admin's value — higher for a comp, lower for a wind-down — always wins.
* **New Maintenance status.** A minimal-cost subscription (e.g. $5/year) that keeps a deck alive — it never expires or suspends, and (since deletion requires suspension, #2146/#2044) can never time out for deletion — while registration stays capped at the trial limit. Derived, not stored: `is_on_maintenance` = actively paid AND cap at (or below) the trial limit, which is exactly what the maintenance Stripe price's `metadata.max_active_users = 5` produces. The Subscription details page gives it its own label and copy (see screenshot); paid decks with higher caps keep the green Subscribed status.

### Why?
An admin adjusting a cap is making a deliberate access decision; the old read-time override silently reversed it (the "cap set to 1 still shows 5" report). And schools that only want to *preserve* a deck between active years need a cheap way to keep it from suspending (and eventually deleting) without paying for seats they aren't using — that's the Maintenance tier.

### How?
* `Tenant.effective_max_active_users` → returns `max_active_users`, always. `Tenant.is_on_maintenance` (new, derived).
* `tenant/notices.py reset_cap_on_new_suspension` (new): computes the episode's first suspended day (day after `trial_end_date`, or after `paid_until` + 30-day grace — whichever is later), records a `cap-reset` `DeckNotice` ledger row for exactly-once semantics, and applies the write via a targeted `QuerySet.update()` + deck-cache invalidation. **Not** gated by `DECK_NOTICES_ENABLED` (it's enforcement, not communication). Episodes that began more than 7 days ago (`CAP_RESET_CATCHUP_DAYS`) are recorded but **not** reset — decks already suspended when this ships (like the live `bingo` deck with its hand-set cap of 1) are grandfathered and never clobbered, while multi-day beat outages still catch up.
* `deck_status_check` runs the reset between the cached-field refresh and the notices pass. No migration needed (the ledger row reuses the existing `KIND_SUSPENDED` kind with a distinct threshold).
* Copy: the suspended banner/status show the live field; the grace-period "will revert to trial limits (max 5)" predictions (banner, page, expiry email) state the constant the reset will write; the Maintenance branch renders before Subscribed on the subscription page.

### Testing?
Test-driven — these fail without the fix:
* `test_effective_max_active_users__always_the_admin_field` — suspended cap-1 → 1 **and** suspended cap-80 → 80 (the revert is now the nightly write, deliberately pinned).
* `test_page__suspended_deck_shows_its_admin_cap` — the original report's scenario: status copy + seats table show 1, not 5.
* `SuspensionCapResetTest` — fresh episode resets 80→5 (and −1→5) exactly once; an admin adjustment afterwards sticks; **old episodes are grandfathered** (recorded, cap untouched); unsuspended decks untouched; runs with `DECK_NOTICES_ENABLED=False` to prove it isn't gated.
* `test_deck_status_check__resets_cap_on_fresh_suspension` — the task wiring.
* `test_is_on_maintenance__paid_at_or_below_trial_cap` + `test_page__maintenance_subscription_gets_its_own_status` — the new status, including "paid with higher cap stays Subscribed".
* Full serial suite: **OK**; `ruff` and `makemigrations --check` clean.

Captured from the real dev deck (staff login). FontAwesome served locally in the captures (the dev proxy blocks its CDN); identical assets to production.

**Before** — suspended deck, admin cap set to 1, but banner and page say max 5 / 4 remaining (the report):
before

**After** — banner, status copy, and seats table all show the admin's cap (1 of 1 seats used, at-limit warning appears):
after

**New Maintenance status** — paid deck with the trial-limit cap:
maintenance

### Anything Else?
* **Ops for the Maintenance tier**: create the $5/year Price in the Stripe dashboard with `metadata.max_active_users = 5` — checkout/sync (PR 6 / #2110) then produces maintenance decks with no further code. Decks can also be put on maintenance by hand (set `paid_until` + cap 5 in the tenant admin).
* Deliberate behavior notes: a suspended deck that was ALREADY suspended before this deploys keeps its current cap (grandfathered — protects hand-set caps like bingo's 1); an admin who wants such a deck at 5 just sets it. A maintenance deck that lapses goes through the normal grace → suspension flow; the reset then finds the cap already at 5.
* Hotfix to `staging`; flows back to `develop` with the next staging sync. Heads-up for that sync: open epic PR #2110 touches the same `tenant/models.py` region — I'll resolve the overlap on the epic branch as usual when this lands.

### Review request
@tylerecouture

https://claude.ai/code/session_01UAo12812TYt3GJgpZi7Pmn

### DIFF
--- a/src/hackerspace_online/context_processors.py
+++ b/src/hackerspace_online/context_processors.py
@@ -38,7 +38,6 @@ def deck_status(request):
     """
     from django.urls import reverse
 
-    from tenant.models import TRIAL_MAX_ACTIVE_USERS
     from tenant.utils import get_current_deck
 
     deck = get_current_deck()
@@ -47,7 +46,4 @@ def deck_status(request):
     return {
         "current_deck": deck,
         "deck_subscribe_url": reverse('decks:subscription'),
-        # what a lapsed deck reverts to -- the grace banner can't derive this from
-        # the deck's effective cap, which is still the paid cap during grace
-        "trial_cap": TRIAL_MAX_ACTIVE_USERS,
     }
--- a/src/hackerspace_online/context_processors.py
+++ b/src/hackerspace_online/context_processors.py
@@ -38,6 +38,7 @@ def deck_status(request):
     """
     from django.urls import reverse
 
+    from tenant.models import TRIAL_MAX_ACTIVE_USERS
     from tenant.utils import get_current_deck
 
     deck = get_current_deck()
@@ -46,4 +47,8 @@ def deck_status(request):
     return {
         "current_deck": deck,
         "deck_subscribe_url": reverse('decks:subscription'),
+        # what a fresh suspension will RESET the cap to (reset_cap_on_new_suspension)
+        # -- the grace banner predicts it, and can't derive it from the deck's
+        # current cap, which is still the paid cap during grace
+        "trial_cap": TRIAL_MAX_ACTIVE_USERS,
     }
--- a/src/templates/deck_status_banner.html
+++ b/src/templates/deck_status_banner.html
@@ -32,7 +32,7 @@ Priority: suspended (everyone) > over-limit (staff) > expiring soon (staff) > tr
                 {% if current_deck.days_until_expiry < 0 %}
                     <strong>Subscription expired</strong> &mdash; this deck is in its grace period, which ends in
                     {{ current_deck.grace_days_remaining }} day{{ current_deck.grace_days_remaining|pluralize }};
-                    it will then revert to trial limits (max {{ trial_cap }} current students).
+                    it will then revert to trial limits (max {{ current_deck.suspension_cap }} current student{{ current_deck.suspension_cap|pluralize }}).
                 {% elif current_deck.is_on_trial %}
                     <strong>Trial ending soon:</strong> this deck's free trial ends in
                     {{ current_deck.days_until_expiry }} day{{ current_deck.days_until_expiry|pluralize }}.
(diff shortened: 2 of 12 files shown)


tylerecouture:
Both requests are in as of `62b7399` (PR title, description, and screenshots updated to match):

**Cap model reworked as you specified** — no more runtime override or `min()`: `effective_max_active_users` is now simply the admin field, always. "Revert to trial limits" is a one-time **write** of the default (5) when a suspension episode begins, applied by the nightly `deck_status_check` (idempotent per episode via the notices ledger, not gated by the notices rollout flag). After that, whatever you set in the admin — higher or lower — wins. One deliberate wrinkle: decks **already suspended** when this deploys are grandfathered (recorded, cap untouched), so your bingo deck's hand-set cap of 1 won't get clobbered back to 5 on the first nightly run; only *new* suspensions get the reset.

**Maintenance status added** — derived rather than stored: any actively paid deck whose cap sits at (or below) the trial limit is Maintenance, which is exactly what a $5/yr Stripe price with `metadata.max_active_users = 5` produces through the sync. It gets its own blue label + copy on the Subscription details page ("won't expire, be suspended, or time out for deletion, but registration stays capped at the trial limit — upgrade any time"); paid decks with higher caps keep the green Subscribed badge. Screenshot in the description. Ops note: just create the maintenance Price in the Stripe dashboard with that metadata — no further code needed; you can also comp a deck onto maintenance by hand (paid_until + cap 5 in the tenant admin).

Full serial suite green; no migrations. All three states in the screenshots are real captures from the dev deck.

---

Issue #2081, opened 2026-07-22
Per-request memory blowups can OOM the web host (unbounded exports + unpaginated views)


tylerecouture:
### Background
Production (`c5a.large`, 3.8 GiB, all containers on one host) has been **OOM-killed**: `dmesg` shows the kernel killing a uwsgi worker twice, each having grown to **~1.6 GiB RSS on a single request**. A 4 GiB swap backstop + bounded uwsgi workers (separate PR) buy headroom, but the durable fix is removing the requests that materialize that much memory.

A code review found the systemic cause and the specific culprits.

### Systemic finding
**`paginate_by` does not appear anywhere in `src/`.** Every Django `ListView` is fully unbounded; the only pagination is two hand-rolled `paginate()` helpers (`quest_manager/views.py`, `notifications/views.py`). So every list screen scales its memory with table size — fine at 41 tiny summer decks, dangerous under September load.

### Ranked culprits (all web-request paths, not celery)

**1. django-import-export Quest/Badge admin exports — HIGH.** `QuestAdmin` (`quest_manager/admin.py:322`, resource `:84`) and `BadgeAdmin` (`badges/admin.py:166`, resource `:31`) are `ImportExportActionModelAdmin` with no `get_export_queryset` override. The admin **Export** button serializes the *entire* table into an in-memory `tablib.Dataset` (all rows at once, then a second full copy as the CSV/XLSX string). `QuestResource` still exports the full `instructions` summernote **HTML** per quest (attachment cap ~16 MB each). Hundreds of quests with large WYSIWYG HTML = the classic hundreds-of-MB→GB single request. Trigger: a teacher/owner clicks Export. **Most probable 1.6 GiB offender.**
→ Fix: override `get_export_queryset` to `.iterator()`/chunk, drop `instructions` from the default export, and/or move export to a celery job that writes a downloadable file.

**2. `ProfileList` (all active students) — HIGH/MED.** `profile_manager/views.py:48` (`get_queryset` `:76` → `Profile.objects.all_students().get_active()`), no `paginate_by`. `queryset_append` (`:59`) `select_related('user__portfolio')` + `Prefetch`es every student's `coursestudent_set` (course+block). Whole set + prefetch materialized and rendered in one pass. `ProfileListInactive` (`:142`, all students *ever*) is worse. Trigger: teacher opens the student list.
→ Fix: `paginate_by` (+ template paginator controls).

[...]

PR #2191, opened 2026-07-29
Always offload "regenerate all maps" to celery (#2081)

tylerecouture:
### What?

The staff **"regenerate all maps"** action (`djcytoscape:regenerate_all`) now **always** offloads to the existing `regenerate_all_maps` celery task, instead of regenerating maps inline in the request whenever there were ≤ 5 maps.

### Why?

Part of #2081 (per-request memory blowups OOM-killing the single-host web box). Item 6:

> `regenerate_all` … only offloads to celery when `scapes.count() > 5` — else loops `scape.regenerate()` inline, building full graph JSON in memory.
> → Fix: always route through the existing `regenerate_all_maps`/`regenerate_map` celery tasks.

Regenerating a map builds that map's full graph JSON in memory. The old inline branch did this for **every** map in a single web request, so per-request memory scaled with the number and size of maps — exactly the kind of unbounded in-request work #2081 is about. The `count > 5` threshold was an arbitrary cutoff that still left the common (small-deck) case doing the work inline.

### How?

- Drop the `count > 5` branch and the inline regeneration loop; always call `regenerate_all_maps.apply_async(args=[request.user.id], queue='default')`.
- The task (`djcytoscape/tasks.py`) is already a complete equivalent of the inline loop: it regenerates each map, **deletes any map whose initial object no longer exists** (notifying the user), and sends a **completion notification** — so no behaviour is lost by moving the work off the request.
- The user now always sees a "processing in the background" message and gets the task's completion notification, rather than an immediate flash message for small decks. The per-map "initial object gone, map deleted" warning is delivered by the task as a notification (it already was, for the >5 path).

The single-map `regenerate` view and `ScapeUpdate`/`ScapeGenerateMap` (which regenerate one map and redirect to it) are intentionally left inline — routing those through celery would change their "show the freshly-regenerated map on redirect" UX and belongs in a separate change.

### Testing?

`djcytoscape/tests/test_views.py`: replaced the old `>5 maps → background` test with `test_regenerate_all__always_offloads_to_background_task`, which mocks `regenerate_all_maps.apply_async` and asserts it is called once with the requesting user's id and that the user is told it's processing in the background. The existing `test_regenerate_all__redirects_to_primary` still covers the redirect. The task's own behaviour (regenerate valid / delete broken / notify) remains covered by `test_tasks.py`.

Ran `djcytoscape.tests.test_views` and `djcytoscape.tests.test_tasks` (15 tests) — all green; `ruff` clean; no model changes / no migration.

N/A — no template/CSS/JS change. The only visible difference is the flash message wording and that the action always completes in the background (with a notification) rather than inline for small decks.

### Anything Else?

One item of the larger #2081 memory work; does **not** close the issue.

### Review request
@tylerecouture

---

## Summary by CodeRabbit

* **New Features**
  * Map regeneration now always runs in the background, allowing requests to return immediately.
  * Added a single informational message indicating regeneration is being processed in the background.

* **Bug Fixes**
  * Improved consistency by removing variable synchronous/asynchronous behavior depending on the number of maps.

* **Tests**
  * Updated regeneration view tests to assert background offloading behavior in all cases.

* **Documentation**
  * Updated code-style guidance for merged-state comments.


src/djcytoscape/views.py
```diff
@@ -229,17 +229,12 @@ def regenerate(request, scape_id):
...
-                scape.regenerate()
-            except scape.InitialObjectDoesNotExist:
-                messages.warning(request, f"The initial object for the '{scape.name} Map' no longer exists. The map has now been removed too.")
-
-        messages.success(request, "All valid quest maps have been regenerated.")
+    # Regenerating every map inline builds each map's full graph JSON in memory within a
+    # single request, which scales with the number/size of maps and contributes to the
+    # per-request memory blowups tracked in #2081. Always offload to the celery task
+    # (which regenerates each map, deletes any whose initial object is gone, and notifies
+    # the requesting user) instead of only doing so past an arbitrary map-count threshold.
```
tylerecouture: These comments refer to deleted code. Stop doing that (please add to CLAUDE.md to prevent this in the future).  There will be missing context for these comments once the PR is merged (e.g. referring to "arbitrary map-count threshold).

tylerecouture: Fixed in 453c3a0 — reworded the comment to describe only the current behaviour (why we offload to celery), with no reference to the removed threshold. Also added a convention to `CLAUDE.md` under Code Style & PR Conventions: comments must describe the code as it is now, never the removed code or previous approach (no "instead of…/no longer…/used to…"), so the "why we changed it" narrative lives in the PR/commit instead of a source comment. Thanks for the catch.

---

PR #2229, opened 2026-07-31
test: add redis integration tests covering the cache + broker paths CI can't see

tylerecouture:
### What?
Adds `src/hackerspace_online/tests/test_redis_integration.py`: two `SimpleTestCase`s that exercise the **installed redis-py / kombu / django-redis stack against a real redis server**:

- **`RedisCacheClientTests`**: the django-redis cache client (`RedisCache` + `DefaultClient`): `set/get/add/delete`, `incr`/`ttl`, `get_or_set`, `set_many`/`get_many`, `delete_pattern`, and the raw client `ping`/`info`.
- **`RedisBrokerTransportTests`**: kombu's redis **broker transport producing *and* consuming** (a kombu produce/consume round-trip, an **in-process Celery worker** that consumes an enqueued task, and a `control.ping` to that worker's fanout/pidbox mailbox), plus a redis **pub/sub** round-trip.

They connect to the same redis the app uses (`settings.REDIS_HOST`/`REDIS_PORT`) on a **dedicated db index (15)** (never touching the app's broker/cache dbs 0/1/2), and **skip (not fail)** when redis is unreachable.

### Why?
The unit suite is structurally blind to redis: test settings force `LocMemCache` (so django-redis never runs) and no Celery worker runs during tests (tasks are only *enqueued*, never *consumed*). So a redis-py / kombu / django-redis incompatibility passes CI green even if it would break the cache or the broker in production.

This blind spot just bit us with the **redis 7 → 8 bump (#2202)**: green CI proved nothing about the two paths that actually matter, and the safety of that merge had to be established by a hand-run A/B smoke test outside CI. The *next* redis-py bump (a future 9.x, or a kombu/celery change) would again sail through green. These tests make CI itself cover those paths, so the next break is caught automatically.

### How?
- `_redis_reachable()` gates the classes via `setUpClass` → `SkipTest`, so a bare local `manage.py test` with no redis still passes; the skip branches carry `# pragma: no cover` (they never run where redis is up).
- A throwaway module-level Celery app (`set_as_current=False`, so it can't displace the real `hackerspace_online.celery` app during the suite) pointed at db 15 drives the worker round-trip via `celery.contrib.testing.worker.start_worker`.
- Everything is isolated on db 15 and flushed per test, so it's safe to run alongside the rest of the suite.
- No production code changes; test-only addition. No CI-config change needed: it runs inside the existing suite, and in CI `REDIS_HOST=redis` (the `redis:7` service) is up, so the tests execute on every push/PR rather than skipping.

### Testing?
Validated locally in an isolated venv against a real redis 7.0.15 server, on **redis-py 8.1.0** (the version `develop` now pins after #2202):

- **redis up →** all 7 tests pass (`Ran 7 tests ... OK`).
- **redis down (dead port) →** both classes skip cleanly, no failures (`OK (skipped=2)`).
- `ruff check` on the new file: clean.

This branch is cut from `develop` post-#2202, so this PR's own CI run is effectively the first time these paths are exercised against redis-py 8 in the real CI environment.

N/A (test-only, no front-end change).

### Anything Else?
Note on the worker test: it uses celery's official in-process `start_worker`, which runs the real async redis transport consumer, so this covers more than a synchronous drain. The one thing deliberately left out is a load/high-concurrency scenario: this is a correctness smoke test, not a benchmark.

### Review request
@tylerecouture

---

### DIFF
--- a/src/hackerspace_online/tests/test_redis_integration.py
+++ b/src/hackerspace_online/tests/test_redis_integration.py
@@ -1,0 +1,220 @@
+"""Integration tests that exercise the installed redis-py / kombu / django-redis
+stack against a REAL redis server -- the paths the rest of the suite can't see.
+
+The default test settings force ``LocMemCache`` (so django-redis never runs) and
+no Celery worker runs during tests (tasks are only enqueued, never consumed).
+So a redis-py / kombu / django-redis incompatibility -- e.g. a future redis-py
+major bump like the redis 7 -> 8 one (#2202) -- sails through the unit suite
+green even though it could break the cache or the broker in production. These
+tests close that blind spot by exercising, against a live redis:
+
+* the **django-redis cache client** (``RedisCache`` + ``DefaultClient``), and
+* **kombu's redis broker transport**, producing AND consuming -- including an
+  in-process Celery worker round-trip and the fanout control mailbox -- plus
+  redis pub/sub (which celery's control plane uses).
+
+They connect to the same redis the app uses (``settings.REDIS_HOST`` /
+``settings.REDIS_PORT``) but on a dedicated db index (``REDIS_TEST_DB``), so they
+never touch the app's broker/cache dbs (0/1/2), and they **skip** (rather than
+fail) when redis is unreachable -- e.g. a bare local ``manage.py test`` with no
+redis running. CI always has the ``redis:7`` service up (``REDIS_HOST=redis``),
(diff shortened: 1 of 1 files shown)


tylerecouture:
@coderabbitai review

Triggering the review now that the earlier rate-limit cooldown has elapsed (the initial run hit the limit before it could start).

---

PR #2236, opened 2026-07-31
refactor: vendor Font Awesome 4.7.0 locally instead of loading from cdnjs

tylerecouture:
### What?
Vendors Font Awesome 4.7.0 for the main LMS tenant so it loads from local static instead of cdnjs. Adds the canonical npm `font-awesome@4.7.0` `css/font-awesome.min.css`, replaces the `FontAwesome` webfonts under `src/static/fonts/` with that package's matched 4.7.0 set, and switches `head_css.html` from the cdnjs `<link>` to `{% static 'css/font-awesome.min.css' %}`.

### Why?
First of a small series making the LMS tenant's front-end assets consistently local. The Bootstrap ecosystem (Bootstrap, jQuery, bootstrap-table, select2 theme) was already vendored; Font Awesome was still pulled from cdnjs for no principled reason. Vendoring it removes a runtime dependency on an external CDN and matches the rest of the stack.

The webfonts already committed under `src/static/fonts/` were a different build and were unused (cdnjs served both the CSS and its own fonts), so this replaces them with the exact 4.7.0 set that matches the vendored CSS, avoiding any CSS/font mismatch.

### How?
- `npm pack font-awesome@4.7.0`, then copied `css/font-awesome.min.css` and the six webfont files into `src/static/`.
- The CSS references `../fonts/fontawesome-webfont.*`, which from `src/static/css/` resolves to `src/static/fonts/`.
- `head_css.html`: removed the cdnjs `<link>` (and the stale commented-out local one), added the local `{% static %}` link.
- No version change (still 4.7.0), so rendering is identical to before. Main LMS tenant only; the public tenant is untouched.

Note on versions/security: 4.7.0 is the final FA 4.x release (no in-major bump available). FA 4 is end-of-life; moving to FA 5/6 is a major upgrade (icon class renames `fa` → `fas`/`far` across ~317 usages) and is intentionally out of scope for this localization PR.

### Testing?
`ruff check src` clean (template + static assets only; no Python touched).

Real-app screenshot caveat: this session has no Docker/Postgres to run a seeded deck, so the capture below is a **headless-Chromium reproduction**. It loads the vendored `css/font-awesome.min.css` from the static root (so `../fonts/` resolves exactly as in the app) and renders a spread of the FA icons the LMS uses. All render as real glyphs from the local files (no missing/tofu boxes), confirming the CSS + webfonts resolve. Because the version is unchanged (4.7.0), the before (cdnjs) and after (local) render identically; a real-deck capture can confirm on staging.

Headless reproduction: vendored local Font Awesome rendering (real-deck capture pending).

FA local render

### Anything Else?
Next in the series, each its own PR: jQuery UI CSS, KaTeX, Chart.js, clipboard.js, and the standalone jQuery loads in djcytoscape/portfolios.

### Review request
@tylerecouture


tylerecouture:
@coderabbitai review

The initial run hit the per-developer rate limit; the stated cooldown has now elapsed, so re-requesting the review.

---

tylerecouture:
@coderabbitai review

PR #2246, opened 2026-08-01
refactor: vendor Chart.js 2.9.4 locally and consolidate the three CDN loads

tylerecouture:
### What?
Vendors Chart.js locally for the main LMS tenant and consolidates three CDN loads (at two different versions) down to one vendored 2.9.4. Adds `src/static/js/Chart-2.9.4.min.js` (Chart.js 2.9.4, version kept in the filename) and points `tag_chart.html`, `summary.html`, and `mark_calculations.html` at it via `{% static %}`.

### Why?
Part of the localize-front-end-assets series. Chart.js was loaded three times from CDNs at inconsistent versions: `tag_chart.html` and `mark_calculations.html` from cdnjs **2.5.0**, and `summary.html` from jsdelivr `chart.js@2` (which floats to the latest 2.x). Vendoring a single 2.9.4 (the last 2.x release) removes the external CDN dependency and makes the version consistent everywhere.

2.5.0 to 2.9.4 is within the same major and API-compatible with the existing chart configs (all use the 2.x `new Chart(ctx, {type, data, options:{scales:{xAxes,yAxes}}})` shape). No `type: 'time'` axes and no moment.js anywhere, so the plain `Chart.min.js` (not the moment-bundled build) is sufficient.

**Version/security note:** Chart.js 2.x is end-of-life. Moving to 3.x/4.x (latest 4.5.1) is a **breaking major**: the scales config changes from `xAxes`/`yAxes` arrays to a `scales` object, plus other API changes, so every chart config here would need rewriting. That upgrade is intentionally out of scope for this localization PR.

### How?
- `npm pack chart.js@2.9.4`, copied `dist/Chart.min.js` to `src/static/js/Chart-2.9.4.min.js`. The version is kept in the filename: it is self-documenting, and since the project does not hash static filenames (no `ManifestStaticFilesStorage`; cache-busting is manual via `?v=` in `head_css.html`), a version bump renames the file and busts the browser cache on its own.
- Replaced the three CDN `<script>` tags with `{% static 'js/Chart-2.9.4.min.js' %}`.
- Added `{% load static %}` to `tag_chart.html` and `summary.html` (`mark_calculations.html` already had it).
- Dependency graph verified: those three pages are the only ones that load Chart.js, and they cover every `new Chart()` usage (the `chart_tags_progress` / `chart_mark_distribution` / `chart_xp_progress` snippets are `{% include %}`'d by them).

### Testing?
`ruff check src` clean (templates + one static JS file; no Python).

Real-app screenshot caveat: no Docker/Postgres in this session, so the capture below is a **headless-Chromium reproduction**. It loads the vendored Chart.js and renders a bar chart with the same 2.x config shape the app uses (`type`, `datasets`, `options.scales.xAxes`/`yAxes`). It renders correctly (axes, labels, legend, title, grid), confirming Chart.js 2.9.4 drives the existing configs from the local file.

Headless reproduction: a bar chart rendered by the vendored local Chart.js 2.9.4 (real-deck capture pending).

Chart.js local render

### Anything Else?
Main LMS tenant only. Remaining in the series: clipboard.js, and the standalone jQuery loads in djcytoscape/portfolios. Per review, vendored assets keep their version in the filename; a small follow-up can retrofit that onto the already-vendored files (FA, KaTeX, and optionally the pre-existing Bootstrap/jQuery/bootstrap-table/select2) for consistency.

### Review request
@tylerecouture


tylerecouture (commented):
Why do we remove the version number form our filenames and static calls? Leaving the version name provide more accessible info, and helps bust the cache when we change them.

tylerecouture:
Good call, and your cache-busting reasoning holds here: the project doesn't hash static filenames (the non-S3 path uses plain `StaticFilesStorage`, no `ManifestStaticFilesStorage`), which is exactly why `head_css.html` cache-busts manually with `?v=`. So a versioned filename genuinely busts the cache on a version bump, and it reads the version at a glance.

I'd defaulted to version-less names only to match the existing vendored files (`bootstrap.min.js`, `jquery.min.js`, `bootstrap-table.min.js`, `select2-bootstrap.min.css`, and `font-awesome.min.css`), but I agree yours is the better convention. Applied here: the file is now `Chart-2.9.4.min.js` and the three `{% static %}` calls point at it (pushed).

For consistency, the already-merged Font Awesome (`font-awesome.min.css`) and KaTeX (`katex.min.css` / `.js`), plus those pre-existing Bootstrap/jQuery/bootstrap-table/select2 files, are still version-less. I'll name the remaining series assets (clipboard.js, jQuery) with versions, and I'm happy to do a small follow-up PR to retrofit versions onto the already-vendored ones so the whole static dir is consistent. Do you want that retrofit to include the pre-existing Bootstrap/jQuery/bootstrap-table/select2 files too, or just the ones from this series (FA, KaTeX)?

---

PR #2250, opened 2026-08-01
refactor: put version numbers in vendored static library filenames

tylerecouture:
### What?
Renames the vendored third-party CSS/JS library files under `src/static/` to carry their upstream version in the filename (updating their `{% static %}` references), and removes three confirmed-dead vendored files. CDN URLs are untouched.

| Old | New |
|-----|-----|
| `css/bootstrap.min.css` | `css/bootstrap-3.4.1.min.css` |
| `css/bootstrap_slate.min.css` | `css/bootstrap_slate-3.4.1.min.css` |
| `css/bootstrap-table.min.css` | `css/bootstrap-table-1.20.2.min.css` |
| `css/select2-bootstrap.min.css` | `css/select2-bootstrap-0.1.0-beta.10.min.css` |
| `css/font-awesome.min.css` | `css/font-awesome-4.7.0.min.css` |
| `katex/katex.min.css` | `katex/katex-0.18.1.min.css` |
| `js/bootstrap.min.js` | `js/bootstrap-3.4.1.min.js` |
| `js/jquery.min.js` | `js/jquery-2.2.4.min.js` |
| `js/bootstrap-table.min.js` | `js/bootstrap-table-1.20.2.min.js` |
| `js/bootstrap-table-export.min.js` | `js/bootstrap-table-export-1.20.2.min.js` |
| `js/tableExport.min.js` | `js/tableExport-1.9.8.min.js` |
| `katex/katex.min.js` | `katex/katex-0.18.1.min.js` |

### Why?
Follows the convention agreed on #2246: the version in a vendored library's filename is self-documenting and, because the project does not hash static filenames (no `ManifestStaticFilesStorage`; cache-busting is manual via `?v=`), a version bump renames the file and busts the browser cache on its own.

### How?
- `git mv` each file to its versioned name.
- Updated only the exact `{% static '...' %}` calls, so CDN URLs that contain the same filename (public tenant's BS5/FA/jQuery, `settings.py`, `tags/forms.py`) are not touched, plus the `src/static/...` doc paths in `head_css.html` comments.
- **Removed three dead files** (confirmed unused by a full-repo search, including the `SUMMERNOTE_CONFIG` css/js lists in settings.py): `css/featherlight.min.css`, `js/featherlight.js` (only trace was a commented-out `//$('img.pop').featherlight(...)` call, now removed), and `js/codemirror-formatting.js` (Summernote loads CodeMirror from cdnjs, not this local addon).
- Verified by grep: no stale `{% static %}` reference to any old name remains; all 12 new files exist; no reference to the deleted files remains; CDN refs unchanged.

### Testing?
`ruff check src` clean. No content or rendering change for the renames (byte-identical files, only renamed) and the deletions remove genuinely unused assets. No screenshot: nothing renders differently.

N/A: rename + dead-code removal, no visual change.

### Anything Else?
- **Chart.js** is versioned separately in #2246 (`Chart-2.9.4.min.js`, merged).
- **Not versioned here** (no clean upstream version to pin): the old Stefan-Petre datetimepicker (`css/datetimepicker.css`, `js/bootstrap-datetimepicker.js`, also slated for replacement in the Bootstrap 5 work) and `js/ie10-viewport-bug-workaround.js` (a version-less Bootstrap snippet).
- **Versioning the custom (ByteDeck-authored) css/js** (`custom_common.css`, etc.) is being handled as a follow-up, pending a choice between filename-versioning and a universal `?v=` (see review discussion), since those files change often and several load via `SUMMERNOTE_CONFIG`.

### Review request
@tylerecouture


tylerecouture (commented):
1. Double check with a thorough search that those unreferenced files aren't used by summernote in settings.py or elsewhere in a harder to find place. If not then delete them. I thought featherlight was used by the portfolio app?

2. Also version custom common, and other custom css and js.

tylerecouture:
**On (1):** done. Full-repo search (templates, JS, Python, and the `SUMMERNOTE_CONFIG` css/js lists in settings.py):
- **featherlight** (css + js): the only trace anywhere is a commented-out `//$('img.pop').featherlight($(this));` in `javascript.html`. Nothing in the portfolios app or elsewhere, so it isn't actually wired up. Deleted both files and that dead line.
- **codemirror-formatting.js**: Summernote's code view loads CodeMirror from cdnjs (`5.29.0`), and its js list only references `js/summernote-*.js`, never this local addon. Deleted.

All three removed in the latest commit (`31dee35`).

**On (2):** happy to, but one thing to decide first, because the custom files differ from the pinned libs. The libs never change, so a versioned filename is pure upside. The custom files (`custom.css`, `custom_common.css`, `custom.js`, the `summernote-*.js` plugins, etc.) change fairly often, and several load via `SUMMERNOTE_CONFIG` in settings.py, not just templates. So filename-versioning them means each future edit has to rename the file and update its refs in two places, versus today's one-line `?v=` bump. Two ways to make them consistent:

- **A. Filename version (like the libs):** `custom_common.css?v=2.1` becomes `custom_common-2.1.css` (updating `head_css.html` + `settings.py`), and files with no version get `-1.0`. Fully consistent, but every future edit is a rename + ref updates.
- **B. Keep `?v=`, make it universal:** add a `?v=` to every custom css/js that lacks one so they all cache-bust the same way. One-line bumps going forward.

For files that change this often I'd lean **B**, but it's your call. I'll do it as a focused follow-up PR either way (it touches settings.py, so it's cleaner kept separate from this rename PR). Which would you prefer?

---

PR #2251, opened 2026-08-01
test: cover bulk-edit unrecognized-action and drop-without-draft branches

tylerecouture:
## What

Covers two reachable, previously-untested branches in `quest_manager/views.py`. Test-only.

## Why

A full-suite coverage run flagged six partial branches in `quest_manager/views.py` (it sits at 99.5%). This PR closes the two that are cleanly reachable with robust tests:

- **`QuestBulkEditView.post`** (`656 -> 663`): a POST with selected quests but an `action` matching none of the handled actions (`delete`/`publish`/`unpublish`/`unarchive`) falls through the elif chain to the final `redirect("quests:quests")`. Only the recognized actions were tested.
- **`drop()`** (`2135 -> 2137`): dropping a submission that has **no** draft comment skips the `if sub.draft_comment: sub.draft_comment.delete()` branch. Only the with-draft-comment path was tested.

## How

In `src/quest_manager/tests/test_views.py`:

- `QuestBulkEditViewTests.test_bulk_edit__unrecognized_action_redirects_without_change`: posts an unhandled action and asserts a plain redirect, no queued message, and the quest left unchanged.
- `SubmissionViewTests.test_drop__deletes_submission_without_draft_comment`: drops a draft-comment-less submission and asserts it is deleted with a redirect.

## Remaining branches (deliberately not covered here)

For transparency, the other four flagged branches were assessed and left out; happy to follow up per your preference:

- **Guarded-unreachable (defensive) fall-throughs** (candidates for a `# pragma: no branch`, not a test):
  - `ApproveView` note/icon builder (`1283 -> 1295`): the "no button matched" exit of the `approve/comment/return/skip` elif chain, which cannot be reached because `post_has_valid_button()` guards the view.
  - `complete()` (`1881 -> 1914`): the "neither complete nor comment" exit, already noted redundant in a code comment ("this case is already covered at the top of the view").
- **Reachable but needs a specific/contrived scenario** (coverable in a follow-up):
  - `get_notification_kwargs` (`1324 -> 1327`): the approving staff user already being one of the student's `current_teachers()`.
  - `complete()` auto-approve path (`1870 -> 1914`): a submission with `do_not_grant_xp=True` completing a no-verification quest.

## Testing

- `QuestBulkEditViewTests` + `SubmissionViewTests` green (39 tests).
- `coverage` confirms branches `656->663` and `2135->2137` are now covered.
- `ruff check` clean.

## Screenshots

N/A: test-only; no user-facing or front-end change.

[...]

PR #2257, opened 2026-08-05
test: finish quest_manager/views.py branch coverage (test reachable, pragma dead)

tylerecouture:
## What

Closes the remaining flagged branches in `quest_manager/views.py`: one is genuinely reachable (now tested), three are guarded-unreachable defensive branches (now marked `# pragma: no branch` with explanations). Follows up on #2251, which covered the other two branches in this file.

## Why

A full-suite coverage run flagged these partial branches. Investigating each:

- **Reachable, so tested** — `ApproveView.get_notification_kwargs`: the approving staff member already being one of the student's `current_teachers()`, in which case `affected_users` is not re-extended with the teachers.
- **Unreachable, so excluded** (each cannot be hit because an earlier guard rules it out; a `# pragma: no branch` with a comment is the honest way to keep coverage meaningful):
  - `ApproveView` note/icon elif chain (`skip_button`): `dispatch()` raises `Http404` unless `post_has_valid_button()`, so one of the four buttons is always present and the chain always matches (the no-match fall-through can't happen).
  - `complete()` elif chain (`comment`): the early-exit guard raises `Http404` unless `"complete"` or `"comment"` is in POST.
  - `complete()` auto-approve rank-up guard (`if not submission.do_not_grant_xp`): `mark_approved()` on the immediately preceding line sets `do_not_grant_xp = False` (its `transfer` arg defaults to `False`), so the guard is always true here and the skip-notify arc is unreachable.

## How

- `src/quest_manager/views.py`: three `# pragma: no branch` markers, each with a comment naming the guard that makes the arc unreachable. No behavior change (comments only).
- `src/quest_manager/tests/test_views.py`: `ApproveViewTest.test_approve__approver_who_is_a_teacher_not_re_added_to_affected_users` drives an approval through the tenant-aware `self.client`, with `current_teachers` patched to return the approver, and asserts (via a `notify.send` mock) that the approval notification's `affected_users` stays `[student]`.

## Testing

- Full `quest_manager` suite green (356 tests); `ApproveViewTest` (17 tests) green.
- `coverage` confirms the reachable branch is now covered and the three pragma'd branches are no longer reported as partial. `quest_manager/views.py` has no remaining uncovered branches.
- `ruff check` clean.

## Screenshots

N/A: test coverage plus no-op-branch pragmas; no user-facing or front-end change.

---

### DIFF
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -17,7 +17,7 @@ from django.contrib.messages import get_messages
 from django.contrib.auth.models import AnonymousUser
 from django.contrib.contenttypes.models import ContentType
 from django.db import connection
-from django.test import RequestFactory, SimpleTestCase
+from django.test import SimpleTestCase
 from django.test.utils import CaptureQueriesContext
 from django.urls import reverse
 from django.http import JsonResponse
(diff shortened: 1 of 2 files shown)


tylerecouture:
@coderabbitai review

---

PR #2276, opened 2026-08-06
feat: September scale-up switch and runbook

tylerecouture:
### What?

The seasonal half of the split you asked for on #2082: a capacity increase that is **off by default** and reverted once the spike passes, plus the runbook for performing it.

**Stacked on #2082**, which keeps the permanent guards. Its base is `claude/sept-uwsgi-memory-guards`, so this diff shows only the September pieces. Merge #2082 first and this retargets to `develop` cleanly.

### Why?

You asked to separate settings we want all the time from settings that only exist for September. The permanent ones are steady-state correctness (a worker ceiling the box can honour, an RSS recycle threshold, a Redis cap that will not break the Celery broker) and should never be reverted. These ones are the opposite: they exist to be switched on for a few weeks and switched back off.

### How?

**`UWSGI_EXTRA_ARGS` is the switch.** The compose command appends it to `uwsgi --ini`, and later flags win, so the worker ceiling can be raised for a bigger instance without editing `uwsgi.aws.ini`:

```bash
UWSGI_EXTRA_ARGS="--processes 12"   # on, after the instance bump
#UWSGI_EXTRA_ARGS=                  # off, back to the in-file default of 6
```

**`SEPTEMBER-RUNBOOK.md`** covers what is procedure rather than config: the EC2 and RDS bump, CloudWatch alarms, what to watch during week one, scale-down, and rollback. It is explicit about which values are permanent defaults from #2082 and which are temporary overrides that must be reverted **before** downsizing, since a 1 GB Redis cap on the small box would recreate the original problem.

### Testing?

No Python changed. Verified the switch renders both ways:

| State | Rendered web command |
| --- | --- |
| unset (off) | `uwsgi --ini uwsgi.aws.ini` |
| `UWSGI_EXTRA_ARGS="--processes 12"` | `uwsgi --ini uwsgi.aws.ini --processes 12` |

`REDIS_MAXMEMORY=1gb` still overrides the 512mb default. `ruff check src` clean.

N/A, deployment config and docs.

### Anything Else?

**On the front end switch and auto-adjustment**, since this is where those questions land:

- **Within the ceiling, scaling is already automatic.** `uwsgi.aws.ini` runs the cheaper subsystem: `cheaper = 2` workers stay alive at idle and it forks up to `processes` under load, releasing them afterwards. `processes` is a **cap, not an allocation**. This PR raises the cap; it does not change the mechanism.
- **A UI toggle cannot do this.** uwsgi reads the worker count when the master starts, so changing it requires restarting the app server, and a browser-reachable control that restarts uwsgi is not something I would want to add. The same applies to the instance class and the Redis cap. The `.env` variable plus a restart is the switch.
- **Real elasticity above the ceiling means autoscaling the host**, which this single box cannot do as built: Let's Encrypt certs are bind mounted from the host, Redis is a container on the same box, and deploys are a git checkout plus `docker compose build` driven by systemd. That is an infrastructure project, not a setting, and I have deliberately kept it out of here.

Happy to open an issue for the autoscaling question if you want it tracked rather than dropped.

### Review request
@tylerecouture

- Claude Code (`Bytedeck - infrastructure stack`)

---

## Summary by CodeRabbit

* **New Features**
  * Added optional configuration for increasing production web-worker capacity during traffic spikes.
  * Added guidance for sizing worker memory and maintaining adequate instance headroom.

* **Documentation**
  * Added a production load-spike runbook covering scaling procedures, monitoring, safeguards, rollback steps, and peak-metric collection.
  * Documented worker capacity limits and temporary scaling procedures.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -63,6 +63,15 @@ ADMINS=Admin Name:admin@exmaple.com,Admin2 Name:admin2@exmaple.com
 
 ## CAPACITY / SCALING (prod/staging) ##########
 
+# Extra uwsgi flags appended to the web container's `uwsgi --ini` command (later
+# flags win), so the worker ceiling can be raised without editing
+# src/uwsgi.aws.ini. Unset means the in-file default, which is what you want
+# outside a load spike. Size any override to the instance: processes *
+# peak-worker-RSS must leave headroom for celery + redis + the OS, so the
+# example below suits c5a.xlarge/8 GiB and is NOT a generic value. See
+# production/SEPTEMBER-RUNBOOK.md.
+#UWSGI_EXTRA_ARGS=--processes 8
+
 # Redis memory ceiling (noeviction). Defaults to 512mb. Hitting the cap makes
 # redis refuse all writes (breaking the celery broker), so size it to the
 # instance + working set; raise it (and give the box RAM) before a load spike.
--- a/.env.example.aws
+++ b/.env.example.aws
@@ -75,6 +75,18 @@ REDIS_PORT=6379
 #REDIS_MAXMEMORY=512mb
 
 
+## CAPACITY / SCALING #################################
+
+# Extra uwsgi flags appended to the web container's `uwsgi --ini` command
+# (later flags win), so the worker ceiling can be raised without editing
+# src/uwsgi.aws.ini. The in-file default (processes=6) suits c5a.large; raise
+# it after bumping the instance, and unset it again on the way back down.
+# Size to the instance: processes * peak-worker-RSS must leave headroom for
+# celery + redis + the OS, so the example below suits c5a.xlarge/8 GiB and is
+# NOT generic. See production/SEPTEMBER-RUNBOOK.md.
+#UWSGI_EXTRA_ARGS=--processes 8
+
+
 ## POSTGRES ############################################
 
 # Point this at the managed database endpoint (e.g. the AWS RDS endpoint) in
--- a/docker-compose.prod.aws.yml
+++ b/docker-compose.prod.aws.yml
@@ -28,7 +28,7 @@ services:
                       python manage.py migrate_schemas --shared &&
                       python manage.py migrate_schemas --executor=multiprocessing &&
                       python manage.py collectstatic --noinput &&
-                      uwsgi --ini uwsgi.aws.ini"
+                      uwsgi --ini uwsgi.aws.ini ${UWSGI_EXTRA_ARGS:-}"
     volumes:
       - .:/app
       # - static_data:/app/src/static  # use static dir as a fallback for hardcoded urls hitting app directly
--- a/production/SEPTEMBER-RUNBOOK.md
+++ b/production/SEPTEMBER-RUNBOOK.md
@@ -19,10 +19,10 @@ Measured on the production host and RDS:
 
 | Component | Observed | Implication |
 |---|---|---|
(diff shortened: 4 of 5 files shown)


tylerecouture:
@coderabbitai review

Context: this PR is deliberately stacked on `claude/sept-uwsgi-memory-guards` (#2082), not `develop`, so auto-review skipped it. It is the September half of a split #2082: the permanent guards stayed there, and this carries only the seasonal capacity work (`UWSGI_EXTRA_ARGS` plus `production/SEPTEMBER-RUNBOOK.md`). It will be retargeted to `develop` once #2082 merges.

- Claude Code (`Bytedeck - infrastructure stack`)

---

tylerecouture:
@coderabbitai review

- Claude Code (`Bytedeck - infrastructure stack`)

---

tylerecouture:
@coderabbitai all nine findings were valid and are fixed in `7958d803`. Responding once rather than nine times, since they all land in the same file. Resolving the threads.

| # | Finding | Fix |
|---|---|---|
| §1 RDS row | Flat "connections are not a bottleneck" from one point-in-time reading | Qualified as **not a *current* bottleneck**, called out that it is a summer reading rather than a September one, and added the response if the §5 `DatabaseConnections` alarm fires (re-measure, then PgBouncer or a larger class) |
| §3 dry-run migration | Production dump into staging | Now requires an **anonymized or synthetic** dump, and says why: the tenant schemas hold student PII, so restoring one into staging widens who can read it. If a real dump is unavoidable: scrub, restrict access, delete once the timing is recorded |
| §3 Performance Insights | Obsolete since the console moved to CloudWatch Database Insights in July 2026 | Rewritten to point at **Database Insights**, naming the mode (Standard), retention (7-day default, no charge) and that Advanced costs extra and is only worth it for execution plans |
| §3 memory-hog review | "linked in §7", but §7 has no such link | Names **#2081** directly |
| §4 steps 5 and 6 | `docker compose` run without entering the project dir, though an SSH session lands in `~` | Both now `cd ~/bytedeck` first, with a note about where the app lives |
| §5 permissions | "attach a read-only policy", which cannot create an alarm | Split into two principals: `cloudwatch:PutMetricAlarm` for whoever creates the alarms, `cloudwatch:PutMetricData` on the **agent's** role (managed `CloudWatchAgentServerPolicy`) for `mem_used_percent` / `disk_used_percent`. Added the failure mode: without it the agent runs and silently publishes nothing, so the alarms sit in `INSUFFICIENT_DATA` |
| §5 notification actions | Thresholds listed, but nothing said to wire them to anyone | Added: point them all at one SNS topic, subscribe the on-call person, record the topic and owner here once it exists, and **test it** by temporarily breaching a threshold. I deliberately did not invent an SNS topic name or owner, since neither exists yet; the runbook now asks whoever creates it to fill them in |
| §6 Redis response | Said to raise `REDIS_MAXMEMORY`, contradicting §7's warning that a bigger cap starves the workers | Order is now explicit: give the box room first (host RAM, or drop a worker), raise the cap second, because raising it on an unchanged box takes that memory from the workers and trades a broker stall for an OOM kill |
| §7 Redis check | Bare `redis-cli`, and sequenced before the restart | Now `$C exec -T redis redis-cli ...` (redis publishes no host port in production) and moved to after `systemctl restart`, so the restart and health check happen once |

Two things went in alongside, both mine rather than yours: the file's em dashes are gone (this repo bans them and I had left them in a file I wrote), and the `noeviction` wording now matches the correction already made in #2082, so the runbook no longer says redis refuses *all* writes at the cap.

- Claude Code (`Bytedeck - infrastructure stack`)

---

Issue #1734, opened 2025-01-26
Deck suspended if not renewed (back to trial mode)


tylerecouture:
How does this effect students?  Only allow 5 to log in at a time?

PR #2290, opened 2026-08-08
feat: name the deadline and link the subscription page in deck notices (#1733)

tylerecouture:
### What

The in-app deck status notification was a bare label with no link ("Bytedeck sent a subscription expiry reminder.", maintainer report, 2026-08-08). It now carries the notice's key fact and links to the deck's subscription page:

- Expiry reminders name the governing deadline while it approaches ("this deck's free trial ends on Aug. 15, 2026 (7 days left).") and switch to the grace wording once it passes ("ended on Aug. 3, 2026 and the grace period ends on Sept. 2, 2026.").
- The suspended notice adds the deletion horizon ("...the grace period has run out; without a subscription the deck may be deleted after Aug. 8, 2027.").
- Limit warnings name the seat usage ("5 of 5 current-student seats are used.").

### How

Two layers:

1. **Notifications framework**: `Notification` gains a `target_url` field (an explicit link destination), set via `notify.send(..., url=...)`. Deck lifecycle notices have no natural target model, and without a target object a notification rendered an *empty* anchor on the list page, so there was no way to link one anywhere. `get_url()` prefers the explicit URL over `target_object.get_absolute_url()`, and the list page renders the verb itself as the link text in the url-only shape. Existing notifications are untouched (the field defaults to empty and every current call path leaves it that way).
2. **Deck notices**: `_deliver` appends `_notification_detail(deck, kind)` to the label and passes `url=reverse('decks:subscription')`. The detail sentence follows the governing clock (#1734 B4: the later of the trial and paid dates picks both the deadline and the trial/subscription wording).

One schema migration (`notifications.0002`, a nullable-equivalent `CharField` with a default, no data migration needed).

### Testing

- New framework tests: the explicit url is stored and becomes the destination in both renderers, the verb-as-link-text shape on the list page, and precedence when a target object is also present.
- New delivery test pinning all three expiry-family wordings (approaching, in grace, suspended with deletion horizon) and the stored `target_url`.
- Updated the limit-warning delivery assertions to the enriched verb.
- Full suite + ruff + `makemigrations --check` pass locally.

### Screenshots

Captured from the running `hackerspace` dev deck as the deck owner, real notices generated by `_deliver`.

Before (the reported problem: no date, no link):

before

After (each new row names its date and links to the subscription page; the old bare rows below show the contrast):

after

---
_Generated by [Claude Code](https://claude.ai/code/session_01UAo12812TYt3GJgpZi7Pmn)_


tylerecouture (commented):
Instead of making the entire thing a link, just add: "See your subscription details page." and link the text "subscription details page".

Image

tylerecouture:
Done in 443066a: the notification body is now plain text ending with "See your subscription details page." where only "subscription details page." is the link. Under the hood `Notification` gained `target_link_text` (the anchor text for `target_url`); when it's blank the url-only shape still falls back to linking the verb so a linked notification can never render an unreachable destination. The after screenshot in the description shows the new rendering (same asset URL, updated image).

---

tylerecouture:
@coderabbitai review

---

Issue #41, opened 2015-10-01
Better message pop up when error with comment (fix validation)


tylerecouture:
when you go submit a quest that requires to get your attention to verify that something is working correctly unless you type something it will give you a message saying that there was a error and the file you are trying to submit may exceed the 20MB limit instead of saying that you have to type something in the quest submission form

tylerecouture:
Fixed. 16MB filesize limit.

Issue #40, opened 2015-09-30
Repeating quests appearance time


tylerecouture:
Repeating quests aren't appearing at the proper time if they were returned.  Probably because the re-completion resets the completion time which is what the repeat is based upon.

tylerecouture:
Fixed I think...  Added a new field to submission model: first_completion_date from which repeats are calculated from.

PR #2298, opened 2026-08-09
HOTFIX: fix: prune stale docker images/cache before the deploy build

tylerecouture:
### What?
Add two best-effort prune commands to `production/server-update.sh`, right before `docker compose build`, so each deploy reclaims disk from earlier deploys:

```sh
docker image prune -f    # dangling images (a previous deploy's superseded image)
docker builder prune -f  # BuildKit cache
```

### Why?
The **staging** deploy failed ([Deploy run #41](https://github.com/bytedeck/bytedeck/actions/runs/31287699932/job/93179422217)) at the `celery` image's `RUN python3 -m pip install -r requirements.txt` with:

```
ERROR: Could not install packages due to an OSError: [Errno 28] No space left on device
```

`server-update.sh` runs `docker compose build` on every deploy but never reclaims space, so image layers and BuildKit cache from earlier deploys accumulate on the self-hosted host until a build can't write. This was the **first failure after a long run of green deploys** (deploy #40, ~5h earlier, was fine), which is the signature of disk creeping to full rather than a code regression. It is unrelated to any application change.

### How?
Prune **before** `$COMPOSE build` so a near-full host self-heals on its next deploy (the pruned build cache is usually the bulk of the growth). Safety:
- Only **unused** data is removed: dangling (untagged) images and build cache. **Named volumes (the Postgres data) are never touched** (no `--volumes`, no `docker system prune`).
- Both prunes are **best-effort** (`|| echo "WARN: ..."`), matching the script's existing pattern for non-critical steps (nginx reload, runner setup), so a prune hiccup can't fail an otherwise-good deploy.
- Trade-off: clearing the BuildKit cache each deploy makes the build re-download/rebuild layers (a little slower), in exchange for a deploy that can't run the host out of disk.

### Testing?
Deploy-infra shell script; no application code paths. Validated `sh -n production/server-update.sh` (syntax OK). The prune commands run as part of the next staging deploy of this branch, so merging this exercises the new path directly.

N/A (deploy script only; no front-end or user-visible change).

### Anything Else?
- **A host already full still needs a one-time manual cleanup**, since the very build that would run this new prune can only start once there's room. On the `bytedeck-staging` host: `docker system prune -af` (keeps named volumes) and `docker builder prune -af`, then re-run the failed `deploy-staging` job (the `workflow_run` payload is preserved) or push to `staging` again. After that, this change keeps it from recurring.
- Targeted at `staging` as a `HOTFIX:` per the branch model, since it's the staging deploy path that's broken and it should reach the hosts without waiting on the normal `develop` release cycle. The same prune benefits production, so it rides `staging -> master` with the rest.
- CodeRabbit doesn't auto-review PRs to `staging`, so I'll request a review by comment.

### Review request
@tylerecouture

---

### DIFF
--- a/production/server-update.sh
+++ b/production/server-update.sh
@@ -10,6 +10,14 @@ cd "$(dirname "$0")/.."
 
 COMPOSE="docker compose -f docker-compose.yml -f docker-compose.prod.aws.yml"
 
+# Reclaim disk before building so image layers and BuildKit cache left by earlier
+# deploys can't fill the host and fail the build mid-`pip install` ([Errno 28] No
+# space left on device). Only unused data is removed: dangling images (a previous
+# deploy's now-superseded image) and build cache. Named volumes (the Postgres
+# data) are never touched. Best-effort: a prune hiccup shouldn't fail the deploy.
+docker image prune -f || echo "WARN: docker image prune failed; continuing."
+docker builder prune -f || echo "WARN: docker builder prune failed; continuing."
+
 $COMPOSE build
 
 # Update the web app's systemd unit file


tylerecouture:
@coderabbitai review

(This PR targets `staging`, which CodeRabbit doesn't auto-review, so requesting a pass manually.)

---

PR #2299, opened 2026-08-09
HOTFIX: fix: make nginx follow the web container's address instead of caching it

tylerecouture:
### What?

Stops nginx 502ing after every `systemctl restart bytedeck.com` on staging and production. nginx now resolves the `web` service name at request time rather than once at config load, so it follows the app container to whatever address it currently has.

Also removes the `nginx -s reload` workaround this bug forced into `server-update.sh`, and gives the systemd unit's `ExecStop` the same `-f` flags as its `ExecStart`.

### Why?

`systemctl restart bytedeck.com` recreates the `web` container, and compose hands it a fresh IP. nginx resolved `web` exactly once, when it parsed its config, and held that address for the life of the worker process. After a restart it kept dialling an address nothing was listening on:

```
connect() failed (111: Connection refused) while connecting to upstream,
upstream: "uwsgi://172.31.0.4:8000"
```

Every request 502'd until nginx itself was reloaded. This is a known-in-repo problem that had been papered over rather than fixed: `server-update.sh` carried an `nginx -s reload` commented "sometimes nginx doesn't reconnect to uwsgi after a restart", and the troubleshooting notes told you to re-run it by hand. That is also why deploys mostly looked fine while a bare `systemctl restart` did not: only the deploy path ran the reload.

The reload is a one-shot patch. It fixes the currently-cached address and then goes stale again the next time the container is recreated.

Worth separating out: there is also a legitimate 502 window on every restart while `web` runs `migrate_schemas` across every tenant schema and then `collectstatic` before uwsgi binds `:8000`. That window is expected and is not what this PR is about. It is now documented in the troubleshooting section, and removing it would mean moving migrations out of the container's start command (not attempted here).

### How?

In `nginx/bytedeck.conf.template`:

- Added `resolver 127.0.0.11 valid=10s ipv6=off;` (Docker's embedded DNS). `valid=10s` bounds how long a resolved address is reused, so the lookup is not literally once per request; the distinction that matters is request-time versus config-load-time. Docker's DNS returns no TTL of its own, so without `valid=` nginx would fall back to caching indefinitely.
- Replaced `uwsgi_pass django;` with `set $web_upstream web;` + `uwsgi_pass $web_upstream:8000;`, and dropped the `upstream django` block. **The variable is load-bearing:** nginx only consults the resolver at request time when the passed address contains one. Naming an `upstream` block, or inlining `uwsgi_pass web:8000`, both resolve once at startup. The removed group had a single static member and no balancing, failover or keepalive settings, so nothing else is lost. The commented-out unix-socket alternative that lived in it is preserved as a note, since a socket path cannot go stale and needs neither resolver nor variable.
- `envsubst` in `nginx/Dockerfile` is restricted to the domain placeholder, so `$web_upstream` survives the build untouched (verified below).

In `production/systemd/bytedeck.com.service`: `ExecStop` now names the same file set as `ExecStart`. A bare `docker compose down` picks up `docker-compose.yml` plus `docker-compose.override.yml`, the development override that is present in the checkout on the server, so teardown was reasoning about a different set of services than startup created. Independent of the 502, but it is right next door.

In `production/server-update.sh` and `production/SERVER-README.md`: dropped the reload workaround now that the root cause is gone, and rewrote the troubleshooting entry to distinguish the expected startup window from a stale address, with commands to tell them apart.

### Testing?

No Python changed, so the suite is untouched. This is nginx behaviour, which the Django tests cannot reach, so I verified it directly against the **real rendered config** by building the actual image from `nginx/` twice, once from `staging` and once from this branch, and moving the backend container's address underneath both.

Setup: a container named `web` on a compose-style network, both nginx variants pointed at it, then `web` destroyed and recreated so it landed on a different IP (`172.18.0.2` to `172.18.0.5`), with the nginx processes left untouched, exactly as a restart leaves them.

| | after 1st address change | after `nginx -s reload` | after 2nd address change |
| --- | --- | --- | --- |
| current config (`staging`) | **502**, `uwsgi://172.18.0.2:8000` refused | 200 | **502** again |
| this branch | **200**, no upstream errors | n/a | **200** |

The decisive evidence is from the new backend container's own access log, which shows the request arriving from `172.18.0.4`, the fixed nginx, carrying a raw uwsgi packet:

```
172.18.0.4 - - [09/Aug/2026:03:15:25 +0000] "\x00\xAE\x01\x00\x0C\x00QUERY_STRING\x00\x00\x0E\x00REQUEST_METHOD\x03\x00GET..."
```

The pre-fix nginx (`172.18.0.3`) never appears in that log at all: it was still dialling the dead `172.18.0.2`. Its error line reproduces the reported failure exactly, down to the wording.

Also checked, on this branch's tree:

- `docker build` of `nginx/` succeeds and `nginx -t` passes on the rendered config.
- The rendered `/etc/nginx/sites-available/bytedeck.conf` contains `resolver 127.0.0.11 valid=10s ipv6=off;` and `uwsgi_pass $web_upstream:8000;`, so envsubst left the variable alone.
- `ruff check src` clean.

(The test host has no IPv6, so the `listen [::]` lines were stripped for the test run only. Nothing else in the config was modified.)

N/A, deployment config. No user-visible change beyond the site not 502ing after a restart.

### Deploy notes

- **The nginx image must be rebuilt**, since the config is baked in at build time by `envsubst`. `production/server-update.sh` runs `docker compose build`, so a normal deploy covers it. A restart without a rebuild will not pick this up.
- No new environment variables. `127.0.0.11` is Docker's built-in DNS, always present on a user-defined network.
- After deploying, the check is that these two agree:
  ```bash
  cd ~/bytedeck
  C="docker compose -f docker-compose.yml -f docker-compose.prod.aws.yml"
  $C logs nginx | grep -o 'upstream: "uwsgi://[^"]*"' | tail -1
  $C ps -q web | xargs docker inspect -f '{{range .NetworkSettings.Networks}}{{.IPAddress}} {{end}}'
  ```

### Anything Else?

- Targeted at `staging` as a hotfix because it affects both live environments today. CodeRabbit does not auto-review PRs whose base is not `develop`, so it is being summoned manually.
- This wants to reach `develop` too, or the next `develop` release will reintroduce the cached-address behaviour. Happy to open the matching PR once this lands.
- No `CHANGELOG.md` entry: this is deploy plumbing with no effect on what a teacher or student sees on a deck.
- Follow-up worth considering separately: the migrations-and-collectstatic startup window is a real, if expected, 502 on every restart. Moving migrations to a step before `up -d` would remove it.

### Review request
@tylerecouture

- Claude Code (`Bytedeck - infrastructure stack`)

### DIFF
--- a/production/systemd/bytedeck.com.service
+++ b/production/systemd/bytedeck.com.service
@@ -15,10 +15,6 @@ Type=oneshot
 RemainAfterExit=yes
 WorkingDirectory=/home/ubuntu/bytedeck
 # WorkingDirectory=/usr/share/nginx/hackerspace
-# Both lines name the same file set. A bare `docker compose down` would instead
-# pick up docker-compose.yml plus docker-compose.override.yml (the development
-# override, which is present in the checkout on the server), so teardown would
-# be reasoning about a different set of services than startup created.
 ExecStart=/usr/bin/docker compose -f docker-compose.yml -f docker-compose.prod.aws.yml up -d
 ExecStop=/usr/bin/docker compose -f docker-compose.yml -f docker-compose.prod.aws.yml down
 TimeoutStartSec=0
--- a/production/systemd/bytedeck.com.service
+++ b/production/systemd/bytedeck.com.service
@@ -15,8 +15,12 @@ Type=oneshot
 RemainAfterExit=yes
 WorkingDirectory=/home/ubuntu/bytedeck
 # WorkingDirectory=/usr/share/nginx/hackerspace
+# Both lines name the same file set. A bare `docker compose down` would instead
+# pick up docker-compose.yml plus docker-compose.override.yml (the development
+# override, which is present in the checkout on the server), so teardown would
+# be reasoning about a different set of services than startup created.
 ExecStart=/usr/bin/docker compose -f docker-compose.yml -f docker-compose.prod.aws.yml up -d
-ExecStop=/usr/bin/docker compose down
+ExecStop=/usr/bin/docker compose -f docker-compose.yml -f docker-compose.prod.aws.yml down
 TimeoutStartSec=0
 
 [Install]
--- a/nginx/bytedeck.conf.template
+++ b/nginx/bytedeck.conf.template
@@ -8,11 +8,24 @@
 # source of truth instead of maintaining divergent copies. Do not edit the
 # generated /etc/nginx/sites-available/bytedeck.conf directly.
 
-# the upstream uwsgi component nginx needs to connect to
-upstream django {
-    # server unix:/bytedeck-volume/app.sock; # for a file socket
-    server web:8000; # for a web port socket (we'll use this first)
-}
+# Resolve the `web` service name through Docker's embedded DNS (127.0.0.11) on
+# every request rather than once at startup, so nginx follows the app container
+# to whatever address it currently has.
+#
+# nginx caches a hostname it resolves at config-load time for the whole life of
+# the worker process. Compose gives the `web` container a fresh IP whenever it
+# is recreated (`systemctl restart bytedeck.com` does exactly that), so a cached
+# address goes stale on every restart and nginx keeps dialling the dead one:
+# `connect() failed (111: Connection refused) ... upstream: "uwsgi://<old-ip>:8000"`,
+# i.e. a 502 on every request until nginx itself is reloaded.
+#
+# `valid=10s` bounds how long a resolved address is reused. Docker's DNS returns
+# no TTL of its own, so without it nginx would cache indefinitely again.
+#
+# None of this applies to a unix socket (`uwsgi_pass unix:/bytedeck-volume/app.sock;`,
+# the alternative the commented-out socket_data mounts in the compose files are
+# for): a path cannot go stale, so it needs no resolver and no variable.
+resolver 127.0.0.11 valid=10s ipv6=off;
 
 server {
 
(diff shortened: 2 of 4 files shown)


tylerecouture:
@coderabbitai review

Summoning explicitly: this is a hotfix targeting `staging`, and `.coderabbit.yaml`'s `auto_review` sets no `base_branches`, so it only auto-reviews PRs against `develop`.

Worth a close look at two things in particular:

1. The `set $web_upstream web;` + `uwsgi_pass $web_upstream:8000;` pairing. The variable is what makes nginx consult the `resolver` at request time; without it the name resolves once at config load and the bug returns. Please confirm that reasoning holds for `uwsgi_pass` specifically (not just `proxy_pass`), and that removing the `upstream django` block loses nothing else.
2. Whether `resolver 127.0.0.11 valid=10s ipv6=off;` at `http` scope has any unintended effect on the other server blocks in the same file (the port 8080 redirect and the `default_server` 444 block).

- Claude Code (`Bytedeck - infrastructure stack`)

---

tylerecouture:
Thanks, both answers line up with what I measured, and your wording correction is right: `valid=10s` lets nginx reuse a resolved address for up to ten seconds, so "on every request" overstated it. The distinction that actually matters is request-time versus config-load-time. Fixed in `1c8dad6e`, and I have reworded the PR description to match.

Noting for the record that the review itself was rate limited before it could be submitted, so this PR still has no review verdict. I will re-summon once the window clears.

- Claude Code (`Bytedeck - infrastructure stack`)

---

production/systemd/bytedeck.com.service
```diff
@@ -15,8 +15,12 @@ Type=oneshot
 RemainAfterExit=yes
 WorkingDirectory=/home/ubuntu/bytedeck
 # WorkingDirectory=/usr/share/nginx/hackerspace
+# Both lines name the same file set. A bare `docker compose down` would instead
+# pick up docker-compose.yml plus docker-compose.override.yml (the development
```
tylerecouture: This code comment has no context since the old code is now gone.  None of this block comment is helpful!

PR #2301, opened 2026-08-09
hotfix: convert Stripe SDK objects to plain dicts at every boundary

tylerecouture:
### What

Fixes the webhook 500s caught live during the staging rehearsal (2026-08-09): every Stripe delivery to `/decks/stripe/webhook/` failed with `AttributeError: get` at `event.get('type', '')`.

**Root cause:** stripe-python 15.x objects support `event['id']`-style indexing but are NOT dicts: `.get()` raises `AttributeError`. The webhook view, all event handlers, `sync_from_stripe_subscription`, the checkout reconciler (the "Activating your subscription" poller, also silently affected), and `stripe_backfill_report` are written dict-style, and every test fed plain-dict doubles through the mocked `tenant.billing.stripe` seam, so the mismatch never surfaced until the first real delivery. The failure was uniform across all event types because `event.get('type')` runs before any handler.

**Fix:** a `to_plain_dict()` helper (JSON round-trip via the SDK object's `str()`; passes real dicts through unchanged) converts at every SDK boundary:

- the webhook view, right after signature verification
- `handle_webhook_event`'s entry (defensive, so any caller is safe)
- the subscription retrieve behind the admin "Sync from Stripe" action and the nightly reconcile
- the checkout-session retrieve behind the post-checkout poller
- the backfill report's subscription listing

### Testing

Three new regression tests pin the seams with the SDK's **actual object types** built via `construct_from` (each asserts `hasattr(obj, 'get')` is False, the exact property that broke staging): the webhook view end-to-end with a real `stripe.Event`, the reconciler with a real `checkout.Session`, and the backfill report with a real `Subscription`. All three fail without the fix. Full suite (1900+), ruff, and `makemigrations --check` pass locally; `billing.py` coverage unchanged at 100%.

### Screenshots

N/A (no user-facing change; the visible effect is webhook deliveries returning 200 instead of 500).

### Anything else

After merging, merge `develop` into `staging` to redeploy the rehearsal environment, then Resend one failed delivery from the Stripe dashboard to confirm a 200 plus a `StripeEventLog` row. Stripe's automatic retries will re-deliver the rest.

---
_Generated by [Claude Code](https://claude.ai/code/session_01UAo12812TYt3GJgpZi7Pmn)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Stripe billing, subscription reporting, checkout reconciliation, and webhook processing compatibility with newer Stripe SDK response objects.
  * Preserved handling for existing dictionary-based responses.
  * Ensured valid webhook events are correctly processed, logged, and associated with the appropriate tenant.
* **Tests**
  * Added regression coverage for real Stripe SDK objects, including case-insensitive owner matching and read-only report behavior.

### DIFF
--- a/src/tenant/billing.py
+++ b/src/tenant/billing.py
@@ -314,6 +314,13 @@ def _sync_deck_from_subscription_id(deck, subscription_id):
     configured with only the webhook secret can still absorb events that sync
     straight from their payloads, but it cannot retrieve, and an exception here
     would turn into a 500 that Stripe retries for days (review find on #2110).
+
+    Args:
+        deck (Tenant): The deck whose billing state the subscription drives.
+        subscription_id (str): The Stripe subscription id (sub_...) to retrieve.
+
+    Returns:
+        str: A short human-readable summary for the caller's log/audit trail.
     """
     if not settings.STRIPE_SECRET_KEY:
         return 'STRIPE_SECRET_KEY unset; retrieve skipped'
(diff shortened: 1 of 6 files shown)


tylerecouture:
@coderabbitai review

---

PR #2306, opened 2026-08-09
test: give every tenant test the view helpers and a tenant client by default

tylerecouture:
### What?

`ByteDeckTenantTestCase` now inherits `ViewTestUtilsMixin` and builds its own `TenantClient`, so a tenant test class gets both without declaring anything.

That deletes a lot of repetition:

| | count |
| --- | --- |
| classes that listed `ViewTestUtilsMixin` | 71 |
| imports of the mixin | 20 |
| `setUp` methods that only built a client | 59 |
| `self.client = TenantClient(self.tenant)` lines removed in total | 89 |

Net: 33 files, +321 / -571.

**No test method was added or removed.** The 33 touched files hold 975 tests before and after, and the full suite runs 1965 green either way.

### Why?

Second PR from the test-suite review. Every tenant test class needed the same two lines of ceremony before it could do anything, and the cost was not just noise: a class that forgot the mixin couldn't use `assertRedirectsLogin` / `assert403` / `assert200`, so its author hand-rolled `self.assertEqual(response.status_code, 302)` instead. 220 of the 290 tenant test classes were in that position. Putting the helpers on the base class is the precondition for the assertion cleanup that follows, and for new view tests to reach for the helper rather than the literal.

### How?

**The client is built in `_pre_setup`, not `setUp`,** so it lands before any subclass hook and can't be forgotten. It has to come *after* `super()._pre_setup()`, which assigns Django's plain `Client` from `client_class` and would otherwise overwrite it. `TenantClient` takes the tenant as a constructor argument, so it can't simply be named as `client_class`. Like Django's own `_pre_setup` this is a `classmethod` invoked once per test method, so each test still gets a fresh client and no login session carries into the next test.

**The 71 classes had to stop naming the mixin**, rather than merely being allowed to. `class Foo(ViewTestUtilsMixin, ByteDeckTenantTestCase)` names a mixin ahead of a base that already inherits it, which is an unresolvable MRO:

```python
>>> class Mixin: pass
>>> class Base(Mixin): pass
>>> class Sub(Mixin, Base): pass
TypeError: Cannot create a consistent method resolution order (MRO) for bases Mixin, Base
```

So this is a `TypeError` at import, not harmless duplication, and the sweep is mandatory rather than cosmetic. The mixin's docstring now says so, so the next person who adds it back finds out why. All 71 were `ByteDeckTenantTestCase`-based, so nothing outside this hierarchy lost the mixin.

**One class needs the plain client.** `SyncFromStripeActionTest` drives the tenant admin, which lives on the *public* schema. A `TenantClient` addresses this tenant's own domain, so its requests switch the connection to this tenant's schema, where the public-schema superuser doesn't exist and `force_login`'s `last_login` write fails with `DatabaseError: Save with update_fields did not affect any rows`. It sets a new `tenant_client = False` opt-out, mirroring the existing `reuse_schema = False` one, with a comment saying why.

That class was the only casualty. It surfaced as three failures in a full-suite run, which is also the honest answer to "how risky is this": the sweep is mechanical, but it does change the client type for every class that previously had no client-building `setUp` at all, and the suite is what proves that only one of them cared.

### Testing?

Full suite against a local Postgres 16 + Redis, on this branch rebased onto the merged `develop`:

```
Ran 1965 tests in 361.231s
OK
```

`ruff check src` clean, `manage.py check` clean, `makemigrations --check --dry-run` reports no changes.

Test-count check, to prove the `setUp` deletions didn't take a test with them:

```
test methods in the 33 changed files, before: 975
test methods in the 33 changed files, after:  975
```

N/A: test infrastructure only, no application code changed.

### Anything Else?

Review this as a noise PR: the interesting part is `src/hackerspace_online/tests/utils.py`, and the other 32 files are the mechanical consequence. `ViewTestUtilsMixin` moves above `ByteDeckTenantTestCase` in that file (it has to be defined before the class that inherits it), so git renders it as a delete-and-add; the class body itself is unchanged apart from the added docstring paragraph.

Next in the sequence is the assertion sweep: replacing the ~430 hand-rolled `assertEqual(response.status_code, ...)` call sites with the mixin helpers, now that every class can reach them.

### Review request
@tylerecouture

- Claude Code (`bytedeck - test suite review`)

---

## Summary by CodeRabbit

* **Tests**
  * Centralized tenant-aware test client setup across the test suite.
  * Simplified numerous test classes by removing redundant local setup and shared utility inheritance.
  * Preserved existing authentication, fixtures, assertions, and behavior coverage.
  * Added support for tests that explicitly require public-schema client behavior.

### DIFF
--- a/src/announcements/tests/test_views.py
+++ b/src/announcements/tests/test_views.py
@@ -7,19 +7,18 @@ from django.forms.models import model_to_dict
 from django.urls import reverse
 from django.utils import timezone
 
-from django_tenants.test.client import TenantClient
 from model_bakery import baker
 
 from announcements.forms import AnnouncementForm
 from announcements.models import Announcement
 from comments.models import Comment
-from hackerspace_online.tests.utils import ByteDeckTenantTestCase, ViewTestUtilsMixin
+from hackerspace_online.tests.utils import ByteDeckTenantTestCase
 from siteconfig.models import SiteConfig
 
 User = get_user_model()
 
 
-class AnnouncementViewTests(ViewTestUtilsMixin, ByteDeckTenantTestCase):
+class AnnouncementViewTests(ByteDeckTenantTestCase):
 
     @classmethod
     def setUpTestData(cls):
@@ -31,10 +30,6 @@ class AnnouncementViewTests(ViewTestUtilsMixin, ByteDeckTenantTestCase):
 
         cls.test_announcement = baker.make(Announcement, draft=False)
         cls.ann_pk = cls.test_announcement.pk
-
-    def setUp(self):
-        """Set up a tenant test client for each test."""
-        self.client = TenantClient(self.tenant)
 
     def test_announcement_form__has_unsaved_changes_guard(self):
         """The announcement create and edit forms opt into the unsaved-changes warning, and the guard script is loaded (#192)."""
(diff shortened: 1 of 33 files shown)


tylerecouture:
@coderabbitai review

PR #2307, opened 2026-08-09
docs: tell sessions to build a real test environment, not default to static review

tylerecouture:
### What?
Adds a directive to the top of CLAUDE.md's **Tests and Linting** section telling Claude Code sessions to actually stand up the environment and run the suite, instead of defaulting to a static-only review.

### Why?
New sessions (web/remote especially) tend to open a task with something like *"No test-runnable environment here (no Postgres/Docker/venv), so this will be a rigorous static review"* and then skip running anything, even though the environment can be built. Static reading misses migration state, tenant-schema behavior, and real pass/fail. This makes "build it and run it" the explicit default, and "static only" the documented fallback of last resort.

### How?
A short bold callout right where a session decides how to test, pointing at the same path CI uses:
- `cp .env.example .env` (if absent), `docker compose up -d`, then `docker compose exec -T web python src/manage.py test src` (and `ruff check src`).
- Notes that `TenantTestCase` builds its own test schemas, so no manual `initdb` is needed just to run the suite.
- Notes the first image build takes a few minutes and should be done once up front, not treated as a blocker.
- Fall back to a static-only review only if the environment genuinely can't be brought up, and then state exactly which step failed (e.g. Docker daemon unavailable, image build error) rather than presuming it can't be done.

Placed in the Tests section (rather than the `Claude Code:` convention bullets) because that's where a session forms the "how do I run this" decision, and it sits near the actual commands.

### Testing?
Docs-only; no code paths affected. Verified no em dashes (house style).

N/A (documentation only).

### Anything Else?
A CLAUDE.md directive relies on the session reading and following it. If you want this enforced rather than advised, a `SessionStart` hook that runs `docker compose up -d` (and warms the image build) would guarantee the environment is up before the model starts, independent of whether it reads this section. Happy to add one as a follow-up if you'd like: say the word.

### Review request
@tylerecouture

---

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -27,6 +27,8 @@ The site only works via `http://localhost:8000` (not `0.0.0.0`) because the mult
 
 ### Tests and Linting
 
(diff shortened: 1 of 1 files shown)


tylerecouture:
@coderabbitai review

(The initial auto-review was rate-limited; requesting a pass now that the cooldown has elapsed.)

---

PR #2309, opened 2026-08-09
test: drop the tenant-client setup the shared base now provides

tylerecouture:
### What?

Follow-up to #2306, cleaning up what that migration left behind. Three groups of changes, all in test code:

1. **Removes tenant-client constructions that are now exact duplicates** of the client `ByteDeckTenantTestCase._pre_setup` builds:
   * a client rebuilt inside a `djcytoscape` test body
   * `CooldownTestMixin.setUp` and `CourseViewTestData.setUp`, whose only remaining job was building the client (both `setUp` methods are removed entirely)
   * 14 constructions in `src/hackerspace_online/tests/test_forms.py`
2. **Corrects docstrings that misdescribe what the code does**: five `setUp` docstrings that still described building a client they no longer build, plus the `ResetPasswordViewTests` class and method docstrings, which claimed to cover users who registered without an email when the test actually posts an address no account uses.
3. **Adds class docstrings to twelve migrated test classes** that had none.

### Why?

#2306 gave every `ByteDeckTenantTestCase` a per-test `TenantClient`, which made the per-class and per-test constructions redundant. I removed most of them there but missed these, and CodeRabbit&#39;s review caught the `djcytoscape` one. Checking that finding turned up the rest.

`test_forms.py` is the notable miss: it was not in #2306&#39;s diff at all, because its clients are built at the top of individual test methods rather than in a `setUp`, so the sweep that found the others did not match it.

The docstring changes are the repo&#39;s &#34;comments must describe the code as it is now&#34; rule: a `setUp` whose docstring reads *&#34;Set up a tenant client for each test&#34;* over a body that only toggles a siteconfig flag actively misleads the next reader.

### How?

`_pre_setup` assigns `cls.client = TenantClient(cls.tenant)` once per test method, so an in-test `self.client = TenantClient(self.tenant)` builds an identical client for the same tenant and shadows it to no effect.

The one risk in removing these is a test that deliberately re-creates the client mid-test to drop a login session. I checked all 14 in `test_forms.py` individually: every one sits in the setup phase of its test, before any `force_login`/`login` call, so none were doing that. The single local `client = TenantClient(self.tenant)` (used to obtain a `WSGIRequest` for the allauth login form) now uses `self.client`, which is the same kind of object.

[...]

PR #2333, opened 2026-08-10
test: say where the remaining bare 302 assertions land

tylerecouture:
### What?

Replaces the remaining `assert302` calls that were hiding a real destination with assertions that name it, and turns the unused, deprecated `assertRedirectsAdmin` helper into a working `assertRedirectsAdminLogin`.

### Why?

`assert302` proves a view redirected, not where to. That is fine when the redirect is the endpoint doing its job (the permission umbrellas), but several of these were standing in for behaviour worth pinning:

```python
# Not this student's submission
self.assert302('quests:submission', args=[s2_pk])
self.assert302('quests:drop', args=[s2_pk])
```

That is the authorization boundary between two students' work. It passes whether the view sends the student back to their own quests page or forwards them to a page about someone else's submission, and it passes if the destination silently changes to anything at all. The same for an anonymous request to the tenant admin: a 302 could be the admin login, or it could be a redirect off to the public tenant.

### How?

Each of the following now asserts a destination:

* a student opening **another student's** submission, drop or past-submission page goes back to their own quests page (`assertRedirectsQuests`),
* a teacher **skipping** a submission returns to the approvals queue,
* **unarchiving** a quest lands on the Drafts tab, which is where unarchiving just put the quest,
* **logging out** lands on the login page (`ACCOUNT_LOGOUT_REDIRECT_URL`),
* an anonymous request for the **tenant admin changelist** goes to the admin login carrying the page it asked for in `?next=` (7 call sites),
* `decks:subscription` uses `assertRedirectsLogin` instead of a 302 plus `assertIn('login', response.url)`, which would also have accepted, say, `/blogin-widget/`.

The admin case needs its own helper, because the django admin sends unauthenticated visitors to `admin:login` and not to `settings.LOGIN_URL`, so `assertRedirectsLogin` is the wrong assertion for an `/admin/` url. `ViewTestUtilsMixin` already had that helper: `assertRedirectsAdmin`, called from nowhere, emitting a `DeprecationWarning` telling callers to use `assertRedirectsLogin` instead. It becomes `assertRedirectsAdminLogin` with the warning dropped, a docstring saying when each of the two is right, `reverse('admin:login')` instead of a hard-coded `/admin/login/`, and `urlencode` for the `?next=` value like `login_url_with_next` does.

The `assert302` calls left alone are the ones inside the student and teacher permission umbrellas (`quests:start`, `quests:hide`, `profiles:comment_ban` and friends), where the 302 means the endpoint ran and #2327 covers what it did.

### Testing?

```
Ran 2070 tests in 295.934s

OK
```

`ruff check src` passes.

The new assertions were checked for real: the tenant admin destination was confirmed by running `tenant.tests.test_admin` (71 tests) against `admin:login` rather than assuming it, and `quest_manager.tests.test_views` (230 tests) plus `hackerspace_online.tests.test_auth` and `tenant.tests.test_views` (95 tests) all pass with the named targets.

Run against a real Postgres 16 + Redis dev environment, not a mocked one.

N/A, test-only change with no front-end or user-visible effect.

### Anything Else?

Sixth in the test-suite cleanup after #2306, #2309, #2321, #2322 and #2327. #2322 made the anonymous redirect assertions say where they went; this one finishes that job for the redirects that were not anonymous, and clears the last dead helper the earlier passes left behind.

### Review request

@tylerecouture

---

## Summary by CodeRabbit

- **Tests**
  - Strengthened authentication and admin access checks to verify redirects to the configured login page, including the original destination.
  - Updated quest workflow tests to confirm correct redirects when unarchiving, accessing unauthorized submissions, and skipping submissions.
  - Standardized subscription access tests to use shared login redirect assertions.
  - Improved test helpers for validating admin login redirects.

### DIFF
--- a/src/hackerspace_online/tests/test_auth.py
+++ b/src/hackerspace_online/tests/test_auth.py
@@ -1,6 +1,7 @@
 import re
 from unittest.mock import patch
 
+from django.conf import settings
 from django.contrib.auth import get_user_model
 from django.core import mail
 from django.shortcuts import reverse
@@ -45,7 +46,8 @@ class NonPublicOnlyAuthViewTests(ByteDeckTenantTestCase):
         """
         self.assert200('account_signup')  # ok
         self.assert200('account_login')  # ok
-        self.assert302('account_logout')  # redirect
+        # logging out lands on ACCOUNT_LOGOUT_REDIRECT_URL, which this project points at the login page
+        self.assertRedirects(self.client.get(reverse('account_logout')), reverse(settings.LOGIN_URL))
         self.assertRedirectsLogin('account_change_password')  # login required
         self.assertRedirectsLogin('account_set_password')  # login required
         self.assert200('account_inactive')  # ok
--- a/src/hackerspace_online/tests/utils.py
+++ b/src/hackerspace_online/tests/utils.py
@@ -15,7 +15,6 @@ from model_bakery import baker
 
 import json
 import re
-import warnings
 
 from urllib.parse import urlencode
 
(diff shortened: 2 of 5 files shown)


tylerecouture:
@coderabbitai review

(The automatic review hit the OSS rate limit when this PR opened; the stated cooldown has elapsed.)

---

tylerecouture:
On the one nitpick (Args/Returns sections on `assertRedirectsAdminLogin`): not taking it, for two reasons.

The docstring already documents the parameters in prose, in the same wording the neighbouring helpers use: "Provide any url and path parameters as args or kwargs". Adding `Args: url_name: URL pattern name passed to reverse()` and `Returns: None` restates the signature without telling a reader anything they could not see, and none of the sibling assertions in `ViewTestUtilsMixin` (`assertRedirectsLogin`, `assertRedirectsHome`, `assert200`, ...) carry those sections, so it would be the odd one out.

That also matches the learning recorded on this repo from #2003, which is quoted back in the #2327 thread: parameter and return-value documentation is appropriate "when it conveys meaningful information, not generic filler". The meaningful part of this helper's contract is *which* login page it expects and why, which is what the docstring spends its second paragraph on.

---

PR #2350, opened 2026-08-12
test: cover the TA publish and editor-clearing rules instead of a TODO

tylerecouture:
### What?

Answers the last bare TODO in the quest CRUD tests with four tests, covering both halves of the TA / editor rule in `QuestFormViewMixin.form_valid`.

### Why?

```python
    # TODO
    # TAs should not be able to make a quest published
    # When a quest is published by a teacher, the editor should be removed
```

Both rules are implemented, in `QuestFormViewMixin.form_valid`:

```python
# TA created quests should not be published.
if self.request.user.profile.is_TA:
    form.instance.published = False
    form.instance.editor = self.request.user
# When a teacher makes a form published, remove editor abilities.
elif form.instance.published:
    form.instance.editor = None
```

Neither was covered where it counts.

**The TA half.** `TAQuestForm` only *hides* the fields a TA should not set:

```python
self.fields['published'].widget = forms.HiddenInput()
...
self.fields['editor'].widget = forms.HiddenInput()
```

A hidden input still binds posted data, so the widget is cosmetic: nothing stops a TA from posting `published=on` (or an `editor` of their choosing) by hand. The only thing that does is the view override above. The existing TA tests post `minimal_valid_form_data`, which has no `published` key at all, and `published` is a checkbox, so its absence already means `False`. `assertFalse(new_quest.published)` in `test_quest_create__TA_creates_drafts_and_deletes_own` therefore passed whether or not the guard existed: it was asserting the form default, not the rule.

**The teacher half.** Clearing the editor on publish was covered for *bulk* edit (`test_bulk_edit__publishes_unpublished_quests`) but not for the ordinary "teacher opens the TA's draft, ticks Published, saves" path, which is how it actually happens.

### How?

Four tests, each posting the field the old data left out:

* `test_quest_create__TA_publishing_is_overridden`: a TA posts `published=True` and an `editor` of someone else; the quest is still saved as a draft with the TA as editor.
* `test_quest_update__TA_publishing_their_draft_is_overridden`: same override when a TA saves the draft they own, so publishing stays a teacher's decision.
* `test_quest_update__teacher_publishing_removes_the_editor`: a teacher publishing a TA's draft clears `editor`, and the TA has genuinely lost access (`is_editable(ta)` is False).
* `test_quest_update__teacher_leaving_a_quest_unpublished_keeps_the_editor`: the other side of that `elif`, so the TA keeps the draft they are still working on.

Also factors the repeated three-line TA setup into a `_make_TA()` helper and uses it in the two existing TA tests.

### Testing?

```
Ran 2095 tests in 358.696s

OK
```

`ruff check src` passes.

Checked against broken code, not just passing code:

* disabling the `is_TA` branch fails 3 tests: both new TA tests, and the existing `test_quest_create__TA_creates_drafts_and_deletes_own` (which was catching the editor half, but not the published half).
* disabling the `elif form.instance.published: form.instance.editor = None` branch fails `test_quest_update__teacher_publishing_removes_the_editor`, and nothing else in the class.

`quest_manager/views.py` was restored afterwards; no source file is changed by this PR.

Run against a real Postgres 16 + Redis dev environment, not a mocked one.

N/A, tests only.

### Anything Else?

Tenth in the test-suite cleanup after #2306, #2309, #2321, #2322, #2327, #2333, #2334, #2335 and #2342. #2342 answered the prereq-formset TODOs; this one answers the TA/editor TODO, which was the remaining "we never checked this" note in that file.

If you would rather the hidden fields were removed from `TAQuestForm` outright (so a TA's POST can't even bind them, instead of relying on the view to override), say the word and I will open that as a separate change: it is a behaviour change rather than a test one, so I have kept it out of this PR.

### Review request

@tylerecouture

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured quests created or edited by teaching assistants remain unpublished and retain the teaching assistant as editor.
  * Confirmed teacher publication removes the previous editor assignment.
  * Preserved the teaching assistant’s editor assignment when a teacher leaves the quest unpublished.

* **Tests**
  * Added coverage for quest publishing and editor-assignment scenarios.

### DIFF
--- a/src/quest_manager/tests/test_views.py
+++ b/src/quest_manager/tests/test_views.py
@@ -1871,6 +1871,17 @@ class QuestCRUDViewsTest(ByteDeckTenantTestCase):
             'time_available': "14:30:59",
         }
 
+    def _make_TA(self, username='test_ta'):
+        """Create and return a TA (a student with permission to work on quest drafts).
+
+        Profiles are created automatically by a User post_save signal, so the flag goes on
+        the profile that already exists.
+        """
+        test_ta = User.objects.create_user(username)
+        test_ta.profile.is_TA = True
+        test_ta.profile.save()
+        return test_ta
+
     def test_quest_form__quick_reply_field_below_tags(self):
         """The Quick Reply Text field renders below the Tags field on the quest form (#2114)."""
         self.client.force_login(self.test_teacher)
@@ -1929,9 +1940,7 @@ class QuestCRUDViewsTest(ByteDeckTenantTestCase):
     def test_quest_create__TA_creates_drafts_and_deletes_own(self):
         """TAs can create draft quests (as editor) and delete their own but not others'."""
         # simulate a logged in TA (teaching assistant = a student with extra permissions)
-        test_ta = User.objects.create_user('test_ta')
-        test_ta.profile.is_TA = True  # profiles are create automatically via User post_save signal
-        test_ta.profile.save()
+        test_ta = self._make_TA()
         self.client.force_login(test_ta)
 
         # Can access the Create view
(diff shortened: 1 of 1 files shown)


tylerecouture:
@coderabbitai review

(The automatic pass was skipped by the review limit when this PR opened; the stated cooldown has elapsed.)

---

tylerecouture:
Thanks for the review. Leaving the one nitpick (Args/Returns block on `_make_TA`) as is: the first line already says "Create and return a TA", and `username='test_ta'` is a self-describing default, so a Google-style block would add three lines without adding information. Test helpers in this suite use short prose docstrings, and pushing a docstring-only commit would cost a full CI run and another review pass on an already-approved PR for no behavioural gain. Happy to add it if you would rather the helpers be uniform, @tylerecouture.

---

Issue #2382, opened 2026-08-13
initdb gives the Shared Library deck an unreachable domain on a fresh database


tylerecouture:
Found while setting up a dev deck to screenshot PR #2362. Reproduced on a fresh database.

### Problem

`initdb.setup_shared_library` creates the Library tenant and then only sets its domain on the `not created` path:

```python
library_tenant, created = Tenant.objects.get_or_create(
    schema_name='library',
    defaults={'name': 'Shared Library'},
)

if not created and not library_tenant.domains.filter(domain='library.' + settings.ROOT_DOMAIN).exists():
    library_tenant.domains.create(domain='library.' + settings.ROOT_DOMAIN, is_primary=True)
```

On a fresh database `created` is True, so that block is skipped and the domain comes from the tenant post-save signal instead, which builds it from the tenant's **name**:

```python
domain = f'{instance.name.lower()}.{settings.ROOT_DOMAIN}'   # src/tenant/signals.py:41
```

The name is `'Shared Library'`, so the deck gets `shared library.localhost`, with a space in it. That host can never be reached, so on a fresh install the Shared Library deck is unreachable at `library.<ROOT_DOMAIN>`:

```
$ python src/manage.py initdb
$ # Tenant.objects.all() with their domains:
public     ['localhost']
library    ['shared library.localhost']
```

The explicit `library.<ROOT_DOMAIN>` line only ever runs on a re-run of `initdb` against a database where the tenant already exists, which is presumably why this has not been noticed: a developer who runs `initdb` twice gets a working deck, and one who runs it once does not.

### Fix

Set the domain unconditionally, not just on the `not created` branch. Either drop the `not created` condition, or pass `name='library'` so the signal's derived domain is right in the first place (the display name would suffer, so the first option is probably better).

Worth a test asserting the Library tenant's primary domain after a fresh `initdb`, since the current logic is only correct on the second run.

### Notes

Existing deployments are unaffected: their Library tenant already has a domain, so the signal's `if instance.domains.exists(): return` guard leaves it alone.

- Claude Code (`bytedeck - library import/export`)

PR #2227, opened 2026-07-31
HOTFIX: fix: close sign-up on suspended decks (#1734)

tylerecouture:
### What?
On a suspended deck the sign-up page now renders a "Sign Up Closed" notice (with the owner-only explanation) instead of the registration form, and POSTs create no account. Implemented at the allauth adapter level: `CustomAccountAdapter.is_open_for_signup` returns False while the current deck is suspended, which also covers **Google sign-ups** (the social-account adapter delegates its `is_open_for_signup` to the account adapter). Public and library schemas are unaffected.

This is the develop PR #2222 re-targeted to `staging` as a hotfix at the maintainer's request (2026-07-31): its two commits cherry-picked onto `staging` unchanged, including all the CodeRabbit review fixes already confirmed there (docstring Args/Returns, guard-proving POST test, i18n on the template copy, no-current-deck unit test). #2222 is closed as superseded by this PR. Follow-up commits from review and CI: 4ed6f87 takes CodeRabbit's two docstring nitpicks from this PR's own review, and fbbf034 carries the `test_GFKChoiceField__choices_and_clean` ordering fix (same change as #2230 and develop's copy of the test) because that pre-existing staging flake redded this PR's CI twice; whichever of this PR or #2230 merges first makes the other's copy a clean no-op.

### Why?
Maintainer find on staging: with a deck suspended, a brand-new user could still register and land in a signed-in session. The suspension copy live in production says "only the deck owner can sign in", so the open sign-up front door contradicts it; this closes it now rather than waiting for the develop release cycle.

### How?
* `hackerspace_online/adapter.py`: the `is_open_for_signup` override (one guard for form + social sign-ups), on a class that now carries its required docstring.
* `templates/account/signup_closed.html` (new): rendered by allauth when signup is closed; extends the site's `base_no_sidebar.html` like the other account pages, with translatable copy and the suspension explanation kept conditional.
* `utilities/tests/test_fields.py`: the carried flake fix (an explicit `order_by('pk')` so the choice-order assertion stops depending on Postgres row order).
* No migrations, no settings changes. Re-gated on the `staging` base.

### Testing?
* Full serial suite on this branch (staging base): **1555 tests, OK**, plus `ruff` clean, re-run green after each follow-up commit.
* Tests: suspended deck renders `signup_closed.html` with the owner-only copy and a valid signup POST renders the closed notice and creates no user; unsuspended deck still renders the normal form; the adapter defers to the default outside deck schemas.

### Screenshots (if front end is affected)
[...]

PR #2385, opened 2026-08-13
fix: stop a POST parameter from choosing which deck's schema a request runs against (#2362)

tylerecouture:
### What?

Ports the merged hotfix #2362 from `staging` to `develop`. Cherry-picked from `6725760` with no changes; it applied cleanly on top of #2381.

### Why?

**`develop` still contains the vulnerable code.** #2362 went in as a hotfix because it is a tenant isolation break on live decks, and hotfixes land on `staging` only. `staging` merges *from* `develop`, not back into it, so nothing carries the fix the other way:

```
$ git show origin/develop:src/library/utils.py | head -20
def from_library_schema_first(request):
    ...
    use_schema_name = request.POST.get('use_schema')
    schema_name = use_schema_name if use_schema_name else current_schema_name

$ git merge-base --is-ancestor 6725760 origin/develop && echo YES || echo NO
NO
```

Without this, the next release cycle reintroduces the vulnerability, and it would look like a regression with no obvious cause. This is the same two-PR pattern as #2222 (to `develop`) / #2227 (`HOTFIX` to `staging`) for #1734.

For the vulnerability itself: `from_library_schema_first()` read `use_schema` straight off the POST body and handed it to `schema_context()`, so the requesting browser chose which tenant schema the whole request ran under. Its only caller, `ajax_quest_info`, is `@login_required` but not staff-gated, and schema names are the decks' public subdomains, so an authenticated user on one deck could read another deck's quest content by posting that deck's schema name with a quest pk. Full detail and the reproduction are in #2362.

### How?

The request now only gets to ask a yes/no question, `use_library_schema=1`, and the schema name is resolved server side from `get_library_schema_name()`:

```python
def library_schema_if_requested(request):
    schema_name = get_library_schema_name() if is_library_schema_requested(request) else request.tenant.schema_name
    return schema_context(schema_name)
```

Identical to what is already running on `staging`.

### Testing?

```
src/library + src/quest_manager    464 tests   OK
ruff check src                     clean
```

The 9 tests in `src/library/tests/test_utils.py` come across with the fix: flag parsing, both branches of the schema resolution, and the endpoint itself (a foreign schema name must not return that schema's content, while the supported flag still renders the Library preview).

The cherry-pick interacts with #2381, which landed on `develop` in between and rewrote much of `library/views.py`. Git auto-merged both files and the combined result passes, including #2381's own guard tests.

No visual change; the only front-end edit is the POST key the accordion sends on Library pages. The Library preview rendering correctly over the new wire format was captured from the running app on #2362:

Library quest preview loading on the hackerspace dev deck

### Anything Else?

Worth deciding how hotfixes get back to `develop` in general, since this one needed a human to notice. Two options: a scheduled `staging` → `develop` back-merge, or a convention that every hotfix opens both PRs at once. CLAUDE.md documents the `staging` direction but not the return trip. I have not filed an issue for this yet since it is a process question rather than a code one; happy to if you would like it tracked.

Filed while working on the same area: #2382, a fresh `initdb` gives the Shared Library deck the domain `shared library.localhost` (with a space), so the deck is unreachable on a new install.

### Review request

@tylerecouture

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Library preview requests to securely select Library content without accepting arbitrary schema names.
  * Ensured regular requests remain isolated to the current tenant’s content.
  * Added validation for supported boolean values when requesting Library content.
  * Updated Library previews to use the improved request handling.

* **Documentation**
  * Added endpoint documentation describing request validation, schema isolation, and response behavior.

### DIFF
--- a/src/library/tests/test_utils.py
+++ b/src/library/tests/test_utils.py
@@ -63,7 +63,15 @@ class LibrarySchemaIfRequestedTestCase(LibraryTenantTestCaseMixin):
         self.factory = RequestFactory()
 
     def _request(self, data):
-        """Return a POST request carrying the current tenant, as the middleware would."""
+        """Build a POST request carrying the current tenant, as the middleware would.
+
+        Args:
+            data (dict): form fields to post.
+
+        Returns:
+            HttpRequest: the POST request, with `tenant` set to the test tenant so
+            the schema resolution under test has a deck to fall back to.
+        """
         request = self.factory.post('/', data)
         request.tenant = connection.tenant
         return request
@@ -112,7 +120,15 @@ class AjaxQuestInfoSchemaIsolationTestCase(LibraryTenantTestCaseMixin):
         cls.test_teacher = User.objects.create_user('isolation_teacher', is_staff=True)
 
     def _post_preview(self, data):
-        """POST the quest-preview endpoint for the Library quest's pk with the given data."""
+        """POST the quest-preview endpoint for the Library quest's pk.
+
+        Args:
+            data (dict): form fields to post, e.g. the schema flag under test.
+
+        Returns:
+            HttpResponse: the endpoint's response. 200 with a `quest_info_html`
+            JSON body when a quest was found, 404 when none was visible.
+        """
         return self.client.post(
             reverse('quests:ajax_quest_info', args=[self.library_quest.id]),
             data=data,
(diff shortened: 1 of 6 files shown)


tylerecouture:
@coderabbitai review

(The automatic pass hit the review rate limit; that window has now passed.)

---

PR #2386, opened 2026-08-13
fix: don't bind the fields TAQuestForm hides from a TA (#2384)

### LINKED ISSUE #2384: TAQuestForm only hides `published` and `editor`, so a TA can still post them
### Problem

`TAQuestForm` keeps `published` and `editor` in the form and swaps their widgets for hidden inputs:

```python
self.fields['published'].widget = forms.HiddenInput()
...
self.fields['editor'].widget = forms.HiddenInput()
```

A hidden input still binds posted data, so the widget only hides the control in the browser. Nothing stops a TA from posting `published=on`, or an `editor` of their choosing, by hand.

What actually enforces the rule is the view, in `QuestFormViewMixin.form_valid`:

```python
# TA created quests should not be published.
if self.request.user.profile.is_TA:
    form.instance.published = False
    form.instance.editor = self.request.user
```

So the behaviour is correct today. The concern is that the guard lives one layer away from the form that is supposed to express the restriction, and a future view (or a new code path that saves a `TAQuestForm` without going through that mixin) would silently lose it.

### Suggested fix

Drop `published` and `editor` from the fields `TAQuestForm` binds at all, so a TA's POST cannot set them even in principle, and keep the view override as a belt-and-braces backstop. That is a behaviour change rather than a test change, hence a separate issue.

Worth checking while doing it whether anything relies on those hidden inputs round-tripping their current values on a TA edit.

### Context

Both halves of the view guard are now covered by tests (#2350: `test_quest_create__TA_publishing_is_overridden`, `test_quest_update__TA_publishing_their_draft_is_overridden`), so the current behaviour is pinned and this change would be safe to make. Offered in #2350's description and filed here so it is not lost when that PR's body was squashed away.

tylerecouture:
### What?
`TAQuestForm` no longer binds the quest fields a student TA may not set. They are dropped from the form entirely instead of being swapped for hidden widgets. Closes #2384.

### Why?
A hidden input still binds posted data, so the widget only hid the control in the browser. Nothing stopped a TA from posting those fields by hand.

`published` and `editor` were safe in practice, because `QuestFormViewMixin.form_valid` overrides both for a TA. The other two fields the form hides were not:

| Field | Hidden by `TAQuestForm` | View backstop | A TA's POST could... |
|---|---|---|---|
| `published` | yes | yes | (overridden) |
| `editor` | yes | yes | (overridden) |
| `archived` | yes | **no** | archive a quest |
| `available_outside_course` | yes | **no** | make a quest available with no course enrolment |

So this is not only about moving the guard next to the form that expresses the restriction: two of the four were genuinely settable. The new view tests fail on `develop` for exactly that reason (`archived` came back `True`, and the update case redirected to the quest list because `QuestUpdate.get_success_url` follows the now-archived quest).

### How?
- `TA_RESTRICTED_QUEST_FIELDS` names the four fields, and `TAQuestForm.Meta.fields` is `QuestForm.Meta.fields` minus that set. An unbound field is one a `ModelForm` never writes to its instance, so a posted value cannot reach the quest even in principle.
- `QuestForm.__init__` builds a crispy layout that names three of them, so `TAQuestForm.__init__` removes them from the inherited layout with a small `remove_layout_fields` helper (recursive, since two sit inside the Advanced accordion). Without that, crispy would be asked to render fields the form does not have.
- `QuestFormViewMixin.form_valid` is untouched and still forces `published` and `editor` for a TA, as the belt-and-braces backstop the issue asks for.

On the round-tripping question in the issue: nothing relied on it. Editing keeps whatever the teacher set (an unbound field is left alone), and creating falls back to the model defaults, which is what the hidden inputs were submitting anyway.

### Testing?
`python src/manage.py test src/quest_manager` passes (261 tests in the touched modules). `ruff check src` clean. `src/quest_manager/forms.py` is at 100% statement and branch coverage.

New tests, all of which fail on `develop`:
- `test_TAQuestForm__does_not_bind_the_restricted_fields`: none of the four are on the form
- `test_TAQuestForm__crispy_layout_drops_the_restricted_fields_and_keeps_the_rest`: rendered through the real crispy layout, so a too-eager removal that emptied a nested accordion group would be caught
- `test_TAQuestForm__posted_restricted_fields_cannot_change_a_quest`: a POST claiming all four leaves all four as the teacher set them
- `test_TAQuestForm__saves_the_editable_fields`: an ordinary TA edit still saves
- `test_quest_create__TA_cannot_set_the_restricted_fields_by_post`
- `test_quest_update__TA_post_leaves_the_teachers_settings_alone`: covers both a claimed value and an omitted one (for a bound `BooleanField`, omitting it reads as `False`, which would silently clear a teacher's setting)
- `test_QuestForm__still_binds_the_TA_restricted_fields`: the teacher's form is unaffected

The two pre-existing tests from #2350 that pin the view override still pass unchanged.

No visual change, so there is nothing to show: all four fields rendered as `<input type="hidden">` for a TA, so removing them changes no pixel of the form. That is asserted rather than assumed: the crispy-render test above checks the restricted inputs are gone and that the fields a TA edits (including ones inside the Advanced accordion) still render.

### Anything Else?
The issue names `published` and `editor`; this covers all four fields the form was hiding, since the reasoning is identical for each and the two unnamed ones were the ones actually exposed. Happy to narrow it to two if you would rather keep the change to the letter of the issue.

### Review request
@tylerecouture

---

### DIFF
--- a/src/quest_manager/forms.py
+++ b/src/quest_manager/forms.py
@@ -16,6 +16,39 @@ from tags.forms import BootstrapTaggitSelect2Widget
 from .models import Category, Quest, CommonData
 
 
+#: Quest fields a student TA may not set. A TA works on drafts, so whether a quest is
+#: visible to students (``published``), who may keep editing it (``editor``), whether it is
+#: filed away (``archived``) and whether it escapes course enrolment
+#: (``available_outside_course``) all stay with the teacher.
+TA_RESTRICTED_QUEST_FIELDS = ('published', 'editor', 'archived', 'available_outside_course')
+
+
+def remove_layout_fields(layout_object, field_names):
+    """Remove the named fields from a crispy-forms layout tree, in place.
+
+    Used when a form subclass drops fields the shared layout still names, so crispy is
+    never asked to render a field the form does not have.
+
+    Args:
+        layout_object: a crispy ``Layout`` (or any nested layout object). Objects without
+            a ``fields`` list, such as ``HTML``, are left alone.
+        field_names: an iterable of field names to drop wherever they appear in the tree.
+
+    Returns:
+        None. The layout is modified in place.
+    """
+    fields = getattr(layout_object, 'fields', None)
+    if fields is None:
+        return
+
+    layout_object.fields = [
+        field for field in fields
+        if not (isinstance(field, str) and field in field_names)
+    ]
+    for child in layout_object.fields:
+        remove_layout_fields(child, field_names)
+
+
 class BadgeLabel:
     def label_from_instance(self, obj):
         return f"{str(obj)} ({obj.xp} XP)"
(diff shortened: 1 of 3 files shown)


tylerecouture:
@coderabbitai review

---

tylerecouture:
@coderabbitai review

---

tylerecouture:
@coderabbitai review

---

Issue #2373, opened 2026-08-13
Library: "already exists" conflicts raise PermissionDenied, so users get a bare 403


tylerecouture:
Found during a full audit of the Shared Library import/export system.

### Problem

Four places raise `PermissionDenied` for what is a **conflict**, not an authorization failure:

- `ImportQuestView.post` (`src/library/views.py:226`): quest with this import id already exists locally
- `ImportCampaignView.post` (`:325`): campaign already exists locally
- `ExportQuestView.post` (`:425`): quest already exists in the Library
- `ExportCampaignView.post` (`:538`): campaign already exists in the Library

The confirmation pages disable the submit button in these cases, so the comment in the code says it "shouldn't happen". It happens easily: a stale tab, a back button, two teachers importing the same campaign, or the browser's form resubmission prompt. The user then gets a bare 403 page with no explanation and no way back.

### Fix

Redirect with a `messages.warning` naming the conflicting object and linking to it, the way the rest of the app handles this. The guards themselves should stay: they are the real server-side check behind the disabled buttons.

### Context

Audit context: PR #2362. Related: #2364 (name collisions on import, which 500 rather than 403).

- Claude Code (`bytedeck - library import/export`)

Issue #2372, opened 2026-08-13
Library: pushes and pulls are not atomic, so a failure can leave unreviewed content published


tylerecouture:
Found during a full audit of the Shared Library import/export system.

### Problem

Every push runs `QuestResource.import_data(use_transactions=True)`, which **commits**, and only then unpublishes the result in a separate `save()`:

- `export_quest_to_library` (`src/library/exporter.py:40-56`): import, commit, then `imported_quest.published = False; save()`.
- `export_campaign_to_library` (`exporter.py:102-121`): same, plus a separate save for the campaign row.
- `import_quest_to` / `import_campaign_to` mirror it on the way down.

`published` defaults to `True` on the model, and library quests being pushed are published on their source deck, so the window between those statements is a window in which unreviewed content is live in the Library. An exception in between leaves it that way permanently, which defeats the review gate from #1949.

The notification and the review email are also sent inside the same request with no commit boundary, so a reviewer can be emailed about content whose visibility fix has not landed.

### Fix

Wrap each push and each pull in a single `transaction.atomic()` covering the serialization, the visibility fixes and the campaign update. Fire the notification and email from `transaction.on_commit()` so they only go out once the content is actually in its reviewed-pending state.

### Context

Audit context: PR #2362.

- Claude Code (`bytedeck - library import/export`)

Issue #2371, opened 2026-08-13
Library: the "content pushed" review email arrives as one run-on paragraph


tylerecouture:
Found during a full audit of the Shared Library import/export system. Reproduced by rendering the template through the real send path.

### Problem

`src/library/templates/library/email/content_pushed.txt` (added for #1949 in PR #2039) is written as plain text with bare newlines, but it is treated as HTML: `send_email_message` renders it into `admin/tenant/email/message.txt` (`{{ message|safe }}`), attaches that as the `text/html` alternative, and produces the plain-text body with `textify()`.

HTML collapses bare newlines and there are no `<p>` or `<br>` tags, so both versions of the email arrive as a single block:

```
mrjones shared a campaign from <https://timberline.bytedeck.com> to the
ByteDeck Library: Intro to Python It has been added to the Library but is not
visible to other decks until a Library admin reviews and publishes it. Review
and publish it here: <https://library.bytedeck.com/quests/campaign/3/> —
ByteDeck Library
```

The content name runs straight into the sentence after it ("Intro to Python It has been added..."), which is exactly the part a reviewer needs to read. This is the first impression a Library reviewer gets of a submission.

### Also

- It is the only tenant email that skips the shared `tenant/email/_email_footer.html`.
- The `.txt` extension misrepresents what the file is; every other tenant email template that goes through this path is `.html`.
- It contains an em dash, which CLAUDE.md forbids in user-facing copy.

### Fix

Rename to `content_pushed.html`, wrap the paragraphs in real tags, include the shared footer, and drop the em dash. A test asserting the rendered body contains the content name as its own block would keep it honest.

Same em-dash issue in the two success messages the sharer sees, `src/library/views.py:457` and `:568`.

### Context

Audit context: PR #2362.

- Claude Code (`bytedeck - library import/export`)

Issue #2370, opened 2026-08-13
Library: the campaign import preview shows a blank heading and "[No description provided]"


tylerecouture:
Found during a full audit of the Shared Library import/export system. Reproduced against a running stack.

### Problem

`confirm_import_campaign.html` reads an `object` variable that `ImportCampaignView` never supplies (the view is a plain `View` rendering an explicit context dict containing `category`, not `object`).

Line 24 then makes it worse:

```django
{% with object as category %}
```

`{% with A as B %}` assigns B = A, so this **rebinds** `category` to the empty `object` for the whole block, shadowing the real campaign the view passed in. Everything inside that reads `category.*` gets nothing.

Line 8 (`{% firstof object.title category_campaign_name %}`) has the same problem: neither name is in this view's context, so the heading renders as just " Campaign". And line 23 (`{% elif object.published or request.user.is_staff %}`) only passes because the view is staff-only, so the intended published check never actually runs.

### Evidence

Rendering the page for a campaign titled "Rendered Campaign Title" with short description "DESCRIPTION-MARKER-ABC":

```
[PROBE] status:                             200
[PROBE] title in page:                      False
[PROBE] description in page:                False
[PROBE] "[No description provided]" in page: True
[PROBE] heading region:                     ''
```

So the page a teacher uses to decide whether to import a campaign shows an empty heading and tells them the campaign has no description, whatever it actually says.

### Fix

Drop the `object` references and the `{% with %}` block; the view already passes `category`. Decide what the `published` branch was meant to do and either implement it or remove it.

Worth a template test asserting the campaign title and description appear on the page, since this failed silently for a long time.

### Context

Audit context: PR #2362. Pairs naturally with #2369 (rendering inside the schema context), which touches the same views.

- Claude Code (`bytedeck - library import/export`)

PR #2387, opened 2026-08-13
fix: stop the conflict-copy export carrying local row ids into the Library (#2367)

tylerecouture:
### What?

When a shared campaign contains a quest that is already in the Library under the same `import_id`, that quest gets copied in as a new entry. This rebuilds how that copy is made: it now travels through `QuestResource` like the rest of the export, instead of `deepcopy`.

### Why?

The copy was built with `deepcopy(local_quest)` and `pk = None`, which kept `editor_id`, `specific_teacher_to_notify_id` and `common_data_id`. Those are primary keys that mean something entirely different in the Library's schema, and both outcomes were wrong.

**When the id was absent over there**, `full_clean()` rejected it and the exporting teacher got an unhandled `ValidationError`, so sharing the campaign 500'd:

```
{'specific_teacher_to_notify': ['user instance with id 3 is not a valid choice.'],
 'editor': ['user instance with id 4 is not a valid choice.']}
```

**When the id happened to exist**, nothing complained and the copy silently adopted a stranger's row. A quest whose `CommonData` sat at pk 1 came out attached to the Library's own pk 1:

```
[SM] local CommonData pk 1   = "LOCAL BLURB"
[SM] library CommonData pk 1 = "LIBRARY BLURB"
[SM] clone: Mislink Quest (Exported on 2026-08-12)
       common_data -> LIBRARY BLURB (pk 1)
```

The same applies to `editor` and `specific_teacher_to_notify`, so a shared quest could arrive attributed to a Library user who had never seen it.

### How?

The copies go through `QuestResource().export(...)`, with the `import_id` and `name` columns rewritten per row before importing into the Library. Its `Meta.exclude` drops the three per-deck FKs (`editor`, `specific_teacher_to_notify`, `common_data`), so this path no longer keeps its own list of dangerous FKs in sync as the model grows.

Two things improve as a side effect, because the copies now get the same treatment as every other exported quest:

- they attach to their campaign through the normal `campaign_import_id` column instead of a hand-assigned FK
- they carry their prerequisites, where the hand-rolled version dropped them silently. Prereqs resolve by `import_id`, so a copy links to the Library's existing version of its prerequisite. That is the same matching every other import uses; it is a better answer than the previous "no prerequisites at all", though it is worth knowing it is import-id matching and not deck-aware.

**Correction to an earlier version of this description**, which said `Meta.exclude` covers everything that cannot cross a schema boundary and that tag handling would arrive later with #1792. Both are wrong, and I have since verified it on a real push:

`tags` is not a `Meta.exclude` entry. Taggit's `TaggableManager` reports `get_internal_type() == "ManyToManyField"`, so django-import-export auto-discovers it and builds a `ManyToManyWidget(model=Tag, field='pk')`. Tag primary keys are per-schema, so tags travel as raw pks and get re-resolved against whatever holds those pks in the destination:

```
[setup] tags on the SOURCE deck : {2: 'photography', 3: 'darkroom'}
[setup] tags in the LIBRARY     : {2: 'library-tag-2', 3: 'library-tag-3'}
[RESULT] the quest in the Library is tagged: ['library-tag-2', 'library-tag-3']
```

So tags already "work" in the sense that the column moves, and #1792 is that bug rather than a missing feature. This PR neither causes nor fixes it: the same thing happens on every other push and pull, because the copy path now shares their mechanism. Flagging it because the earlier wording implied this path was clean of cross-schema pk hazards, and it is not.

The naming loop moves out into `build_library_clone_name` and now consults **archived** Library quests. It used `Quest.objects`, whose default manager hides them, while the unique constraint they hold does not care: an archived quest sitting on the candidate name would have tripped the very `IntegrityError` the loop exists to avoid.

### Testing?

Eight new tests. Three of them fail against the old implementation with the exact errors above, verified by temporarily restoring the `deepcopy` behaviour behind the new signature:

```
ERROR: test_export_campaign_and_copy_quests__succeeds_when_the_quest_has_an_editor
       ValidationError: {'specific_teacher_to_notify': ['user instance with id 3 is not a valid choice.'],
                         'editor': ['user instance with id 4 is not a valid choice.']}
FAIL:  test_export_campaign_and_copy_quests__does_not_adopt_a_library_row_at_the_same_pk
       AssertionError:  is not None
ERROR: test_export_campaign_and_copy_quests__copy_does_not_collide_with_an_archived_name
```

The rest cover the copy keeping its content (so dropping the FKs did not gut it), the no-conflicts case being a no-op, `build_library_clone_name` both avoiding taken names and staying inside the field's `max_length`, and the copy path's error wrapping.

The existing test that covers copy naming and publish state across repeat exports still passes unchanged, so the behaviour users see is the same where it was already correct.

```
src/library                                  76 tests   OK
full suite (python src/manage.py test src)   2130 tests  OK (skipped=5)
ruff check src                               clean
```

Coverage: `library/exporter.py` is at 100% branch coverage. It briefly was not, and I said otherwise here before Codecov corrected me: the error-wrapping branch on the copy path had no test, and `226d2cf` adds one to match the equivalent branches on the other two export functions.

No front-end change: this is the server-side copy path behind the existing campaign share button. The user-visible outcome is that sharing a campaign with an already-shared quest stops 500ing, which the tests above cover.

### Anything Else?

Step 3 of the Shared Library audit (after #2362 and #2381). Deliberately **not** in this PR:

- **#2372, atomicity.** Both directions still commit the import and then fix visibility in a separate save, so a failure in between can leave unreviewed content published in the Library. It touches every entry point plus the notification and email, so it is cleaner as its own change; that is next.
- **#2364** (name collision on import still 500s), **#2373** (conflicts still raise `PermissionDenied`), **#2369** / **#2370** / **#2371**.

One thing worth a maintainer's eye: the copies now carry prerequisites, which they did not before. If you would rather they arrive with no prerequisites (on the grounds that a copy is a fresh start), say so and I will strip that column for the copy path specifically.

A wider finding, raised separately with @tylerecouture: this PR routes the copy path *through* `QuestResource` on the grounds that the resource is the safe, shared mechanism. A follow-up audit of the whole Library transfer layer found that the resource is itself the source of several content bugs (the tag pk issue above, prerequisites flattened to one `&`-joined string, `common_data` and `Category.map_order` dropped), and that the tablib `Dataset` it builds never leaves the process. That does not change this PR, which fixes a live cross-tenant defect today, but it does mean this code is likely to be rewritten rather than extended. Nothing to do here.

### Review request

@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Conflicting quests are now copied into the Library with unique names and fresh identifiers.
  * Copied quests remain unpublished and preserve their content.
  * Archived-name conflicts are handled automatically.

* **Bug Fixes**
  * Improved handling of validation and integrity errors during quest imports.
  * Ensured copied quests do not retain invalid references to their original location.

### DIFF
--- a/src/library/exporter.py
+++ b/src/library/exporter.py
@@ -1,6 +1,7 @@
 from datetime import date
-from copy import deepcopy
 from uuid import uuid4
+
+import tablib
 
 from django_tenants.utils import schema_context
 from quest_manager.admin import QuestResource
@@ -9,6 +10,97 @@ from django.core.exceptions import ValidationError
 from django.db import IntegrityError
 
 from .utils import library_schema_context, get_library_conflicting_quests
+
+
+def build_library_clone_name(local_name, taken_names):
+    """Return a name for a clone of `local_name` that is free in the Library.
+
+    Quest names are unique per schema and capped at the model's max_length, so a
+    clone of a quest whose original is already in the Library needs a name of its
+    own. Falls back to numbered suffixes when the dated one is also taken.
+
+    Args:
+        local_name (str): the source quest's name.
+        taken_names (set[str]): every name already spoken for in the Library.
+            Must include archived quests: they still hold their name against the
+            unique constraint even though the default manager hides them.
+
+    Returns:
+        str: a name not in `taken_names`, within the field's max_length.
+    """
+    max_len = Quest._meta.get_field('name').max_length or 50
+    base_suffix = f" (Exported on {date.today()})"
+
+    candidate = local_name[:max_len - len(base_suffix)] + base_suffix
+    counter = 1
+    while candidate in taken_names:
+        full_suffix = f"{base_suffix} #{counter}"
+        candidate = local_name[:max_len - len(full_suffix)] + full_suffix
+        counter += 1
+
+    return candidate
+
+
+def clone_quests_into_library(*, source_schema, quests):
+    """Copy quests into the Library as fresh entries, leaving the originals alone.
+
+    Used when a campaign is shared but some of its quests are already in the
+    Library under the same `import_id`: those get a new copy rather than
+    overwriting what is there.
+
+    The copies travel through `QuestResource`, the same serialization the rest of
+    the export uses, with the `import_id` and `name` columns rewritten per row.
+    Going through the resource is what keeps this safe: `Meta.exclude` drops
+    `editor`, `specific_teacher_to_notify` and `common_data`, which are primary
+    keys that mean something different in the Library's schema and must not
+    travel. It also means the copies pick up campaign and prerequisite handling
+    for free instead of re-implementing a subset of it.
+
+    Args:
+        source_schema (str): the tenant schema holding the quests being copied.
+        quests (list[Quest]): quests loaded from `source_schema`.
+
+    Returns:
+        ImportResult | None: the result of importing the copies, or None when
+        there was nothing to copy.
+    """
+    if not quests:
+        return None
+
+    with schema_context(source_schema):
+        export_data = QuestResource().export(quests)
+
+    with library_schema_context():
+        # Archived quests count: the unique constraint does not care that the
+        # default manager hides them.
+        taken_names = set(Quest.objects.all_including_archived().values_list('name', flat=True))
+
+        rows = export_data.dict
+        visibility_map = {}
+        for row in rows:
+            clone_import_id = str(uuid4())
+            row['import_id'] = clone_import_id
+            row['name'] = build_library_clone_name(row['name'], taken_names)
+            taken_names.add(row['name'])
+            # Copies land unpublished like everything else pushed to the Library
+            visibility_map[clone_import_id] = False
+
+        clone_data = tablib.Dataset()
+        clone_data.headers = export_data.headers
+        for row in rows:
+            clone_data.append([row[header] for header in clone_data.headers])
+
+        try:
+            return QuestResource().import_data(
+                clone_data,
+                dry_run=False,
+                raise_errors=True,
+                use_transactions=True,
+                import_campaign=True,
+                local_visibility_map=visibility_map,
+            )
+        except (ValidationError, IntegrityError) as e:
+            raise ValidationError(f"Failed to copy conflicting quests to library schema: {e}") from e
 
 
 def export_quest_to_library(*, source_schema, quest_import_id):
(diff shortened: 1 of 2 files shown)


tylerecouture:
@coderabbitai review

(The automatic pass hit the review rate limit; that window has now passed.)

---

tylerecouture:
@coderabbitai review

---

tylerecouture:
@coderabbitai review

---

tylerecouture:
Both nitpicks from the review, one taken and one declined.

**Test rename: done in `8b347c0`.** Fair catch, the name overpromised. It creates no archived quest, it passes `taken_names` directly, so it is really about the dated name and the numbered fallback. Renamed to `test_build_library_clone_name__falls_back_to_a_numbered_suffix`. The archived case keeps its own test, `test_export_campaign_and_copy_quests__copy_does_not_collide_with_an_archived_name`, which does create one.

**Narrowing the `taken_names` query: not doing it here.** The suggestion is sound in principle and the review rates it "poor tradeoff" itself, which matches my read:

- The Library currently holds single-digit quests. This loads one column of short strings once per campaign export, which is not a cost worth trading correctness risk for.
- The safe prefix length is genuinely subtle. `build_library_clone_name` truncates to `max_len - len(suffix)`, and the suffix grows with the collision counter, so the prefix would have to be at or below the *shortest* truncation any future counter could produce. Get it wrong and the uniqueness check silently misses a taken name, which is the `IntegrityError` this function exists to prevent. That is a bad trade for a query that is cheap today.
- Library scale is already tracked in #2379, which covers the parts that actually hurt first: the quests tab loads every published Library quest per request and renders a detail container for each, with no pagination or server-side search. If the Library grows enough for this query to matter, that issue is where it should be handled, with the whole page in view rather than one `values_list`.

Happy to revisit if the Library grows past the point where either is theoretical.

---

tylerecouture:
The "Lint with ruff" failure on `8b347c0` is not from this PR.

ruff 0.16.3 landed upstream. `requirements.txt` asks for `>=0.16.2,<0.17`, so CI installs 0.16.3 while `.pre-commit-config.yaml` still pins `v0.16.2`, and the workflow's version-drift guard fails:

```
```

Same mismatch on `develop`, so it hits every open PR right now, not just this one. My commit only renamed a test; `ruff check src` passes locally on both 0.16.2 and 0.16.3.

Fixed in #2394, a one-line hook bump. Once that merges I will re-run CI here.

---

tylerecouture:
@coderabbitai review

---

tylerecouture:
@coderabbitai review

PR #2390, opened 2026-08-13
test: answer the two commented-out notes in the tenant admin tests

tylerecouture:
### What?

Replaces the two blocks of commented-out test code in `tenant/tests/test_admin.py` with tests that run.

### Why?

**The TODO.** `test_save_model__public_tenant_creates_schema` asserted the schema name and then gave up:

```python
# TODO: Not working, can't figure out why?
# When try to use client.get() the context switches back to the public tenant
# WHY!?!?
```

followed by twenty commented-out lines with a `print(connection.schema_name)  # public` next to them. So the one thing an admin cares about after creating a deck, that the deck actually works, was never checked.

It does work, and the reason for the symptom is worth writing down: requests are routed by **hostname**, not by whatever `connection.set_tenant()` was called before them. `TenantMainMiddleware.process_request` starts with `connection.set_schema_to_public()` and re-resolves the tenant from the `Host` header on every request, which is exactly the "switches back to the public tenant" that was observed. `TenantClient` supplies that header from the deck's primary domain, and `tenant_save_callback` creates that domain on save, so the lookup resolves and the schema sticks. With `SHOW_PUBLIC_IF_NO_TENANT_FOUND = True` a deck whose host does not resolve silently stays on public instead of erroring, which is what makes this fail quietly rather than loudly.

**The FIXME.** A commented-out test expected `save_model` to raise on bad names:

```python
# with self.assertRaises(ValidationError):
#     non_public_tenant_bad_name = Tenant(name="Non Public")
#     self.tenant_model_admin.save_model(obj=non_public_tenant_bad_name, ...)
```

That could never have passed. `ModelAdmin.save_model` calls `obj.save()`, and `check_tenant_name` is a **model field validator**, so it only runs through `full_clean()` (which `Tenant.save` does not call). Bad names are rejected one layer up, by the admin form.

### How?

* `test_save_model__public_tenant_creates_a_working_deck` (renamed from `..._creates_schema`): keeps the schema-name assertion, adds that the subdomain keeps the name as typed while the schema converts the dash to an underscore (`non-public.localhost` vs `non_public`), and then uses a `TenantClient` to load the login page and, as a teacher created in the new schema, the quests page, asserting the request resolved to the new deck and the connection is still on its schema.
* `test_clean__rejects_names_that_would_not_work_as_a_subdomain` (new, in `TenantAdminFormTest`): asserts the admin form rejects the two names from the note plus their lower-case variants and the dash cases, each with the specific message it produces. The validator itself is already covered directly in `test_models.py`; this covers the admin path the note was about.
* Drops the commented-out `ValidationError` import that only the removed FIXME needed.

### Testing?

```
Ran 2127 tests in 487.594s

OK
```

`ruff check src` passes.

Checked against broken code, not just passing code:

* deriving the deck's subdomain from the schema name instead of the deck name (`non_public.localhost`) fails the working-deck test on the domain assertion.
* removing the consecutive-dashes rule from `check_tenant_name` fails the form test with `'The name cannot have two consecutive dashes.' not found in ['Invalid string used for the tenant name.']`.

Run against a real Postgres 16 + Redis dev environment, not a mocked one. No source file is changed by this PR.

N/A, tests only.

### Anything Else?

Twelfth in the test-suite cleanup after #2306, #2309, #2321, #2322, #2327, #2333, #2334, #2335, #2342, #2350 and #2352. This clears the last of the TODO/FIXME notes I set out to answer in the test suite.

The working-deck test creates a real schema, so it runs the tenant migrations and initial data (about 30s). That cost was already being paid by the existing test; the added client work is a few seconds on top.

### Review request

@tylerecouture

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant creation verification to ensure new tenant subdomains route correctly and support login.
  * Confirmed tenant connections remain isolated to the appropriate tenant schema.
  * Added validation for invalid tenant names, including spaces, special characters, and malformed dashes.
  * Improved error messaging for invalid tenant names.

### DIFF
--- a/src/tenant/tests/test_admin.py
+++ b/src/tenant/tests/test_admin.py
@@ -17,10 +17,10 @@ from django.contrib.contenttypes.models import ContentType
 from django.contrib.messages.storage.fallback import FallbackStorage
 from django.contrib.sites.models import Site
 from django.core.exceptions import PermissionDenied
+from django.db import connection
 from django.http import HttpResponse
 from django.template.response import TemplateResponse
 from django.test import RequestFactory, override_settings
-# from django.core.exceptions import ValidationError
 from django.urls import path, reverse
 from django.utils import timezone
 
@@ -512,8 +512,13 @@ class PublicTenantTestAdminPublic(ByteDeckTenantTestCase):
 
         mock_add_message.assert_called()
 
-    def test_save_model__public_tenant_creates_schema(self):
-        """save_model on the public tenant creates a new tenant with a sanitized schema name."""
+    def test_save_model__public_tenant_creates_a_working_deck(self):
+        """save_model on the public tenant creates a deck that is actually usable: the schema name
+        is sanitized, the deck answers on its own subdomain, and a teacher can sign in and load a page.
+
+        A schema and a Tenant row are not enough on their own. Requests are routed by hostname, so a
+        deck is only reachable once it also has a TenantDomain, which the save creates.
+        """
         with tenant_context(self.public_tenant):
             non_public_tenant = Tenant(
                 name="Non-Public",  # Not a valid name, but not validated in this test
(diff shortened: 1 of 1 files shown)


tylerecouture:
The red `Lint with ruff` here is not this diff. It fails at the pin-agreement step before any linting happens:

```
```

`requirements.txt` holds a range (`ruff>=0.16.2,<0.17`) that CI resolves at run time, while `.pre-commit-config.yaml` names an exact rev, so publishing ruff 0.16.3 turned every PR red without a commit: #2388's lint passed at 14:48Z and this one failed at 15:21Z. Filed as #2391, with the one-line hook bump that clears it (the pre-commit Dependabot ecosystem should also open that PR on its next daily run).

`ruff check src` is clean on this branch locally, and the other five checks here are green. I have left the fix out of this PR rather than mixing repo-infrastructure into a tests-only change; happy to push the one-liner here instead if you would rather unblock it that way.

---

Issue #2393, opened 2026-08-13
The ruff pin drift guard breaks CI every time ruff publishes a patch release


tylerecouture:
Filed from PR #2392, which unblocks the current instance rather than fixing the shape of the problem.

### Problem

The lint workflow checks that the ruff pinned by the pre-commit hook is exactly the ruff that `requirements.txt` installs:

```bash
hook=$(grep -A2 'astral-sh/ruff-pre-commit' .pre-commit-config.yaml | sed -n 's/^ *rev: *v//p' | head -1)
installed=$(ruff --version | awk '{print $2}')
if [ "$hook" != "$installed" ]; then ... exit 1; fi
```

The guard is worth keeping: without it, the hook and CI can lint with different ruff versions and disagree. But the two sides are specified differently:

* `.pre-commit-config.yaml` pins an **exact** version (`rev: v0.16.3`)
* `requirements.txt` gives a **range** (`ruff>=0.16.3,<0.17`), which pip resolves to whatever the newest matching release is

So the moment Astral publishes 0.16.4, `installed` moves and `hook` does not, and **every open pull request goes red on a lint job unrelated to its diff**. That has now happened at least twice (the current one is in #2392; the range was last synced by #2337).

The failure is also confusing in the moment: the error names your PR, but the cause is an upstream release plus a config file your PR never touched.

### Options

1. **Pin the requirement exactly** (`ruff==0.16.3`) so both sides are exact, and let Dependabot bump it. Dependabot can also update `.pre-commit-config.yaml` if the repo adds a `package-ecosystem: pre-commit` entry, so a single PR could move both together. This makes the guard trivially true and drift a real, reviewed event.
2. **Derive one from the other** rather than asserting equality: install ruff in CI from the pre-commit rev (or generate the rev from the requirement), so there is one source of truth and nothing to keep in sync.
3. Keep the range and accept the periodic breakage, but make the guard's message point at the fix (bump both pins) instead of reading like the PR did something wrong. That is the status quo plus better wording; it does not stop the red builds.

Option 1 is the smallest change and matches how the rest of `requirements.txt` treats tools that must match CI.

### Notes

Whatever is chosen, the `<0.17` cap and its rationale (a new 0.minor can land a new rule set) should survive.

- Claude Code (`Bytedeck - GitHub issues`)

Issue #2391, opened 2026-08-13
Lint CI goes red on every PR whenever ruff publishes a patch release


tylerecouture:
### What happened

Since about 15:00 UTC today every PR fails `Lint with ruff` at the pin-agreement step, with no commit having changed anything:

```
##[error]pre-commit ruff hook is v0.16.2, but requirements.txt resolves to 0.16.3.
Update .pre-commit-config.yaml and the ruff pin in requirements.txt together.
```

Evidence that it is time-based rather than change-based: #2388's lint job **passed** at 14:48Z, and #2390's **failed** at 15:21Z. Nothing about the pins changed in between; ruff 0.16.3 was published, and CI started resolving to it.

### Why

The two sides of the guard are not the same kind of pin:

* `requirements.txt` holds a **range**: `ruff>=0.16.2,<0.17`, and `lint.yml` does `pip install "$ruff_spec"`, which resolves to the newest matching release **at job run time**.
* `.pre-commit-config.yaml` holds an **exact** rev: `rev: v0.16.2`, bumped only when the pre-commit Dependabot ecosystem opens a PR.

So the moment upstream ships any 0.16.x, CI installs it, the hook still names the old one, and the guard fails. The comment on the install step ("requirements.txt stays the single place the version is set") describes the intent, but a range is not a version: it is a promise that is only kept until the next release.

The guard itself is doing its job (a real mismatch between hook and CI is worth catching, per #2348); the problem is that one side floats.

### Fix

Immediate unblock: bump the hook to match the release that is now being installed.

```yaml
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.16.3
```

The pre-commit Dependabot ecosystem should do this on its next daily run, so this may clear itself; it is worth doing by hand if PRs are blocked in the meantime.

Durable fix: pin ruff exactly in `requirements.txt` (`ruff==0.16.3`) so CI installs precisely what the hook names, and both only move when a Dependabot PR lands. That keeps the upper-cap intent (no new rule set arrives unpinned) and makes it stronger: no new rules arrive at all without a commit. The tradeoff is one more Dependabot PR to merge per patch release, and a window where the pip PR and the pre-commit PR are merged separately still trips the guard, which is a real mismatch worth seeing.

### Context

[...]

PR #2394, opened 2026-08-13
ci: bump the ruff pre-commit hook to v0.16.3 to match the requirements pin

tylerecouture:
### What?

One line: `.pre-commit-config.yaml`'s ruff hook goes from `rev: v0.16.2` to `v0.16.3`.

### Why?

**"Lint with ruff" is currently failing on every open PR**, including ones whose diffs are clean. ruff 0.16.3 was released upstream; `requirements.txt` asks for `ruff>=0.16.2,<0.17`, so CI now installs 0.16.3 while the pre-commit hook was still pinned to v0.16.2, and the workflow's drift guard trips:

```
Installing ruff>=0.16.2,<0.17
Successfully installed ruff-0.16.3
Update .pre-commit-config.yaml and the ruff pin in requirements.txt together.
```

Nothing in the repo changed: the same mismatch is present on `develop`, so this is not specific to any branch.

```
$ git show origin/develop:.pre-commit-config.yaml | grep -A1 ruff-pre-commit
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.16.2
$ git show origin/develop:requirements.txt | grep -m1 '^ruff'
ruff>=0.16.2,<0.17
```

The guard is doing exactly what it was added for: the hook and CI must lint with the same version so they cannot disagree about what passes. The fix is to move the hook to the version the pin already resolves to.

### How?

Bumped the hook rev only. The `requirements.txt` pin is left as-is: `>=0.16.2,<0.17` already permits 0.16.3, and the comment on it explains the cap is there so a new rule set cannot land unpinned, which a patch release does not do.

### Testing?

Checked that 0.16.3 does not find anything new in this codebase, so this really is version alignment and not a lint change in disguise:

```
$ pip install ruff==0.16.3 && ruff check src
All checks passed!
```

Full suite is unaffected; nothing outside the pre-commit config changes.

N/A, CI configuration only.

### Anything Else?

Found while driving #2387 to green: its "Lint with ruff" job failed on a commit that only renamed a test, which is what surfaced the drift. #2385 and every other open PR will hit the same thing until this lands, so this is worth merging ahead of them.

If you would rather the two versions could never drift again, the alternative is pinning `requirements.txt` to an exact `ruff==` and bumping both together on purpose, which trades the automatic patch pickup for never seeing this failure. Happy to open an issue for that if it is the direction you want; I have not filed one since the current setup is a deliberate choice and this is its intended failure mode working correctly.

### Review request

@tylerecouture

---

## Summary by CodeRabbit

* **Chores**
  * Updated the automated code-quality checks to use the latest Ruff pre-commit revision.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,7 +23,7 @@ repos:
   # Keep this rev in step with the ruff pin in requirements.txt, so the hook and
   # CI lint with the same version and cannot disagree about what passes.
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.16.2
+    rev: v0.16.3
     hooks:
       - id: ruff
         args: [ "--fix" ]


tylerecouture:
@coderabbitai review

---

tylerecouture:
Closing this as a duplicate of #2392, which fixes the same red lint job and was opened an hour earlier. I did not find it before opening this one.

#2392 is also the better change: it bumps the `requirements.txt` floor to `>=0.16.3` alongside the hook rev, where this PR bumps the hook only. Leaving the floor at `>=0.16.2` allows the mismatch to reappear from the other side (a fresh venv resolving to 0.16.2 while the hook names v0.16.3), so moving both together is the more complete fix.

Two corrections to what this description says, for anyone reading back:

* "Happy to open an issue for that if it is the direction you want; I have not filed one since..." is stale. Two issues already track exactly this: #2391 and #2393, both filed before this PR existed, and #2393 lays out the options (pin `ruff==` exactly and let Dependabot move both, or derive one pin from the other) better than my aside did.
* The claim that this is "worth merging ahead of" #2385 and #2387 still holds, but it applies to #2392 now, not to this PR. Those two are red on the same drift and will go green once #2392 lands.

The branch is left in place in case #2392 stalls; nothing here is needed otherwise.

---

PR #2425, opened 2026-08-14
fix: keep file answers when a submission fails validation (#2165)

### LINKED ISSUE #2165: File answers are silently lost when a submission fails validation on re-render
Found during a self-review of the submission-questions feature (#1304, PRs #2085/#2099).

### Problem
File answers upload only at submit (autosave handles text only). If the answer formset fails validation for any reason — e.g. a *different* required text answer left blank — the page re-renders and every `ClearableFileInput` is empty, because browsers never repopulate `<input type=file>`. There is **no notice** that the previously chosen file was dropped. Consequences:
- Required file question: a confusing second-round "You must upload a file…" error even though the student did attach one.
- Optional file question: nothing flags it — the resubmit succeeds and the answer publishes as "(no answer)" though the student believes they attached a file.

### Evidence
- `src/quest_manager/views.py` re-renders the bound formset on invalid; `src/questions/forms.py` file field; nothing in `submission.html` warns about dropped files.

### Fix
On re-render after formset errors, surface a clear warning next to file fields (or in the non-form-error area): "Files can't be kept after a failed submit — please re-select your file." Ideally, persist accepted uploads to the draft row so they survive a re-render (larger change).

### Notes
Depends on #2099. The main `attachments` field has the same underlying browser behavior (pre-existing), but questions make it consequential because answers are marked.

tylerecouture:
### What?
Keeps a student's uploaded file answer when the answer formset fails validation, instead of silently dropping it.

Closes #2165

### Why?
File answers upload only at submit (autosave handles text only), and a browser never repopulates a file input. So if the formset failed validation for *any* reason, most commonly a different question's required text answer being blank, the page came back with every file field empty and nothing to say the upload was gone:

- **Required file question:** a second-round "You must upload a file for this type of question" error, even though the student did attach one.
- **Optional file question:** nothing flags it at all. The student fixes the text answer, resubmits, and the answer publishes empty while they believe their file went with it.

The second case is the quiet one: the work is marked with a missing attachment and neither the student nor the marker knows why.

### How?
- **`questions/utils.py`**: new `save_draft_file_answers(question_formset, uploaded_files)`, which persists the file answers that validated onto their draft rows and returns how many it saved. Text answers already survive a round trip via autosave; this gives files the same treatment. Deliberately narrow:
  - only fields present in `request.FILES` this time, so a later attempt that leaves the input alone (as browsers force it to) cannot wipe the file already kept;
  - only uploads that reached `cleaned_data`, so a file rejected for type or size is not stored and its error still stands on the retry;
  - never a published row, guarded defensively with a `# pragma: no cover` and a note, since the formset is only ever built over this submission's own draft rows.
- **`quest_manager/views.py`**: call it on the invalid path before re-rendering, and tell the student their file was kept so they know not to re-attach it.
- **`submission.html`**: show the file currently held against a question, linked, using the same `filename` filter the marking display uses. The input itself still renders empty (nothing can change that), so without this the student has no way to tell the upload survived.

### Testing?
Three new tests in `CompleteWithQuestionsTest`, covering each branch of the new helper:

- `test_complete__uploaded_file_survives_a_failed_submit`: file kept on the draft row and shown on the re-rendered page.
- `test_complete__rejected_file_is_not_saved`: a `.txt` sent to an image-only question is not stored.
- `test_complete__file_field_left_alone_keeps_the_saved_file`: the second attempt fixes the text, leaves the file input empty, and the kept file survives and publishes with the completion.

Verified test-driven: with the fix reverted, the first test fails with `<FieldFile: None> is not true : the attached file was dropped on re-render`.

Local run on a Python 3.12 venv against Postgres 16:
- `python src/manage.py test quest_manager questions` -> **523 tests, OK**
- `ruff check src`, `manage.py check`, `makemigrations --check --dry-run` -> all clean
- `coverage` on `src/questions/utils.py` -> **100%**, branches included

Captured from the real `hackerspace` dev deck as `student1`, on a quest with a required short answer, a required long answer, and an optional image question. The student attaches a file, leaves the required answers blank, and submits.

**Before** (the file is gone, with nothing to indicate it)

before, file question after a failed submit

**After** (the file is held against the question, and the input is clearly optional now)

after, file question after a failed submit

**After** (the notice at the top of the page)

after, kept-file notice

### Anything Else?
The issue also notes that the main `attachments` field on the comment form has the same underlying browser behaviour. That is pre-existing and outside the question flow, so I left it alone here rather than widening the change; happy to take it as a follow-up if you want the comment attachments kept too.

While setting the dev deck up for these screenshots I had to run `migrate_schemas`, since the seeded `hackerspace` schema predated several `develop` migrations. Nothing to do with this PR, just noting it in case other sessions hit the same thing.

### Review request
@tylerecouture

---
_Generated by [Claude Code](https://claude.ai/code/session_012Ptw8gWroqLnsLWieBaY67)_

## Summary by CodeRabbit

* **New Features**
  * Previously saved response files are now shown with download links and filenames.
  * Clear guidance explains that selecting a new file replaces the existing upload.

* **Bug Fixes**
  * Valid file uploads are preserved when another answer prevents submission.
  * Previously saved files remain intact when the file field is left unchanged.
  * Rejected uploads are no longer saved.

### DIFF
--- a/src/quest_manager/templates/quest_manager/submission.html
+++ b/src/quest_manager/templates/quest_manager/submission.html
@@ -2,6 +2,7 @@
 {% load static %}
 {% load crispy_forms_tags %}
 {% load question_tags %}
+{% load comment_tags %}
 
 {% block head %}
     {{ submission_form.media }}
@@ -123,6 +124,16 @@
                 {{ form.question.instructions|unwrap_p }}
               </div>
               {% crispy form %}
+              {% comment %} A file kept from an earlier submit that failed validation: the file input
+                 above always renders empty (browsers never repopulate it), so without this the
+                 student has no way to tell their upload survived, and would re-attach it (#2165). {% endcomment %}
+              {% if form.instance.response_file %}
+                <p class="help-block">
+                  <i class="fa fa-paperclip fa-fw"></i> Attached:
+                  <a href="{{ form.instance.response_file.url }}" target="_blank">{{ form.instance.response_file|filename|safe }}</a>
+                  <small>(already saved: only choose a file above if you want to replace it)</small>
+                </p>
+              {% endif %}
             </div>
           {% endfor %}
           </div>
--- a/src/quest_manager/views.py
+++ b/src/quest_manager/views.py
@@ -31,7 +31,7 @@ from comments.models import Comment, Document
 from comments.sanitize import sanitize_comment_html
 from questions.forms import QuestionSubmissionFormsetFactory
 from questions.models import QuestionSubmission, QuestionType
-from questions.utils import sync_draft_question_submissions
+from questions.utils import save_draft_file_answers, sync_draft_question_submissions
 from courses.models import Block, CourseStudent
 from library.utils import is_library_schema_requested, library_schema_if_requested
 from notifications.signals import notify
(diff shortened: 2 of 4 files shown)


tylerecouture:
@coderabbitai review

---

Issue #2396, opened 2026-08-13
Library: a shared campaign loses its quest-map placement (Category.map_order is not carried)


tylerecouture:
### Problem

`Category.map_order` does not travel through the Shared Library. A teacher who arranged their campaigns on the quest map loses that arrangement: every campaign pushed to the Library, and every campaign pulled into another deck, arrives at `map_order = 0` and the map's layout engine places it wherever it likes.

Verified on a real round trip (deck to Library to a deck that has never seen the campaign):

```
FIELD                 1. SOURCE DECK    2. IN LIBRARY    3. IMPORTING DECK
CAMPAIGN.map_order    7                 0                0
CAMPAIGN.icon         'icons/campaign.png'  'icons/campaign.png'  'icons/campaign.png'
```

The value is lost on the **push**, so it cannot be recovered on the pull.

### Cause

The campaign is not exported as a model. It is smuggled through four flat columns declared on `QuestResource` (`src/quest_manager/admin.py:105-108`: `campaign_title`, `campaign_icon`, `campaign_short_description`, `campaign_import_id`) and rebuilt on the far side from those four values only:

* `QuestResource.generate_campaign` (`src/quest_manager/admin.py:279-284`) constructs `Category(title=..., icon=..., short_description=..., import_id=...)`
* `library/exporter.py:169-175` constructs the same set plus `published`

`map_order` is in neither, and `grep map_order src/library/ src/quest_manager/admin.py` returns nothing.

`Category.map_order` was added on 2026-07-18 in `971cf450` (#2016), a commit that touched neither `admin.py` nor `src/library/`. So it has silently not travelled since it was introduced. Nothing failed, nothing was reported.

### Why this is worth more than a one-line fix

The campaign travels under a hand-written field list, which is exactly the failure mode that produced this: adding a field to `Category` does not make it travel, and nothing anywhere says so. Four days after `map_order` landed, `Quest.quick_reply` (`fbdd1a38`, #2103) was added and *did* travel, because `Quest` goes through `Meta.exclude` (a denylist) rather than a hand-written list. Same root cause in both cases: the developer adding the field had no reason to know the Library existed.

So the minimal fix is to add `map_order` to both construction sites. The fix worth having is a test that pins the set of fields that travel for **both** `Quest` and `Category` against a reviewed constant, so that adding a field to either model fails CI until someone records a decision about whether it should be shared. Without that, the next field added to `Category` repeats this exactly.

[...]

Issue #2367, opened 2026-08-13
Library: the conflict-clone export carries local row ids into the Library schema


tylerecouture:
Found during a full audit of the Shared Library import/export system. Both failure modes reproduced against a running stack.

### Problem

When a campaign contains a quest that already exists in the Library, `export_campaign_and_copy_quests` (`src/library/exporter.py:180-205`) clones it with `deepcopy(local_quest)`, clears the pk, and saves into the Library schema. The clone keeps `editor_id`, `specific_teacher_to_notify_id` and `common_data_id` intact: integers that mean something completely different in the other schema.

Two outcomes, both wrong:

**The id does not exist in the Library schema.** `full_clean()` validates the FK and the export dies with an unhandled `ValidationError`, so the exporting teacher gets a 500:

```
[PROBE] export raised: ValidationError
        {'editor': ['user instance with id 4 is not a valid choice.']}
```

**The id does exist.** No error, and the clone silently adopts an unrelated row belonging to the Library deck:

```
[SM] local CommonData pk 1   = "LOCAL BLURB"
[SM] library CommonData pk 1 = "LIBRARY BLURB"
[SM] clone: Mislink Quest (Exported on 2026-08-12)
       common_data -> LIBRARY BLURB (pk 1)
```

The same applies to `editor` and `specific_teacher_to_notify`, which means a shared quest can end up attributed to a Library user who has never seen it.

The clone path also drops tags and prerequisites, which the non-clone export path at least attempts to carry.

### Also in this function

The name-uniqueness loop at `exporter.py:196` uses `Quest.objects`, whose default manager excludes archived quests:

```python
while Quest.objects.filter(name=potential_name).exists():
```

So an archived Library quest holding the candidate name still trips the unique constraint the loop exists to avoid. It should use `Quest.objects.all_including_archived()`.

### Fix

Build the clone from an explicit field allow-list rather than `deepcopy`, nulling every FK that cannot cross a schema boundary, and route it through the same `QuestResource` serialization the rest of the export uses so it inherits campaign, prereq and (once #1792 lands) tag handling instead of re-implementing a subset. Fix the archived-quest gap in the naming loop at the same time.

### Context

Introduced with #1884 / PR #1890. Audit context: PR #2362.

- Claude Code (`bytedeck - library import/export`)

PR #2462, opened 2026-08-16
feat: warn the sharer that a General Info block will not travel (#2398)

### LINKED ISSUE #2398: Library: a shared quest arrives without its CommonData ("General Info") block
### Problem

`Quest.common_data` does not travel through the Shared Library. A quest that relies on a shared preamble ("how to record and submit a video", "lab safety rules", a marking rubric used across a campaign) arrives on the destination deck missing that entire panel, and its own instructions may refer to text that is no longer on the page.

Nothing warns the teacher on either side: not on the push confirmation screen, not on the import confirmation screen, and not in the success message.

Verified on a real round trip (deck to Library to a deck that has never seen the quest):

```
FIELD          1. SOURCE DECK              2. IN LIBRARY    3. IMPORTING DECK
common_data    <CommonData: Shared Preamble>    None             None
```

Lost on the **push**, so it cannot be recovered on the pull.

### Cause

Two things, and the second is why this is not a one-line fix.

1. `common_data` is listed in `QuestResource.Meta.exclude` (`src/quest_manager/admin.py:113`), so it is neither exported nor imported, and nothing reconstructs it.

2. **`CommonData` has no `import_id`.** `Quest` and `Category` both carry a UUID `import_id` that gives them a stable identity across schemas; `CommonData` (`src/quest_manager/models.py:182-188`) has only `title`, `instructions` and `active`. So even if the exclusion were lifted, there is no key to match a `CommonData` row in one schema to the same row in another, and the raw FK id is meaningless across schemas. Copying the id is exactly the bug #2367 fixed for `editor` and `specific_teacher_to_notify`: a quest whose `CommonData` sat at pk 1 came out attached to the Library's own pk 1, a completely different block of text.

Excluding it is therefore the *correct* behaviour given the model as it stands. The gap is that the content silently does not arrive.

### Options

1. **Give `CommonData` an `import_id`** and transfer it like `Category`: match by import_id on the far side, create it if absent, reuse it if present. Most faithful, and consistent with how the other shared models already work. Needs a migration and a decision about what happens when the destination already has a `CommonData` with the same title but different text.
2. **Inline the text on export.** Copy the `CommonData.instructions` into the quest's own `instructions` when it is shared, so nothing is lost and no new identity is needed. Simple, but it silently un-shares the block: editing the preamble on the source deck no longer updates the shared copy, and re-sharing would duplicate the text.
3. **Tell the sharer it will not travel** and leave the behaviour alone. This is the cheapest, and it is what #2374 is for. It does not fix the content loss, but it stops the loss being a surprise.

Option 3 is worth doing regardless, since it is true of several things (prerequisites pointing outside the shared set, tags, campaign map placement). Option 1 is the real fix.

### Context

Found during the Shared Library transfer-layer audit. Same family as #2162 (submission questions), #1915 (competency links) and #2375 (badge prerequisites): related content that a quest depends on and that a transfer silently leaves behind. #2374 covers telling the user. #2378 is where the field contract should live.

The characterization test `test_round_trip__drops_common_data` in `src/library/tests/test_transfer_contract.py` pins the current behaviour and names this issue, so whichever option is chosen, that test is where the change gets recorded.

tylerecouture:
### What?

Sharing a quest that uses a **General Info** block (`CommonData`) now tells the sharer that the block stays behind, naming it. Nothing else about the transfer changes: the block still does not travel, and the copy in the Library still arrives without it.

Also records two decisions you settled on the issues, in the characterization tests that were waiting on them:

* the General Info block stays behind on purpose (#2398), rather than "assert it arrives once this is fixed";
* a campaign's `map_order` is deliberately deck-relative (#2396), so it needs no warning at all: nothing was lost, since the sharer's placement was never the importer's to receive.

Closes #2398.

### Why?

> Tell the sharer it will not travel. (#2398)

`CommonData` is the shared block of text several quests can point at: a marking rubric, lab safety rules, how to record and submit a video. It has no `import_id`, so there is no key that means the same block in another deck's schema, and copying the raw id would attach the quest to whatever text happens to sit at that id on the far side. That is exactly the bug #2367 fixed for `editor` and `specific_teacher_to_notify`.

So leaving it behind is the correct behaviour, and this PR does not change it. The gap was that **nobody was told**: the quest arrived on another deck missing a whole panel its own instructions may refer to ("follow the safety rules above"), and the loss was invisible on both sides. The sharer is the one who can act on it, which is the same principle as the other Library losses: content is a self-contained package, and what could not travel is the sharer's business.

### How?

Follows the existing loss-reporting path rather than adding a new one:

* `snapshot_quest` carries `common_data_title` (the title only: the block itself deliberately does not travel).
* `write_quests` aggregates the titles, deduplicated, into a new `TransferResult.dropped_common_data`. `export_campaign_and_copy_quests` unions the exported and cloned halves, as it already does for the other loss fields.
* `warn_sharer_about_dropped_common_data` is wired into both share views, beside the unmet-prereq and archived-quest warnings.

One block or several reads the same way, matching `warn_sharer_about_skipped_quests`:

> The shared General Info 'Lab Safety Rules' does not travel to the Library, so the copy there arrives without that panel. Paste the text into the quest instructions and share it again if this info is required in the quest.

> Shared General Info does not travel to the Library, so the copies there arrive without those panels: 'How to Record and Submit a Video', 'Lab Safety Rules'. Paste the text into the quest instructions and share them again if this info is required in the quests.

The fix is offered as a condition rather than an instruction, per review: pasting the text in is only worth doing if the quest actually needs it. The message names the block rather than the quests, because in a campaign one block is usually shared by several of them.

### Testing?

Four tests in `LibrarySharerWarningTests`: a quest whose block is named, a quest with no block that stays quiet, a campaign where one block shared by two quests is named exactly once, and a campaign where two different blocks are named in a single message.

```
src/library                                   154 tests   OK
full suite (python src/manage.py test src)   2430 tests   OK
```

Run after merging the latest `develop` in, which was needed: #2467 landed submission questions in `snapshot_quest` and conflicted with the `common_data_title` key added here. Both keys are now in the snapshot and the whole library suite passes with them together.

Captured from the running `hackerspace` dev deck, signed in as the deck owner, sharing a real quest that uses a real General Info block.

**Before:** sharing a quest whose instructions depend on 'Lab Safety Rules'. The push succeeds and says nothing about the block.

Before: only the success message

**After:** the same share, on this branch.

After: the success message plus a warning naming the block

**After, campaign with two different blocks:** both are named in one message.

After: a campaign share naming two General Info blocks

### Anything Else?

**The real fix is still open, and now tracked as #2463.** This is option 3 from #2398 ("tell the sharer"), which stops the loss being a surprise but does not stop the loss, and the workaround it suggests (paste the text into each quest) un-shares the block. Option 1, giving `CommonData` an `import_id` and transferring it like `Category`, is filed as #2463 with the one decision it needs written down: what to do when the destination already has a block with the same title but different text.

**#2396 needs no code.** Your answer ("the importer will place it how they want") is already what the transfer does, so the only thing missing was the record of the decision, which this PR puts in the characterization test.

### Review request

@tylerecouture


src/library/views.py
```diff
@@ -153,6 +153,40 @@ def redirect_failed_import(request, error, redirect_to):
...
+    """
+    if not dropped_common_data:
+        return
+
+    names = ', '.join(f"'{title}'" for title in dropped_common_data)
+    messages.warning(
+        request,
+        f"The shared General Info {names} does not travel to the Library, so the copy "
+        "there arrives without that panel. Paste the text into the instructions of the "
+        "quests that need it and share again."
```
tylerecouture: ...and share it again if this info is required in the quest."

tylerecouture: Taken, in `dd4ecf5`. Making it conditional is better: pasting the text in is only worth doing if the quest actually needs it, and my version read like a chore you always have to do.

Both branches now end that way:

> The shared General Info 'Lab Safety Rules' does not travel to the Library, so the copy there arrives without that panel. Paste the text into the quest instructions and share it again if this info is required in the quest.

> Shared General Info does not travel to the Library, so the copies there arrive without those panels: 'How to Record and Submit a Video', 'Lab Safety Rules'. Paste the text into the quest instructions and share them again if this info is required in the quests.

"the quest instructions" rather than "the quest's instructions" because this same message fires for a campaign, where one block is usually shared by several quests.

Screenshots in the PR description are re-captured from the running deck with the new wording.

---

Issue #2459, opened 2026-08-16
perf: work out each student's per-course XP once when archiving a semester


tylerecouture:
Follow-up from the review of #2456 (per-course XP, #2440). Not done there because the measured
cost does not justify the restructuring yet, but it grows with the deck, so worth tracking.

### What happens now

`CourseStudent.objects.calc_semester_grades()` asks each registration for its own XP, and
`CourseStudent.xp()` recomputes the same per-student aggregates
(`QuestSubmission.objects.xp_by_course(user)`, `BadgeAssertion.objects.xp_by_course(user)`,
`CourseStudent.objects.calculate_xp(user)`) once per registration. For a student in two courses
those identical queries run twice, and `save()` triggers an XP cache invalidation that asks again.

### What it costs

Measured on a seeded `hackerspace` dev deck (19 registrations, 11 students), archiving the open
semester inside a rolled-back transaction, counting only real queries (django-tenants
`SET search_path` statements excluded):

| | queries | query time |
| --- | --- | --- |
| per-course XP (as merged) | 447 | 0.12s |
| a flat even share (pre-#2440 behaviour) | 250 | 0.06s |

So the per-course work is roughly a 1.8x multiplier on a once-per-term operation that currently
takes a tenth of a second. It scales linearly with registrations, so a deck an order of magnitude
larger is still around a second.

### What to do

Compute each user's `xp_by_course` maps and adjustment total once, then hand them to each of that
user's registrations instead of every registration recomputing them. The single-course path in
`xp()` already short-circuits, so this only affects multicourse students.

Worth doing when a real deck's semester close feels slow, or alongside other archiving work.

- Claude Code (bytedeck - semester workflow redesign)

tylerecouture:
A second call site for this, found while building the per-course XP charts (#2453).

`courses:ajax_progress_chart` plots a point per class day, and each point asks the registration
what it held on that day, so `xp_for_registrations()` (and the two `xp_by_course()` queries
behind it) runs once per day of the semester rather than once for the whole series.

Measured on the seeded `hackerspace` deck, POSTing the chart endpoint for a student in two
courses, counting real queries (django-tenants `SET search_path` excluded):

| | queries | query time |
| --- | --- | --- |
| per-course chart (as built) | 132 | 0.08s |
| the deck-wide chart it replaces | 76 | 0.04s |

That is a 16-class-day semester, and it scales linearly with the days behind you, so a full term
[...]

Issue #2440, opened 2026-08-15
A student's XP is split evenly between their courses instead of attributed to the one it was earned in


tylerecouture:
Split out of #2157 (its Phase 3 "per-registration XP caching" item and open question 4), where it turned out not to be a concurrent-semesters problem after all. Filing it on its own so it doesn't get lost now that #2157's Phase 3 is finished.

### What

`Profile.xp_cached` is one number per student. A student taking two courses has it divided evenly between them:

```python
# src/profile_manager/models.py
def xp_per_course(self):
    course_count = self.num_courses()
    if not course_count or course_count == 0:
        return 0
    return self.xp_cached / course_count
```

So a student who earns 300 XP entirely in Programming and nothing in Art shows 150 in each. `Profile.mark()` divides the same way (`courses[0].calc_mark(self.xp_cached) / len(courses)`), and archiving stamps `CourseStudent.final_xp` from `xp_per_course()`, so the even split is what gets written into the permanent record at the end of term.

### Why it is not urgent, and why it is not a #1781 problem

#2157 listed per-registration XP caching as Phase 3 work on the reasoning that a single scalar becomes ambiguous once a deck can run several semesters. That reasoning does not survive the answer to that issue's open question 1: a student is registered in a single semester (they can take several courses within it), enforced in `CourseStudent.clean()` since #2437. So `xp_cached` belongs unambiguously to that one semester, and every read of it resolves through `courses.models.semester_for(user)`. Concurrent semesters are fine as they stand.

What is left is the per-*course* attribution, which has been an even split since long before #1781 and is wrong for any student taking more than one course. (Related history: #163, closed for the quest-to-course side of this.)

### What a fix looks like

Cache XP per `CourseStudent` rather than per `Profile`: a live sibling of the existing `final_xp` column. Every XP source already carries enough information to attribute it, since submissions and assertions are stamped with a semester and the student's registrations name the courses, but the mapping from a quest to a course is the real design question: quests are not currently owned by a course, so "which course did this XP come from?" has no answer yet without one. That is the part to settle first, and it may be the same question #403 and #546 are circling.

Worth deciding alongside:

* whether `final_xp` for already-archived semesters is left as the historical even split (recommended: it is a record of what was reported at the time) or backfilled;
* whether `Profile.xp_cached` stays as a denormalized total for ranks and the header XP display, which do want a single deck-wide number, or goes entirely.

- Claude Code (bytedeck - semester workflow redesign)

[...]

PR #2464, opened 2026-08-16
feat: chart each of a multicourse student's courses separately (#2453)

### LINKED ISSUE #2453: Show a multicourse student a rank and an XP chart per course
Follow-up to #2440, to build once its PR merges. Now that a student's XP is attributed per course rather than split evenly, the two places that report progress back to them still speak as though they have one number.

Both parts only apply to a student in **more than one course**. A student with a single course sees exactly what they see today, and neither part should introduce a dropdown or a tab strip where there is nothing to choose between.

---

### 1. A rank per course

Today the navbar shows one rank, derived from the student's whole-deck `xp_cached`:

```html
<!-- src/templates/navbar-fixed.html -->
<a title="Rank: {{request.user.profile.rank}}" href="{% url 'courses:ranks' %}">
    <i class="{{request.user.profile.rank.fa_icon}}"></i>
    <span class="visible-xs-inline-block">&nbsp;&nbsp;{{request.user.profile.rank}}</span>
</a>
```

A student in two courses has a different amount of XP in each, so they have a different rank in each. The navbar item becomes a dropdown:

* **Closed**: the icon of their **highest** rank, exactly where the single icon sits now.
* **Open**: one row per course, each showing that course's rank icon with the course name beside it.
* Rank names on hover (the native `title` attribute, per the front-end conventions in `CLAUDE.md`, not JS tooltips).
* Each row links where the current rank icon links (`courses:ranks`).

`Rank.objects.get_rank(xp)` already turns an XP number into a rank, and after #2440 `CourseStudent.xp()` gives the per-course XP, so the data is all there: it is presentation plus a small helper (something like `CourseStudent.rank()`).

Worth deciding while building: whether the mobile (`visible-xs`) collapsed view lists every rank or just the highest, and whether `Profile.rank()` keeps its current whole-deck meaning for the places that want one number (`Rank.condition_met_as_prerequisite` reads `xp_cached`, so it should).

### 2. An XP chart per course

The mark calculations page has a tab strip already:

```html
<!-- src/courses/templates/courses/mark_calculations.html -->
<li role="presentation" class="active"><a href="#progress" role="tab" data-toggle="tab">XP Progress</a></li>
<li role="presentation"><a href="#distribution" role="tab" data-toggle="tab">Mark Distribution</a></li>
<li role="presentation"><a href="#tag-distribution" role="tab" data-toggle="tab">Tags Distribution</a></li>
```

The XP Progress chart currently plots one line from the student's whole-deck XP over their semester's class days (`courses:ajax_progress_chart`). For a multicourse student it should be one chart per course.

Shown only when **both** `display_marks_calculation` and `students_choose_xp_course` are on: without the second, every course's chart would be the same line at a different scale, which is worse than one chart.

Buttons or a second row of tabs at the top of the XP Progress pane, one per course. `ajax_progress_chart` would take a course and plot that registration's XP to date, which needs a per-course version of `Profile.xp_to_date()`; the semester dates it charts against are already per registration after #2439.

---

Both parts hang off the same helper work, so they are filed together and would sensibly ship in one PR. Split them if the rank dropdown turns out to be the fiddlier half.

- Claude Code (bytedeck - semester workflow redesign)

tylerecouture:
### What?

The XP Progress chart on the mark calculations page becomes one chart per course for a student
in more than one, chosen with a button per course above it. Each course's line is its own XP over
the semester, drawn against that course's own XP-for-100%.

A student in one course sees no buttons and no change.

Closes #2453, whose other half (the per-course rank dropdown) merged as #2461.

### Why?

The chart plotted `profile.xp_to_date(day) / num_courses`: the student's deck-wide total, split
evenly. Since #2440 that is not what any of their courses holds. A student with 140 XP in Maths
and 12 in Art was shown one line at 76, a number neither course has, and the same line whichever
course they were thinking about.

### How?

`xp_by_course()` on both the quest and badge managers takes an `up_to_date`, and
`CourseStudent.xp_to_date()` runs the same assigned-versus-shared split as of a day rather than
as of now. The chart endpoint takes the course to plot in its POST and asks that registration for
each day's figure, so the series is the course's own.

The response also carries the charted course's `xp_for_100_percent`, since courses can be out of
different amounts and the axis and mark lines have to be redrawn when a student switches.

**The charts follow the courses, not the `students_choose_xp_course` setting.** That setting
decides whether students are *asked* where new XP goes; it does not un-assign what is already
assigned, and nothing that reads a student's XP consults it, including the marks table at the top
of this page. A deck that turns it off still has courses holding different amounts (that table
still shows 140% and 12%), so gating the charts on it would have left a multicourse student one
unlabelled chart that was silently their first course, contradicting the table above it.

**Bugs fixed along the way.**

- The old chart divided each day's point by the course count but not the weekend top-up added
  afterwards (the correction for XP earned on a non-class day), so a student who earned XP over a
  weekend saw the last point of their line jump to their undivided total. That is the spike in
  the "before" screenshot. Both figures come from the registration now.
- The charted course id arrives from the page's POST and reached the query unchecked, so a
  request naming something that is not a number (`course=abc`) raised
  `ValueError: Field 'id' expected a number but got 'abc'` and returned a 500. It falls back to
  one of the student's own courses now.

**What it costs.** The chart asks per class day, so I measured the endpoint on a seeded dev deck,
counting real queries (django-tenants `SET search_path` excluded):

| | queries | query time |
| --- | --- | --- |
| before (deck-wide line) | 76 | 0.04s |
| after (a course's own line) | 132 | 0.08s |

That is a 16-class-day semester and it scales with the days behind you, so a full term is nearer
800 queries for one chart load. It is on demand rather than per page, and only multicourse
students reach the split at all (a single course short-circuits), so rewriting the chart's day
loop, with its weekend and holiday handling, was not worth it here. Noted with the numbers on
#2459, which already tracks computing this once instead of per call.

### Testing?

`python src/manage.py test src` (2403 tests) and `ruff check src` both pass, and no file in the
codebase is below 100% statement and branch coverage.

New tests cover the endpoint charting the named course's own XP (one course's line ends at 40
where the other's ends at 0), scaling to the charted course's total, falling back to a course the
student is in when the request names none, names someone else's, sends the empty value the page
posts for a single-course student, or sends junk; the page offering a button per course for a
multicourse student with the setting on and with it off; and `CourseStudent.xp_to_date()` both
dividing as of a date and answering with the whole total for a registration that is not current.

Captured on a seeded `hackerspace` deck, as a student with 140 XP in Mathematics 10 and 12 XP in
Digital Art 11, both courses out of 1000.

**Before**: one line at the even share, with the weekend top-up spiking the final point up to the
undivided total.

before

**After**, with a button per course. Digital Art 11 holds 12 XP, so its line barely leaves the
axis:

after, art

Clicking Mathematics 10 redraws it against that course's 140 XP and its own trend:

after, maths

### Anything Else?

**The Mark Distribution and Tags Distribution tabs are unchanged.** Mark distribution is a
deck-wide comparison rather than a personal progress line, and tags are not attributed per course
at all, so neither has a per-course meaning to show yet.

**`Profile.num_courses()` is gone.** Its last caller was the chart, dividing a student's total by
how many courses they are in; the chart asks a registration for its own XP now, so nothing called
it.

### Review request
@tylerecouture

- Claude Code (bytedeck - semester workflow redesign)

### DIFF
--- a/src/badges/models.py
+++ b/src/badges/models.py
@@ -525,7 +525,7 @@ class BadgeAssertionManager(models.Manager):
         ]
         return by_type
 
-    def xp_by_course(self, user):
+    def xp_by_course(self, user, up_to_date=None):
         """How much of this student's badge XP counts toward each of their courses.
 
         A badge granted alongside a quest carries that quest's course; one granted on its own
@@ -534,13 +534,16 @@ class BadgeAssertionManager(models.Manager):
 
         Args:
             user: the student whose badge XP is being divided.
+            up_to_date (date): count only badges granted by then, for charting their progress
+                through the semester (issue #2453). Defaults to all of them.
 
         Returns:
             dict: course id to XP, with None collecting everything left unassigned.
         """
-        rows = self.get_queryset(True, user=user).grant_xp().get_user(user).values('course').annotate(
-            xp_sum=Sum('badge__xp'),
-        )
+        qs = self.get_queryset(True, user=user).grant_xp().get_user(user)
+        if up_to_date is not None:
+            qs = qs.get_issued_before(up_to_date)
+        rows = qs.values('course').annotate(xp_sum=Sum('badge__xp'))
         return {row['course']: row['xp_sum'] or 0 for row in rows}
 
     def calculate_xp(self, user):
--- a/src/courses/models.py
+++ b/src/courses/models.py
@@ -716,7 +716,7 @@ class CourseStudentManager(models.Manager):
         """
         return self.current_courses(user).aggregate(total=Sum('xp_adjustment'))['total'] or 0
 
-    def xp_for_registrations(self, user, profile=None):
+    def xp_for_registrations(self, user, profile=None, up_to_date=None):
         """Every current registration this student holds, with the XP counting toward it.
 
         The assigned-versus-shared split (issue #2440) is the same question for every course a
@@ -728,21 +728,24 @@ class CourseStudentManager(models.Manager):
             user: the student whose XP is being divided.
             profile (Profile): their profile, for a caller holding one whose xp_cached is
                 newer than the database. Defaults to reading it.
+            up_to_date (date): divide what they had earned by then rather than their total,
+                which is how each course's progress chart is plotted (issue #2453).
 
         Returns:
             list[tuple]: (CourseStudent, xp) in registration order, empty when the student is
             in no course. A student in one course has their whole total against it.
         """
         profile = profile if profile is not None else user.profile
+        total = profile.xp_cached if up_to_date is None else profile.xp_to_date(up_to_date)
         registrations = list(self.current_courses(user))
         if len(registrations) <= 1:
-            return [(registration, profile.xp_cached) for registration in registrations]
+            return [(registration, total) for registration in registrations]
 
-        # xp_cached is the student's deck-wide total: quest XP, badge XP, and every one of
+        # the total is the student's deck-wide figure: quest XP, badge XP, and every one of
         # their registrations' adjustments. Take out the parts that belong to a particular
         # course, share what is left, and hand each registration back its own pieces.
-        xp_by_course = QuestSubmission.objects.xp_by_course(user)
-        for course_id, xp in BadgeAssertion.objects.xp_by_course(user).items():
+        xp_by_course = QuestSubmission.objects.xp_by_course(user, up_to_date=up_to_date)
+        for course_id, xp in BadgeAssertion.objects.xp_by_course(user, up_to_date=up_to_date).items():
             xp_by_course[course_id] = xp_by_course.get(course_id, 0) + xp
 
         # Only courses the student still holds can claim XP: work assigned to a course whose
@@ -756,7 +759,7 @@ class CourseStudentManager(models.Manager):
         }
         assigned = sum(xp for course_id, xp in xp_by_course.items() if course_id in course_ids)
         adjustments = sum(registration.xp_adjustment for registration in registrations)
-        shared = (profile.xp_cached - assigned - adjustments) / len(registrations)
+        shared = (total - assigned - adjustments) / len(registrations)
 
         return [
             (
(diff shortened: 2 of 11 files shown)


tylerecouture:
@coderabbitai review

tylerecouture:
Thanks, this found a real problem. I fixed it in `c59a707f`, but the other way round from the suggestion, because keeping the shared calculation would have made the page contradict itself.

**Why not restore the shared line.** `students_choose_xp_course` decides whether students are *asked* where new XP goes. It does not un-assign XP already assigned: `CourseStudent.xp()` never consults the setting, it reads the `course` column. So on a deck that used the setting and then turned it off, every reader still attributes per course, including the marks table at the top of this very page. I checked on the seeded deck with the setting off:

```
setting off; marks table shows:
  Digital Art 11: 12 XP -> 12.0%
  Mathematics 10: 140 XP -> 140.0%
```

Charting the even share underneath that table would put two different stories on one page.

**What I did instead.** The charts follow the courses rather than the setting: a student holding more than one course gets a button per course, always. That removes the case you identified (a single unlabelled chart that is silently the first course) without introducing the contradiction, and it means the chart and the table agree in every configuration.

This does depart from what #2453 asked for, and the issue's reasoning was mine: I wrote that with the setting off "every course's chart would be the same line at a different scale". That is only true on a deck that has never attributed anything. Once any XP carries a course, the lines genuinely differ, and courses can be out of different amounts of XP anyway, so the scales differ regardless.

Covered by `test_mark_calculations__still_charts_per_course_with_the_setting_off`, which is the regression test you asked for: two registrations, XP assigned to one, setting off, buttons still rendered and the two courses still holding 50 and 0.

**One more thing while I was in there.** The charted course id comes straight from the page's POST, and anything non-numeric reached the query as-is: a request with `course=abc` raised `ValueError: Field 'id' expected a number but got 'abc'`, a 500. It falls back to one of the student's own courses now, and the test covers the empty value the page sends for a single-course student along with junk.

Full suite green (2403 tests) and no file below 100% coverage.

- Claude Code (bytedeck - semester workflow redesign)

---

Issue #2163, opened 2026-07-24
Library import preview renders the Submission Questions panel against the wrong schema


tylerecouture:
Found during a self-review of the submission-questions feature (#1304, PRs #2085/#2099).

### Problem
The Shared Library import-preview page includes `quest_detail_content.html`, which now renders a "Submission Questions" panel. `ImportQuestView.get` fetches the library quest inside `library_schema_context()` but renders the template **outside** it. The panel's `quest.question_set.all` is a lazy queryset evaluated at render time, so it queries the **current deck's** `questions_question` table with the library quest's pk — showing an unrelated local quest's questions (or none). "Manage Questions" likewise links to the local deck's `questions:list` for a pk-colliding quest, where staff could edit/delete an unrelated quest's questions believing they're the library quest's.

### Evidence
- `src/library/views.py` `ImportQuestView.get` renders outside `library_schema_context()`; `src/quest_manager/templates/quest_manager/quest_detail_content.html` panel evaluates `quest.question_set.all` at render.
- `confirm_export_quest.html` uses the local quest, so it's schema-correct — only the import preview is affected.

### Fix
Suppress the panel on library previews (pass a flag like the existing `dont_edit_prereqs` used in the same template), or evaluate/render the panel inside the correct schema context. Pairs naturally with #2162 (adding real library support for questions).

### Notes
Depends on #2099 (the panel is introduced there) — fix after #2099 merges.

- Claude Code (`Bytedeck - multiple submission types`)

tylerecouture:
Confirmed during a full audit of the Library import/export system, and it is more general than the submission-questions panel.

The same mechanism (collect inside `library_schema_context()`, render outside it, lazy queryset resolves against the local deck) also breaks tags on the Library campaign detail page and the campaign import preview:

```
[LZ2] 'chemistry' rendered on the page: False   # the tag is on the library quest
```

`LibraryQuestListView` escapes it only because it happens to `prefetch_related('tags')` (`views.py:106`), which forces evaluation in time. That accidental save is load-bearing: dropping the prefetch as a performance tweak would silently break tags there too.

Filed #2369 for the general fix (one helper that renders the response *inside* the schema context, used by every Library view), which closes this issue as a side effect rather than patching the panel alone. Suggest treating #2369 as the fix and this as the report, unless you'd rather ship the narrow `dont_edit_prereqs`-style flag first.

- Claude Code (`bytedeck - library import/export`)

---
[...]

PR #2467, opened 2026-08-16
fix: carry a quest's submission questions through the Shared Library (#2162)

### LINKED ISSUE #2162: Shared Library export/import does not carry a quest's submission questions
Found during a self-review of the submission-questions feature (#1304, PRs #2085/#2099).

### Problem
Quests exported to / imported from the Shared Library travel through `QuestResource`, which serializes Quest fields plus prereqs and campaign — but never the quest's `Question` rows. A quest whose instructions say "answer the questions below" arrives in another deck as an empty-form quest. No crash, just silent content loss. The same resource backs the admin CSV import/export, so those drop questions too.

### Evidence
- `src/library/exporter.py` / `src/library/importer.py` route quests through `QuestResource` (`src/quest_manager/admin.py`); its `Meta.exclude` and the only related handling are prereqs and campaign. `grep -i question src/library/` returns nothing.
- The conflict-clone path (`exporter.py`, `deepcopy` + `pk=None`) copies only Quest fields; deepcopy of an instance doesn't copy reverse FKs.

### Fix
Serialize a quest's questions in `QuestResource` (e.g. a dehydrated JSON column recreated in `after_import`, matched by the quest's `import_id` + question `ordinal`; note `Question` has no `import_id` today), and copy `solution_file` media across schemas. Until implemented, the export UI should state that questions are not included.

### Notes
Independent of #2099 (`Question` model is on `develop` via #2085). Larger than the other fixes — cross-schema serialization + media copy. Consider scoping to text-only questions first if file media copy is complex.

tylerecouture:
Closes #2162

### What?

A quest's submission questions now travel with it to the Shared Library and back.

### Why?

The transfer copied a quest's own fields, its tags, its campaign and its prerequisites, and nothing else. Questions are a separate table hanging off the quest, so they were simply not read: a quest whose instructions say "answer the questions below" arrived on the importing deck as an empty-form quest, and nothing reported the loss on either side. The sharer saw a successful push and the importer saw a successful pull.

That affects every path through the Library: the single-quest push and pull, the campaign push and pull, and the conflict-copy path that puts a second copy beside the Library's existing one.

### How?

`snapshot_quest` now includes the quest's questions, and `_write_quest_row` writes them on the far side, both built on the same helpers the rest of the module uses (`_copied_field_names` for what travels, `_read_field` so `solution_file` crosses as its stored path like every other file field).

Questions are matched by **ordinal**, which is the only identity they have across schemas: they carry no `import_id`, and the model already holds one question per ordinal per quest.

* A question at an ordinal the destination already uses is **updated in place**, so answers students already gave stay attached to the question they answered.
* An ordinal the arriving quest no longer uses is **removed**, so re-sharing a quest whose author deleted a question actually deletes it here. Answers to a deleted question survive it (`QuestionSubmission.question` is `SET_NULL`) and show in the marking view as answers to a question that is gone.

Replacing rather than merging is the bargain the rest of the quest is already written under: a re-import overwrites the quest's own name and instructions with the shared version, so a question the destination added to an imported quest goes the same way as an edit it made to that quest's text.

A question that cannot be written raises `LibraryTransferError` naming the quest and the ordinal, whether it fails validation or the database rejects it, so a view can put it in front of the sharer instead of a model-level error escaping as a 500. The write is wrapped in its own `transaction.atomic()`, matching how the quest's own row is written.

The campaign export and import querysets prefetch `question_set`, since snapshotting reads it once per quest.

### Testing?

`python src/manage.py test src` green, `ruff check src` and `manage.py check` clean.

New `LibraryTransferQuestionContractTests` in `library/tests/test_transfer_contract.py`, sitting beside the existing quest and campaign contracts. Seven of the eight fail on `develop` without this change:

* a round trip carries all three question types with every transferred field;
* the marker's `solution_file` arrives pointing at the same stored file;
* re-importing updates a changed question in place, keeping its row;
* re-importing removes a question the author deleted upstream;
* the conflict-copy path carries questions onto the fresh copy;
* a question that fails validation names the quest and the question;
* a question the database rejects does too (`Question.save` patched to raise, since that is by definition the case `full_clean` cannot see).

The eighth is a field inventory (`test_question_field_inventory__every_field_is_classified`), matching the ones Quest and Category already have: adding a field to `Question` now fails the suite until someone decides whether it belongs to the content or to the deck it was written on.

Real decks on the dev environment: `hackerspace` shares a quest with three questions (short answer, long answer, file upload) to the Library, and `importdeck` imports it. Both shots are `importdeck`'s Manage Questions page for the imported quest, as `teacher2`:

| Before | After |
| --- | --- |
| before | after |

Before, the imported quest says "No questions to display" even though the shared quest asks three. After, all three arrive with their type, instructions, required flag and marker notes.

### Anything Else?

The Library's own import **preview** still renders its Submission Questions panel against the importing deck's schema rather than the Library's ([#2163](https://github.com/bytedeck/bytedeck/issues/2163)), so it can show an unrelated local quest's questions. That is a separate bug in the view, and it is the next one I am picking up.

### Review request

@tylerecouture

- Claude Code (Bytedeck - multiple submission types)

### DIFF
--- a/src/library/exporter.py
+++ b/src/library/exporter.py
@@ -141,8 +141,13 @@ def export_campaign_to_library(*, source_schema, campaign_import_id, skip_import
     """
     with schema_context(source_schema):
         category = Category.objects.get(import_id=campaign_import_id)
-        # select_related: snapshot_quest reads quest.campaign for every row.
-        quests = Quest.objects.select_related('campaign').filter(published=True, campaign=category)
+        # select_related: snapshot_quest reads quest.campaign for every row, and its
+        # questions, which are a query each without the prefetch.
+        quests = (
+            Quest.objects.select_related('campaign')
+            .prefetch_related('question_set')
+            .filter(published=True, campaign=category)
+        )
 
         # filter out conflicts (they'll be cloned and exported later)
         if skip_import_ids:
--- a/src/library/importer.py
+++ b/src/library/importer.py
@@ -28,7 +28,13 @@ def import_campaign_to(*, destination_schema, quest_import_ids, campaign_import_
             instance because one of its names is already taken there.
     """
     with library_schema_context():
-        quests = Quest.objects.select_related('campaign').filter(published=True, import_id__in=quest_import_ids)
+        # select_related/prefetch_related: snapshot_quest reads each quest's campaign and
+        # its questions, which are a query each otherwise.
+        quests = (
+            Quest.objects.select_related('campaign')
+            .prefetch_related('question_set')
+            .filter(published=True, import_id__in=quest_import_ids)
+        )
         snapshots = [snapshot_quest(quest) for quest in quests]
 
     with schema_context(destination_schema):
--- a/src/library/tests/test_transfer_contract.py
+++ b/src/library/tests/test_transfer_contract.py
@@ -13,12 +13,13 @@ or a regression, and the failure is where the decision gets recorded.
 """
 
 import datetime
+from unittest.mock import patch
 
 from django.core.exceptions import ValidationError
 from django.test import SimpleTestCase
 from model_bakery import baker
 
-from django.db import connection
+from django.db import IntegrityError, connection
 from django_tenants.utils import schema_context
 from taggit.models import Tag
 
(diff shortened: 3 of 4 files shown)


tylerecouture:
@coderabbitai review

tylerecouture:
Good call on the error-contract parity: taken in 676fc1b.

`_write_questions` now catches `IntegrityError` alongside `ValidationError` and reports both the same way the quest's own write path does, and the write is wrapped in its own `transaction.atomic()` so a database rejection is contained exactly as `_write_quest_row` contains one.

Covered by a new test that patches `Question.save` to raise, since that narrow case is by definition the one `full_clean` cannot see: it asserts the sharer gets a `LibraryTransferError` naming the quest, the question's ordinal and the database's reason.

- Claude Code (Bytedeck - multiple submission types)

---

Issue #2369, opened 2026-08-13
Library: pages render lazy querysets after leaving the library schema, so data comes from the wrong deck


tylerecouture:
Found during a full audit of the Shared Library import/export system. Reproduced against a running stack. This is the general form of #2163.

### Problem

Library views collect their objects inside `library_schema_context()` and render **outside** it. Anything still lazy at render time then queries the *local* deck's tables using the Library object's pk. No error, just quietly wrong data on the page a teacher uses to decide whether to import something.

Confirmed instances:

- `CategoryDetailView.get_context_data` (`src/library/views.py:607`) does `list(category.current_quests())` with no prefetch, so `{{ q.tags.all }}` in `category_detail_content.html` resolves against the local deck.
- `ImportCampaignView.get` (`src/library/views.py:283`) does the same for the import preview.
- #2163 reports the same mechanism for the submission-questions panel in `quest_detail_content.html`.

### Evidence

A library quest tagged `chemistry`, rendered on the Library campaign detail page:

```
[LZ2] 'chemistry' rendered on the page: False
```

`LibraryQuestListView` escapes this only by accident: it happens to `prefetch_related('tags')` (`views.py:106`), which forces evaluation while the schema context is still open. Remove that prefetch as a performance tweak and the tags silently break there too.

```
[LZ] tags actually on the library quest: ['python', 'robotics']
[LZ] 'robotics' rendered on the page: True    # only because of the prefetch
```

### Fix

Rather than patching each instance, add one helper that renders the response **inside** the schema context, and use it in every Library view. That removes the whole class of bug and also lets the views stop hand-copying scalars (`category_id`, `category_name`, `category_quest_count`, ...) out into the context dict purely to dodge it, which is where #2370 comes from.

Closing #2163 falls out of this.

### Context

Audit context: PR #2362. Related: #1792 (tags missing from the Library) is this plus tags never being serialized in the first place.

- Claude Code (`bytedeck - library import/export`)

tylerecouture:
Another instance of this bug, outside the page-rendering paths this issue was filed for, so it does not get missed when the view cases are fixed.

`library/exporter.py:89-100`, in `export_campaign_to_library`, builds a queryset inside the source schema and first evaluates it after leaving that block:

```python
with schema_context(source_schema):
    ...
    quests = Quest.objects.filter(published=True, campaign=category)   # line ~89, lazy
    ...
# block has exited here
[...]

PR #2488, opened 2026-08-17
fix: show the campaign's name and blurb on the import preview (#2370)

### LINKED ISSUE #2370: Library: the campaign import preview shows a blank heading and "[No description provided]"

tylerecouture:
### What?

The campaign import confirmation page now says which campaign it is about. Its heading names the campaign, and its own short description is shown instead of "[No description provided]".

Closes #2370.

### Why?

`confirm_import_campaign.html` read an `object` variable that `ImportCampaignView` never supplies, and then did this:

```django
{% with object as category %}
```

`{% with A as B %}` assigns B = A, so that rebound `category` to the missing `object` for the entire block, shadowing the campaign the view had passed in. Every `category.*` lookup inside resolved to nothing. The heading (`{% firstof object.title category_campaign_name %}`) named two variables the view does not pass either, so it rendered as bare " Campaign".

The result is that the page a teacher makes the import decision on showed no campaign name and asserted the campaign had no description, whatever it actually said. It failed silently, which is why it lasted.

### How?

Three things, all in the template:

* Read the title from `category`, which the view already passes.
* Drop the `{% with %}` block, so the rest of the template sees the real campaign.
* Drop the `{% elif object.published or request.user.is_staff %}` branch and the "currently unavailable" message it guarded. That condition never tested what it appears to: `object` is absent, so it passed on `is_staff` alone, and the view is staff-only. Publication is enforced before the template is reached (`get_published_library_object` redirects an unpublished campaign back to the Library with an explanation), so the template only ever renders a published campaign and the dead branch was unreachable. The disabled-button check inside it went the same way: it tested `local_category` from inside a branch that only runs when `local_category` is falsy.

**The hand-copied scalars in the context dict stay.** #2370 suggests they exist only to dodge the shadowing, and for `category_name` that is true, but not for the rest: `quest_count()`, `xp_sum()` and `get_icon_url()` each hit the database, and this page renders after the Library schema context has closed, so reading them off the campaign object at render time would query the importing deck instead. That is #2369, which merged an hour ago. Only the already-loaded fields (`title`, `short_description`) are read from the object.

### Testing?

`LibraryImportCampaignPreviewTests`: the heading names the campaign, the description is shown, a campaign that genuinely has no description still says so, and the quest list and counts render. The first two failed before this change.

```
src/library                                   176 tests   OK
full suite (python src/manage.py test src)   2496 tests   OK
```

From the running `hackerspace` dev deck, signed in as the deck owner, previewing the Library's "Orientation" campaign, which has a real short description.

**Before:** the heading is just "Campaign", and the campaign is reported as having no description.

Before: bare Campaign heading and No description provided

**After:** the campaign is named, and its description is shown.

After: Orientation Campaign heading and the real description

### Anything Else?

**Worth knowing what else `{% with object as category %}` was hiding.** With the shadowing gone, `category_detail_content.html` reads the real campaign on this page for the first time. Its `{% firstof category_X category.X %}` pairs are what keep that safe: the view's pre-computed value still wins for every lookup that would otherwise query, which is why this PR does not simplify them away. If those scalars are ever removed, the schema-context hazard from #2369 comes back through this template rather than through the views.

### Review request

@tylerecouture

---

## Summary by CodeRabbit

* **New Features**
  * Improved the campaign import flow with clearer category titles.
  * Import previews now display campaign titles, descriptions, included quest details, and published quest counts.

* **Bug Fixes**
  * Removed unnecessary restrictions that prevented importing unpublished campaigns or campaigns without staff users.
  * Import is no longer disabled when a local category already exists.
  * Removed the unavailable-campaign message from the import flow.

### DIFF
--- a/src/library/templates/library/confirm_import_campaign.html
+++ b/src/library/templates/library/confirm_import_campaign.html
@@ -5,7 +5,7 @@
 {% block heading %}
   <i class="fa fa-compass pull-right"></i>
   {% block heading_inner %}
-    {% firstof object.title category_campaign_name %} Campaign
+    {{ category.title }} Campaign
   {% endblock %}
 {% endblock %}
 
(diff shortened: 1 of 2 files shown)


tylerecouture:
@coderabbitai review

tylerecouture:
Took the nitpick from the review: the `LibraryImportCampaignPreviewTests` docstring described the removed template behaviour, which is the "comments describe the code as it is now" rule. Fixed in `bdc25610` (note this resets the approval; it is a docstring-only change).

Two lines in the auto-generated summary above read as behaviour changes that this PR does not make, so for the record:

> Removed unnecessary restrictions that prevented importing unpublished campaigns or campaigns without staff users.

Unpublished campaigns are still not importable. That gate lives in the view, not the template: `get_published_library_object` returns `None` for an unpublished campaign and `ImportCampaignView.get` redirects back to the Library with "a Library admin still has to review and publish it". The template branch removed here (`{% elif object.published or request.user.is_staff %}`) never enforced it, because `object` is not a variable this view passes, so the condition was decided entirely by `is_staff` in a view that is already `staff_member_required`. Covered by the existing unpublished-campaign redirect test.

> Import is no longer disabled when a local category already exists.

The button is still not offered in that case. When `local_category` is set, the page takes the first branch, which shows the existing campaign and no form at all. The `disabled` attribute that went away was inside the *other* branch, which only runs when `local_category` is falsy, so it could never apply.

Both are unchanged behaviour with dead code removed, not loosened restrictions.

---

PR #2499, opened 2026-08-18
fix: make a push to the Library all-or-nothing (#2372)

### LINKED ISSUE #2372: Library: pushes and pulls are not atomic, so a failure can leave unreviewed content published

tylerecouture:
### What?

A push to the Library now lands whole or not at all. The content, the notification it raises and the origin rows that attribute it are written in one transaction, and the review email waits for that transaction to commit.

Closes #2372.

### Why?

A campaign push writes in several steps: the campaign's own quests, then the campaign row, then the conflict clones, and afterwards the notification and the origin rows. `write_quests` is atomic per batch, so each *step* held, but nothing held the steps together.

A failure between them left the Library holding part of a campaign: quests under a campaign nobody meant to push, some with prerequisites pointing at quests that never arrived, and no record of where any of it came from. The sharer is told the push failed, so nobody goes looking.

The review email made it worse, because it went out the moment the code reached it. A reviewer could be invited to come and look at content that a later failure rolled back. An email is the one part of this that cannot be undone.

**Half of the issue is already fixed, and this PR says so rather than claiming credit for it.** #2372 also describes a window in which unreviewed content was briefly *live* in the Library, between a committing import and a follow-up `save()` that unpublished it. That was the tablib path, and it is gone: content is written unpublished from the outset (`write_quests(..., published=False)`), so there is no window left to close.

### How?

* `export_campaign_and_copy_quests` wraps its whole sequence in `transaction.atomic()`. The body moved into `_push_campaign` so the transaction is one line at the top rather than an indent around forty.
* Both export views wrap the push, the `notify.send` and `record_push_origin` in one `transaction.atomic()`.
* `email_library_staff_of_push` dispatches through `transaction.on_commit`, so the Celery task is only queued once the content is really there.

Schema switching inside a transaction is fine here, and is what the transfer layer already relies on: `schema_context` changes the search path on the same connection, so one transaction spans both decks.

### Testing?

`LibraryPushAtomicityTests`, four tests around a campaign push made to fail at the origin-row write, which is the last thing in the transaction and so the narrowest case:

* the Library is left with nothing: no campaign row, neither quest
* no review email goes out
* the email **is** sent on the successful path, so the previous test cannot be satisfied by the email quietly never being sent
* an ordinary push still works, so the transaction is not drawn too tight

I checked the two failure-path tests actually fail against the unfixed code rather than assuming they were testing anything. Worth recording that the first version of the "no email" test was **vacuous**: the Library deck in that fixture had no staff member with an email address, so the helper returned early and nothing was ever dispatched, in either version. It now creates a Library reviewer with an address.

Two existing email tests needed updating: under `TestCase` the surrounding transaction never commits, so a deferred callback never runs. They now wrap their POST in `captureOnCommitCallbacks(execute=True)`.

```
src/library                                   193 tests   OK
full suite (python src/manage.py test src)   2528 tests   OK
```

None: this changes transaction boundaries and the timing of a background email. Nothing a user sees changes on the successful path, and the failing path already showed a server error. The behaviour that changed is what is *left behind* afterwards, which is what the tests assert.

### Anything Else?

**The import direction was already atomic** (`import_campaign_to` wraps its arrival, added with the transfer rewrite), so this PR is the push side catching up rather than a new pattern.

**`on_commit` has a cost worth naming.** Any future test that asserts on the review email has to run its POST inside `captureOnCommitCallbacks`, or it will look as though no email was sent. The two existing tests carry a comment saying so.

### Review request

@tylerecouture

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Campaign and quest exports now complete atomically, preventing partial content from being saved when an export fails.
  - Review emails are sent only after a successful export is committed.
  - Related notifications and attribution records now stay consistent with exported content.

### DIFF
--- a/src/library/exporter.py
+++ b/src/library/exporter.py
@@ -1,6 +1,7 @@
 from datetime import date
 from uuid import uuid4
 
+from django.db import transaction
 from django_tenants.utils import schema_context
 from quest_manager.models import Quest, Category
 from django.core.exceptions import ValidationError
@@ -184,6 +185,27 @@ def export_campaign_and_copy_quests(source_schema, campaign_import_id):
         Category.DoesNotExist: If the campaign is not found in the source schema.
     """
 
+    # One transaction for the whole push. `write_quests` is atomic per batch, but this
+    # runs several of them (the campaign's own quests, then the conflict clones) with
+    # a campaign write between, so without this a failure part-way leaves the Library
+    # holding some of a campaign and the sharer being told nothing arrived (#2372).
+    with transaction.atomic():
+        return _push_campaign(source_schema, campaign_import_id)
+
+
+def _push_campaign(source_schema, campaign_import_id):
+    """Do the work of `export_campaign_and_copy_quests`, inside its transaction.
+
+    Split out so the transaction is one line at the top rather than an indent around
+    everything, and so the steps below read in the order they happen.
+
+    Args:
+        source_schema (str): Tenant schema that contains the source campaign.
+        campaign_import_id (UUID): Import ID of the campaign to export.
+
+    Returns:
+        TransferResult: as `export_campaign_and_copy_quests` describes.
+    """
     # Step 1: get local campaign and quests first (in source schema)
     with schema_context(source_schema):
         local_campaign = Category.objects.get(import_id=campaign_import_id)
--- a/src/library/tests/test_views.py
+++ b/src/library/tests/test_views.py
@@ -488,7 +488,11 @@ class QuestLibraryTestsCase(LibraryTenantTestCaseMixin):
         self.client.force_login(self.test_teacher)
 
         url = reverse('library:export_quest', args=[self.local_quest.import_id])
-        response = self.client.post(url, data=AGREED_LICENCE)
+        # captureOnCommitCallbacks: the email waits for the push to commit, so that a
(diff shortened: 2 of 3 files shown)


tylerecouture:
@coderabbitai review

PR #2516, opened 2026-08-19
fix: give the Library review email its paragraphs back (#2371)

### LINKED ISSUE #2371: Library: the "content pushed" review email arrives as one run-on paragraph

tylerecouture:
### What?

The Library's "content pushed" review email is a real HTML template with real paragraphs, includes the shared footer, and is named `.html` like every other tenant email.

Closes #2371.

### Why?

`content_pushed.txt` was written as plain text with bare newlines, but it is never treated as plain text: `send_email_message` renders it into `admin/tenant/email/message.txt` (`{{ message|safe }}`), attaches that as the `text/html` alternative, and derives the plain-text part from it with `textify`.

HTML collapses bare newlines, and the template had no tags, so both versions of the email arrived as one block. The content's name ran straight into the sentence after it, which is exactly the part a reviewer needs to read, and this is the first impression a Library reviewer gets of a submission.

### How?

Renamed to `content_pushed.html`, paragraphs wrapped in real `<p>` tags, and the name given a paragraph of its own so nothing can run into it. The `get_template` call in `library/views.py` follows the rename.

It was also the only tenant email skipping `tenant/email/_email_footer.html`, so it now includes it, which replaces the old hand-written "ByteDeck Library" sign-off with the same closing every other platform email carries.

The em dash in that sign-off went with the rewrite, per the project's copy rules. It never rendered as one anyway: `textify` passed it through as a literal six-character escape sequence (backslash, `u`, `2014`), which is what a recipient actually saw.

The two other em dashes the issue points at (`library/views.py:457` and `:568`) were already gone, removed by #2501 when those two messages were converted to `format_html`.

### Testing?

Two new tests:

* `test_export_post__review_email_reads_as_paragraphs` pushes a quest for real and asserts the dispatched message has the name in its own block, that the next paragraph starts a block of its own, and that the break survives into the `textify`d plain-text part (asserting the run-on string is *not* present, which is the actual bug);
* `test_export_post__review_email_carries_the_shared_footer` asserts the footer is there and the old sign-off is gone.

Both fail against the old template. Full suite: 2594 tests, OK. `ruff check src` clean.

Both captures are from a real send on the dev deck, taken from the message the file-based email backend wrote to `_sent_mail/`, with the `text/html` part opened in the browser.

**Before.** One run-on block. "Intro to Python It has been added to the Library..." is the sentence a reviewer has to untangle, and the sign-off arrives as a raw escape sequence.

before

**After.** Paragraphs, the shared footer, and the name on its own line.

after

The plain-text part of the same two sends, which is what a text-only client shows. Two things are lightly edited so this description's renderer does not eat them: `textify` wraps bare URLs in angle brackets, and the old sign-off carried the escape sequence described above, shown here as `[u2014]`.

Before:

    owner shared a campaign from http://hackerspace.localhost:8000 to the
    ByteDeck Library: Intro to Python It has been added to the Library but is not
    visible to other decks until a Library admin reviews and publishes it. Review
    and publish it here: http://library.localhost:8000/quests/25/ [u2014] ByteDeck
    Library

After:

    owner shared a campaign to the ByteDeck Library from
    http://hackerspace.localhost:8000:

    **Intro to Python**

    It has been added to the Library, but is not visible to other decks until a
    Library admin reviews and publishes it.

    Review and publish it here:

    http://library.localhost:8000/quests/25/

    If you have any questions, contact us at
    [contact@bytedeck.com](mailto:contact@bytedeck.com).

    Bytedeck

    [\[Logo\]](https://d10ge8y4vx8iud.cloudfront.net/static/public/images/wordmark-v2.png)

    _Bytedeck is a non-profit Society registered in British Columbia. The board
    members are volunteers, and put many hours into running and further developing
    Bytedeck. Interested in joining our board and helping us steer the direction
    of this awesome app?_ _[Contact us](mailto:contact@bytedeck.com)!_

Note the run-on in the "before": the campaign's name and the sentence after it share a line, which is the whole of #2371.

### Anything Else?

The rename is the part worth a second look: `get_template` resolves by string, so nothing else can be referencing the old name, and I grepped to confirm the one call site is the only one.

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Added an HTML email notification when shared content is submitted to the ByteDeck Library for review.
  - Notifications include content details, source deck links, review access, and publication status.

- **Improvements**
  - Improved email formatting, including preserved paragraph breaks and the standard platform footer.


tylerecouture:
@coderabbitai review

PR #2546, opened 2026-08-20
fix: carry a prerequisite's NOT, count and OR half across the Library (#2535)

### LINKED ISSUE #2535: Library: NOT / repeat-count / OR prerequisites are silently simplified on the way across, and a NOT gate inverts
## What happens

The Library rebuilds prerequisites by calling `Prereq.add_simple_prereq`, which writes a plain single-target requirement (`src/library/transfer.py:356`, `src/prerequisites/models.py:563-594`). Three parts of a prerequisite are therefore dropped without anyone being told:

* `prereq_invert` (the **NOT**),
* `prereq_count` (the **x N times**),
* the alternate **OR** target (`or_prereq_object` and its own invert/count).

The count and the OR half degrade into a weaker gate, which is bad but visible if someone looks. The **invert is the dangerous one: it flips the meaning.** A quest gated on "NOT *Placement Test*" (available only to students who have **not** done the placement test) arrives on the importing deck gated on "*Placement Test*", so it is now available only to students who **have** done it. The gate is not merely lost, it means the opposite of what the author wrote.

Nothing warns about any of this: the "did not travel" message (`src/library/views.py:350-377`) only fires for prerequisites whose *target* could not travel, not for parts of a prerequisite that were discarded.

## Expected

At minimum, warn. The sharer already gets a message about gates whose target cannot travel; a simplified gate belongs in the same family, and the importer should hear about it too (right now the importer is told nothing at all, see the note below).

Better, carry the fields: `prereq_invert` and `prereq_count` are plain fields on the same row and could travel with it. The OR half is a second generic target, so it travels exactly when the main target does, under the same "is it in the destination" rule.

If inverted prerequisites are considered out of scope to carry, refusing to share content that has one is more honest than silently inverting it.

## Note on the importer's side

Both import views read only `imported.renamed_quests` and discard `unmet_prereqs` entirely (`src/library/views.py:766`, `:895`), so the importing teacher is currently told nothing about any prerequisite that did not survive. `src/library/tests/test_views.py:2277-2347` asserts that silence deliberately, so changing it is a decision, not a bug fix.

Found while verifying behaviour for the Shared Library wiki page. The wiki page now warns teachers to check the gating on anything they import before publishing it.

tylerecouture:
### What?

A prerequisite crossing the Library was rebuilt from its target alone, dropping three parts of the gate: `prereq_invert` (the NOT), `prereq_count` (the xN), and the alternate OR half.

Those now travel with it, and a lost alternative is reported separately from a lost gate.

Closes #2535

### Why?

The count and the OR half degrade into a weaker gate, which is bad but at least visible if someone looks. **The invert is worse: it flips the meaning.**

A quest gated on `NOT Placement Test`, available only to students who have *not* done the placement test, arrived gated on `Placement Test`, available only to students who *have*. The gate was not lost; it meant the opposite. Nothing on either deck said so, and the rule about who may see the quest silently inverted.

`_write_prereqs` rebuilt every gate through `Prereq.add_simple_prereq`, which by design writes a plain positive single-target requirement:

```python
new_prereq = cls(
    parent_content_type=..., parent_object_id=...,
    prereq_content_type=..., prereq_object_id=...,
)
```

There is no field for the invert, the count, or the alternative, so all three were dropped at the moment of writing.

### How?

**Snapshot both halves.** `_snapshot_prereqs` now returns `{'main': half, 'alternate': half | None}`, where each half carries `import_id`, `name`, `count` and `invert`. A new `_snapshot_prereq_half` applies the existing shareable-target rule (quests, campaigns and badges cross; ranks, grades, blocks and courses cannot) to each half independently.

**Rebuild the row rather than calling `add_simple_prereq`.** The alternate half travels under the same rule as the main one: rebuilt when the destination holds its target, reported when it does not.

**De-duplicate on the whole requirement.** `_prereq_signature` compares both halves in full, because `X` and `NOT X` gate a quest for opposite sets of students and `X x3` is not `X`, so the target alone cannot say whether a requirement is already present. This keeps the existing add-only behaviour (a repeat import does not stack a second copy) correct now that gates differ by more than their target.

**Report the two losses apart.** The existing warning says the copy "is not gated on it and anyone importing it will get it ungated". True for a gate whose target could not travel; false when only the alternative was lost, since the gate survives and the copy ends up *stricter* than its author wrote. `TransferResult` gains `unmet_alternates`, with a warning of its own:

> A gate kept its requirement but lost an alternative: 'The Other Route' is not in the Library, so where the gating offered it as another way through, the copy no longer does. Anyone importing this gets the stricter version. Share the alternative too if both routes should be available.

The fix a teacher needs differs between the two: an ungated quest needs re-gating on the far side, a narrowed one needs the alternative shared alongside it.

### Testing?

Six new tests. Five in `LibraryTransferPrereqContractTests`, all round-tripping through a campaign push and import (the path where gating survives at all):

* `test_push_and_pull__a_NOT_prereq_arrives_still_inverted`
* `test_push_and_pull__a_repeat_count_arrives_intact`
* `test_push_and_pull__an_OR_alternative_arrives_intact`
* `test_push_and_pull__an_OR_alternative_this_deck_lacks_is_dropped_and_named`
* `test_export_campaign__a_gate_that_cannot_travel_takes_its_alternative_with_it`

Plus `test_export_quest__warns_the_sharer_that_an_OR_alternative_could_not_travel` in `LibrarySharerWarningTests`, covering the new message end to end through the export view.

Verified in both directions: reverting `src/library/transfer.py` to `develop` fails four of them (`FAILED (failures=3, errors=1)`). The fifth and sixth exercise paths the old code did not reach; `test_import_twice__does_not_stack_a_second_copy_of_the_same_gate` passes either way by design, guarding the new signature-based de-duplication against regression rather than demonstrating the bug.

Full suite: `Ran 2651 tests ... OK`. `ruff check src`: clean. Coverage on `src/library/` is 100% (0 misses, 0 partial branches).

N/A: no template or styling changes. The only user-visible change is the new warning message, quoted verbatim above and asserted by `test_export_quest__warns_the_sharer_that_an_OR_alternative_could_not_travel`.

### Anything Else?

**One deliberate non-change.** When the *main* half cannot be rebuilt but the alternative could be, the whole requirement is still dropped rather than promoting the alternative to main. Promoting would silently rewrite which requirement is primary, and the existing contract for a gate that cannot be rebuilt is to report it rather than guess. `test_export_campaign__a_gate_that_cannot_travel_takes_its_alternative_with_it` pins that, and both names are reported as missing gating so the sharer sees the stronger of the two facts.

**The wiki already warns teachers about this.** The Shared Library page staged on `pr-assets` lists "Prerequisites are simplified on the way across" under Known Limitations. That line can come out once this merges.

### Review request
@tylerecouture

---

### DIFF
--- a/src/library/exporter.py
+++ b/src/library/exporter.py
@@ -259,6 +259,7 @@ def _push_campaign(source_schema, campaign_import_id):
     return TransferResult(
         quests=exported.quests + cloned.quests,
         unmet_prereqs=sorted(set(exported.unmet_prereqs) | set(cloned.unmet_prereqs)),
+        unmet_alternates=sorted(set(exported.unmet_alternates) | set(cloned.unmet_alternates)),
         skipped_quests=skipped_quests,
         dropped_common_data=sorted(set(exported.dropped_common_data) | set(cloned.dropped_common_data)),
     )
--- a/src/library/tests/test_transfer_contract.py
+++ b/src/library/tests/test_transfer_contract.py
@@ -15,6 +15,7 @@ or a regression, and the failure is where the decision gets recorded.
 import datetime
 from unittest.mock import patch
 
+from django.contrib.contenttypes.models import ContentType
 from django.core.exceptions import ValidationError
 from django.test import SimpleTestCase
 from model_bakery import baker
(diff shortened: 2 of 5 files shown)


tylerecouture:
@coderabbitai review

The review limit was reached when this opened, so no pass has run against `f8a4726b` yet. CI is green and the diff is unchanged since.

---

tylerecouture:
Heads-up for whoever reviews: PR #2545, opened a couple of minutes before this one from a different session, fixes the same issue #2535, so one of the two should close. The substantive differences, to make the pick quick:

* Both carry the NOT, the count and the OR half, rebuild the row directly instead of `add_simple_prereq`, and apply the same shareability rule to the alternate.
* Reporting a lost alternative: this PR gives it its own `unmet_alternates` warning saying the copy got stricter; #2545 names it in the existing "did not travel" list.
* Repeat transfers: this PR de-duplicates on the whole condition, so re-sharing a changed gate writes a second gate alongside the Library's old one (both must then be met); #2545 refreshes the matched gate in place on Library pushes (raised by review there) and stays add-only by target on deck imports, so re-imports cannot touch a teacher's adjusted copy.
* #2545 also carries before/after screenshots of the import preview from the running app.

Whichever is kept, the other session's improvements can be folded in: if #2545 is the keeper I am happy to adopt the separate stricter-copy warning from here; if this one is, the in-place refresh on re-share is the piece worth carrying over.

- Claude Code (`bytedeck - github issues workflow`)

---

tylerecouture:
Closing as a duplicate of #2545, which fixes the same issue and was opened two minutes before this one. Two sessions picked up #2535 at the same time and neither saw the other.

#2545 is the better of the two to keep: it also refreshes the Library's existing gate when a quest is re-shared, so a corrected condition propagates, and it carries real before/after screenshots of the import preview. Neither is in this PR.

The one thing here that #2545 does not have is the split reporting of a dropped OR alternative (`unmet_alternates` plus its own warning), because the existing warning tells the sharer the quest "will get it ungated" when in that case it arrives gated but narrower. I have raised that on #2545 and filed it as #2549 so it survives either way.

The branch stays available if any of this is worth lifting.

---

Issue #403, opened 2020-04-09
Option to have XP not associated with grades (for specific course)


mizzgoldberg:
Add config option to turn off grades completely, or turn them off for one block or course. XP just for joy. Suggest: start with block.

tylerecouture:
related to #544

tylerecouture:
Todo: add a course-specific option to use calcs or not?

PR #2584, opened 2026-08-23
fix: return the course scale in the empty progress-chart response (#403)

tylerecouture:
### What?
Makes `courses.views.ajax_progress_chart` return the charted course's scale (`uses_marks` and `xp_for_100_percent`) in its **empty** response, so the chart JSON has one consistent shape whether or not there is data to plot.

### Why?
`develop` currently fails `courses.tests.test_views.MarkCalculationsViewTests.test_ajax_progress_chart__reports_whether_the_charted_course_uses_marks` with `KeyError: 'uses_marks'`. Root cause is a semantic merge:
- The view's early-return path (no open semester, or a semester with no class days behind it yet) was added by `#2412` and returned only `days_in_semester` + `xp_data`.
- `uses_marks` (and the test asserting it) were added later by `#2524`/[#403](https://github.com/bytedeck/bytedeck/issues/403), but only on the **main** return path.

So the response shape depended on whether there was data to chart. `#2524`'s test passed when it merged (its fixture semester still had class days), but its semester is time-relative, so once no class days were behind it the request hit the early return and the key was gone. This also affected real students: someone at the very start of term (no class days yet) got a chart response missing the keys the front end reads.

This was found when merging `develop` into `staging` for the 1.32.0 release (staging CI went red on this one test); the fix has already been applied directly on `staging` to unblock the release, and this PR forward-ports it so `develop` is not left broken.

### How?
- Work out the charted course once, up front, so it is available to both paths.
- Include `uses_marks` and `xp_for_100_percent` in the early-return dict (falling back to `False` / `0` when there is no course to chart, e.g. no open semester).
- Removed the now-duplicate `_registrations_to_chart` call and the later `charted_course` recompute.
- Updated the two tests that pinned the old minimal empty shape (`TestAjax_ProgressChart.test_ajax_xp_data__empty_*`) to the new shape, referencing the fixture's own course values so they don't depend on model-bakery's field randomization.

### Testing?
`python src/manage.py test courses` passes (335 tests), including the previously-failing test and the two updated pin tests; `ruff check src` is clean.

N/A (the change is a JSON response shape; no visual change, and it prevents a broken chart for a start-of-term student).

### Anything Else?
Straight forward-port of the staging fix (same commit). No behavior change for a student who has data to plot.

### Review request
@tylerecouture

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Progress charts now consistently display the selected course’s scoring scale and marks settings, even when no chart data or dated class days are available.
  - Empty charts without a selected course correctly show a zero scale with marks disabled.

### DIFF
--- a/src/courses/tests/test_views.py
+++ b/src/courses/tests/test_views.py
@@ -2280,7 +2280,14 @@ class TestAjax_ProgressChart(ByteDeckTenantTestCase):
             reverse('courses:ajax_progress_chart', args=[self.student.pk]), HTTP_X_REQUESTED_WITH='XMLHttpRequest')
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(json.loads(response.content), {'days_in_semester': 0, 'xp_data': []})
+        # no points to plot, but the chart is still told the charted course's scale so its axis
+        # and mark lines set up correctly (issue #403): the response shape is the same either way
+        self.assertEqual(json.loads(response.content), {
+            'days_in_semester': 0,
+            'xp_data': [],
+            'xp_for_100_percent': self.course.xp_for_100_percent,
+            'uses_marks': self.course.uses_marks,
+        })
 
     def test_ajax_xp_data__empty_with_no_open_semester(self):
         """Between semesters there is no semester to chart progress through, so the chart gets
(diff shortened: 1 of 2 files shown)


tylerecouture:
Hotfixes merged and pushed locally.  PR not needed.

tylerecouture:
Hotfixes merged and pushed locally. PR not needed.

PR #2599, opened 2026-08-24
feat: search the approvals and submissions tabs on the server (#2597)

### LINKED ISSUE #2597: Quest approvals and submissions tabs: the search box only searches the current page

tylerecouture:
### What?

The quest approvals tabs (In Progress, Submitted, Approved, Flagged) and the quest submission tabs (In Progress, Completed, Past) now search on the server, over the whole tab rather than the page on screen.

Closes #2597

### Why?

These tabs are paginated and kept bootstrap-table's client-side search, which filters the row data the browser holds. On a queue of two hundred submissions that is the thirty rendered, so typing a student's name reported nothing found while their submission sat on page four.

That is worse than the sorting version of the same bug, which #2582 fixed: a sort confined to one page at least still shows you rows. A search that finds nothing reads as *this student has nothing waiting*, and there is no cue that the answer is on another page.

### How?

**Each tab searches what its own columns show.** The two flags on the helper exist for that reason: matching on a column a tab does not display would return rows whose match the reader cannot see anywhere, which reads as the list ignoring what they typed.

* The **approvals** tabs name the student, so they match the username and the preferred full name under it. That means four fields: `username`, `profile.preferred_name`, the account `first_name` behind it, and `last_name`. A teacher types whichever of them they know, and a surname finds the student whether or not they set a preferred name.
* The **submissions** tabs name the campaign and the tags instead, so those are matched there alongside the quest name.

Both keep the multi-word rule the quest lists use: every word has to match something, so adding a word narrows.

**`search_submissions()` sits beside `search_quests()`** in `quest_manager/listing.py`, so the two searches are read together.

**`sorted_submission_page` becomes `submission_tab_page`**, which narrows, orders and cuts the page in one place and hands back a `SubmissionTab`. The seven call sites each pass the flags for the columns their tab shows, which is one line per tab instead of three.

### Testing?

Ten tests in `SubmissionTabSearchTests`. The fixture puts one distinctive student at the far end of a 35-submission queue, so a search that only reached the rendered rows could not find them: searching reaches them; every name the User column shows matches, including the account first name behind a preferred one; several words narrow; the campaign and tag columns are searched only on the tab that displays them; the student's own Completed tab searches its campaign and tags; an empty result explains itself and keeps the box filled; the table no longer asks bootstrap-table to search; and a search and a sort apply together.

**Verified in both directions.** Reverting the four changed source files to the previous head fails all ten.

Full suite: `Ran 2752 tests ... OK`. `ruff check src`, `manage.py check` and `makemigrations --check --dry-run` all clean. `src/quest_manager/listing.py` stays at 100% coverage (19 statements, 12 branches, 0 misses, 0 partial branches).

Captured from a running dev deck (`hackerspace.localhost:8000`, signed in as the deck owner) with 40 submissions awaiting approval, so the queue spans two pages.

**Before**: no search box; bootstrap-table's own field searched the rendered page.

before: the approvals queue with no server-side search

**After**: the same `?q=` box the quest tabs and the Library use.

after: the approvals queue with a search box

**After**, searching `shari`. That student is on page two of the default queue, which is exactly the submission the old search reported as not found.

after: searching the approvals queue for a student on page two

### Anything Else?

Two things turned up on the way, both small and both fixed here:

**Six tests patched a manager method to return a plain list.** No manager method returns a list, and a seventh test in the same class already patched with a queryset. The list form worked only because nothing had ever called `.count()` or `.filter()` on the result; it broke the moment the tab counted its matches. They now use the queryset form the class already established. Worth noting that the sorting added in #2582 would have hit the same thing had those tests passed a `?sort=`.

**The match count read "1 submission match".** The verb agrees with the number now, here and on the quest tabs, with a test pinning it.

`multiKeywordSearch` stays in `bootstrap-table-accordion.js`: several unpaginated tables still use it.

This finishes the set: #2410 (PR #2593), #2582 (PR #2594), the quest tabs (PR #2598) and now the submission tabs. Every paginated quest list searches and orders on the server.

### Review request
@tylerecouture

---

## Summary by CodeRabbit

* **New Features**
  * Added server-side search for submissions across approvals and submissions tabs.
  * Search supports quest names, campaigns, tags, and submitter details where applicable.
  * Added clear-search controls, matching-result counts, and no-results messages.
  * Search applies across all matching submissions, not only the current page.

* **Bug Fixes**
  * Improved pluralization for quest search results.

### DIFF
--- a/src/quest_manager/listing.py
+++ b/src/quest_manager/listing.py
@@ -1,9 +1,12 @@
-"""Searching and ordering shared by the quest lists that are paginated a page at a time.
+"""Searching and ordering for the quest lists that are paginated a page at a time.
 
 The deck's quest tabs and the Library's quests tab list the same model through the same
-columns, so they search and order it the same way. Both are paginated, which is why this
-happens in the database: a page is all the browser holds, so filtering or ordering there
-answers a question about the page rather than about the list (#2379, #2410, #2582).
+columns, so they search and order it the same way. The approvals and submissions tabs list
+submissions of those quests, and search them by what their own columns show.
+
+All of it happens in the database because all of these lists are paginated: a page is what
+the browser holds, so filtering or ordering there answers a question about the page rather
+than about the list (#2379, #2410, #2582, #2597).
 """
 from django.db.models import Q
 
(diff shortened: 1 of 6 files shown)


tylerecouture:
@coderabbitai review

Issue #2569, opened 2026-08-20
Em dashes remain in submission-questions user-facing copy and code comments


tylerecouture:
Found in a full audit of the submission questions feature. The project bans em dashes anywhere (user-facing copy, docs, and code comments; maintainer decision 2026-07-30).

### User-facing
- `src/quest_manager/templates/quest_manager/submission.html:196`: the Copy Quest tooltip a TA reads (`title="Copy Quest &mdash; available to you because you're a TA: ..."`).
- `src/quest_manager/templates/quest_manager/preview_content_submissions.html:49`: the identical tooltip.

### Code comments/docstrings
- `src/questions/templatetags/question_tags.py:83-84`
- `src/questions/forms.py:81`
- `src/questions/utils.py:93`
- `src/questions/tests/test_integration.py:830` and `:948`
- `src/quest_manager/views.py:435` (`QuestCopy.form_valid` docstring) and `:1840` (questions-changed guard comment)

### Suggested fix
Replace each em dash with a colon or parentheses. The user-facing tooltips are the priority (the feature's own list page is already tested against em dashes at `src/questions/tests/test_views.py:108-109`, but these tooltips live outside those pages).

- Claude Code (`Bytedeck - multiple submission types`)

tylerecouture:
Picking this up. Three corrections after checking every site against current `develop`.

**1. The tooltips use `&mdash;`, not the character.** That is why they are easy to miss: a search for `—` returns nothing for them. CLAUDE.md bans both spellings, so they count.

**2. There is a third copy of that tooltip.** The issue lists `submission.html` and `preview_content_submissions.html`; the same string is also in `quest_manager/templates/quest_manager/buttons.html:108`. Fixing two of three would leave a TA reading the em dash on whichever page shows the third.

**3. Two listed sites are already clean.** `questions/utils.py:93` and one of the two in `questions/tests/test_integration.py` were fixed in passing while I was in those files earlier tonight (#2567 and #2560). The remaining one in that file is real.

Not touching `questions/tests/test_views.py:108-109`: those two lines are `assertNotContains(response, "—")` and `assertNotContains(response, "&mdash;")`, so the character is the subject of the assertion. Same for `tenant/tests/test_billing.py:797`.

[...]

PR #2621, opened 2026-08-26
feat: warn before leaving a submission or question form with unsaved text (#2572)

### LINKED ISSUE #2572: A submission form should warn before navigating away with unsaved answer/comment text (opt into the existing guard)
Found in a full audit of the submission questions feature.

### The problem
Drafts on the submission page persist only via the 60-second interval or the manual button, with no `beforeunload` protection. The repo already has an opt-in guard, `src/static/js/warn-unsaved-changes.js` (loaded site-wide from `templates/base.html`, triggered by `form[data-warn-unsaved]`, already summernote-aware), used by `quest_form.html` and `announcements/form.html`. But the student submission form tag (`submission.html:101`) lacks the attribute, as does the teacher's question editor (`questions/templates/questions/question_form.html:9`).

### Impact
A student typing an answer (or comment) who clicks a navbar link, hits back, or closes the tab within the 60-second autosave window leaves with no warning and silently loses up to a minute of typing. The same repo already gives teachers a native "Leave site?" prompt when editing a quest or announcement, so students editing submissions get strictly worse protection. The teacher question-editor form has no guard at all.

### Suggested fix
Add `data-warn-unsaved` to the student submission form (and `questions/question_form.html`); clear the dirty flag on successful `ajax_save_draft` responses if the always-armed prompt is too noisy alongside autosave.

tylerecouture:
### What?

The student submission form, the marker's approve form and the teacher's question editor now warn before you leave them with text that has not reached the server.

Closes #2572

### Why?

The guard has existed since #192. `warn-unsaved-changes.js` is loaded site-wide from `base.html` and opts a form in on `data-warn-unsaved`; `quest_form.html` and `announcements/form.html` use it. The three forms where the most typing happens never did.

A student's answers and comment reach the server on a submit or the 60-second autosave, and nothing else: no keystroke handler, no `change` handler, no `beforeunload`. Click the navbar in between and up to a minute of typing is gone, silently. A teacher half way through marking feedback, or through a question's instructions, solution and marker notes, loses all of it the same way, with nothing saved anywhere until they submit.

### How?

`data-warn-unsaved` on all three forms. On the submission page the comment box is a `<div id="main-comment-form">`, not a nested form, so the one attribute covers the comment editor and every answer editor at once.

**Arming it unconditionally would have been worse than not doing this.** The common exit is: student types, the autosave saves everything, student clicks away. A guard that cannot see the autosave says "Changes you made may not be saved" there every single time, and a warning that is usually wrong is exactly the one people learn to click straight through.

So the guard counts edits instead of setting a flag, and offers a two-call API on the form element for saves that are not submits:

```js
var sent = form.warnUnsaved.edits();   // before the request leaves
form.warnUnsaved.saved(sent);          // once the server confirms it
```

`save_draft` reports every successful save that way. **Passing the count captured before the request is the whole point:** an edit typed while the request was in flight leaves the guard armed, because the server did not store it. Clearing unconditionally on success would trade a noisy false positive for a silent false negative, which is the failure this issue is about.

That is not a new idea in this file. `save_draft` already snapshots `chosen_files_by_input()` before the POST and compares with `same_files(current, sent)` afterward, for the same reason and against the same race. The dirty count now gets the same treatment.

### Testing?

Full suite `Ran 2857 tests in 491.871s ... OK`, plus `ruff check src`, `manage.py check` and `makemigrations --check --dry-run` (no changes detected).

Two new tests, both failing on unpatched templates, covering all three forms.

**They nearly did not.** `assertContains(response, "data-warn-unsaved")` passes on *every page in the site*, because `base.html` names the attribute in the HTML comment above the script tag. Both tests initially passed against `develop`, which is how I found it. They now match `<form[^>]*\sdata-warn-unsaved` instead, and the constant carries a comment saying why:

```
FAIL: test_create__form_warns_about_unsaved_changes
FAIL: test_submission_view__both_forms_warn_about_unsaved_changes
Ran 2 tests ... FAILED (failures=2)
```

The guard's behaviour is JavaScript, so it is driven in a real browser on a seeded `hackerspace` deck instead. Each line dispatches a cancelable `beforeunload` and reports whether the handler cancelled it, which is precisely the condition Chrome uses to decide whether to prompt:

```
form has data-warn-unsaved: true
guard API attached: true
1. untouched page armed?           false (expect false)
2. after typing a short answer     true (expect true)
3. after a successful Save Draft   false (expect false)
4. after editing again             true (expect true)
5. edit made while save in flight  true (expect true)
6. saved again, clean              false (expect false)
7. after typing in the editor      true (expect true)
```

Line 5 is the race, with a real 200 from `/quests/submission/save/`: the save succeeds and the guard stays armed, because the keystroke landed after the request left.

Then the real prompt, by actually closing the tab:

```
native dialogs seen on close with unsaved text: ["beforeunload: \"\""]
native dialogs seen on close after Save Draft:  []
```

And the marker's side, including that submitting is still silent:

```
staff form guarded: /quests/submission/60/approve/
1. untouched                  false (expect false)
2. after typing feedback      true (expect true)
3. submitted, dialogs seen:   []      (expect [])
   landed on: /quests/approvals/
```

**This change renders no new pixels, and the thing it produces cannot be screenshotted.** The visible outcome is Chrome's own "Leave site? Changes you made may not be saved." prompt, which is browser UI drawn outside the page, so no page-level screenshot tool reaches it; and any automation able to trigger it (CDP, and therefore Playwright and devtools) also intercepts it before the browser draws it, which is why the transcripts above report the dialog rather than picture it.

It is not a new dialog in any case: it is character for character the prompt this repo has shown on the quest form and the announcement form since #192. Editing a quest, typing into it and clicking the navbar shows exactly what a student will now see.

### Anything Else?

**Drop Quest draws the prompt too**, because it is a plain link. A student who has typed something and then decides to drop the quest gets "Leave site?" and then the drop page's own confirmation: two dialogs for one intent, and the first is arguably a false alarm since dropping discards the draft anyway. I have left it rather than widening this PR, since the same is already true of the Cancel button on the quest form (a link inside a guarded form, since #192). Say the word and I will exempt that link.

The two em dashes in `warn-unsaved-changes.js`'s comments are replaced, per the convention about cleaning them from files you touch. It is not in the directories #2569's checker scans yet.

### Review request
@tylerecouture

---

## Summary by CodeRabbit

- **New Features**
  - Added unsaved-change warnings to submission and question forms.
  - Warnings now account for edits made while drafts are being saved, preventing newer changes from being marked as saved incorrectly.
  - Navigation warnings are suppressed while a form is being submitted.

- **Bug Fixes**
  - Improved draft-saving behavior so unsaved edits are reliably preserved and tracked.

### DIFF
--- a/src/quest_manager/templates/quest_manager/submission.html
+++ b/src/quest_manager/templates/quest_manager/submission.html
@@ -95,10 +95,14 @@
       <h3 class="panel-title">Quest Submission Form</h3>
     </div>
     <div class="panel-body panel-summernote">
+      {% comment %} data-warn-unsaved opts both forms into warn-unsaved-changes.js, so leaving the
+         page with answer, comment or marking text that has not reached the server draws the
+         browser's "Leave site?" prompt (#2572). The comment box is a div, not a nested form, so
+         the attribute here covers it and every answer editor. {% endcomment %}
       {% if request.user.is_staff %}
-        <form method="POST" enctype="multipart/form-data" action="{% url 'quests:approve' submission.id%}">
+        <form method="POST" enctype="multipart/form-data" action="{% url 'quests:approve' submission.id%}" data-warn-unsaved>
       {% else %}
-        <form id="submission-main-form" method="POST" enctype="multipart/form-data" action="{% url 'quests:complete' submission.id%}">
+        <form id="submission-main-form" method="POST" enctype="multipart/form-data" action="{% url 'quests:complete' submission.id%}" data-warn-unsaved>
       {% endif %}
 
         {% csrf_token %}
@@ -333,6 +337,7 @@
                 // the file that went out, so it is only allowed to touch an input still
                 // holding it (#2563).
                 var sentFiles = chosen_files_by_input();
+                var sentEdits = edits_so_far();
                 formData.set('csrfmiddlewaretoken', '{{ csrf_token }}');
                 formData.set('comment', draftComment);
                 formData.set('submission_id', '{{ submission.id }}');
@@ -350,6 +355,7 @@
                     success : function(json) {
                       $('#draft-error').hide();
                       show_saved_files(json, sentFiles);
+                      mark_draft_saved(sentEdits);
                       hide_check_show_text();
                     },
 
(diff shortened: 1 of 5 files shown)


tylerecouture:
@coderabbitai review

Issue #2683, opened 2026-09-11
Student list: the group filter vanishes instead of showing an empty control when no group is in an open semester


tylerecouture:
The group filter added in #2081 / PR #2199 is rendered behind `{% if block_filter_choices %}` in `src/profile_manager/templates/profile_manager/profile_list.html`, and `ProfileList.get_block_filter_choices()` only offers groups that have a `CourseStudent` registration in an **open** semester.

So on a deck where no group has a registration in an open semester (a new deck, or between semesters), the control is simply not on the page. A teacher who used it last term and looks for it sees nothing, with no indication of why.

Two things hide it, and only the first is in question here:

1. no group has a registration in an open semester, so `block_filter_choices` is empty and the whole `{% if %}` block is skipped
2. it is offered only on "All Active" and "Current Semester", not on the per-group, Staff or Inactive lists (deliberate: the first is already one group, and the other two render no group column)

Options for (1):

- always render it for staff, with only the "All Groups" option, so the control keeps its place
- render it disabled with a short "no groups this semester" note
- leave as is

Raised on the student list toolbar PR (branch `claude/profile-list-toolbar-layout`), which does not change this behaviour.

- Claude Code (`bytedeck - single session`)

Issue #2115, opened 2026-07-23
Submission status page


tylerecouture:
1. "Status Breakdown" should be an H3 and "Status Summary Stats"
2. After the stats table, add an HR to seperate the content below it.
3. Then add another H3: "Status by Student"
4. Change "enrolled" to "current" for consistancy
5. Change "All active" to just "All" to remove confusion between "Active" and "Current" (maybe check the rest of the app for this too)
6. Move the filter to the right, with the search<img width="915" height="564" alt="Image" src="https://github.com/user-attachments/assets/16703c36-d30c-4599-828a-a4b383921309" />

PR #2686, opened 2026-09-11
style: put the student list filters in the table's toolbar row (#2081)

tylerecouture:
### What?

Puts the student list's group filter and search box into the table's own toolbar row, at the right of it, alongside the bootstrap-table column show/hide button.

- the table's `data-toolbar` points at the filter form, so bootstrap-table moves it into `.fixed-table-toolbar`
- the form takes the shared `.bt-toolbar-filter` class (#2115), which right-aligns that row
- that class no longer leaves the row 15px short of the table's right edge

### Why?

The filters rendered on a row of their own above the table, which left the column button stranded on a second row below and to the right, so the header area took two rows to say very little.

They now sit where the rest of the app puts its search, per review feedback on this PR.

### How?

bootstrap-table moves whatever element `data-toolbar` names into `.fixed-table-toolbar`, and `.bt-toolbar-filter` makes that row a right-aligned flex row. The search ends up left of the column button, which is bootstrap-table's own arrangement for a table carrying both: rendering a bare table with `data-search` + `data-show-columns` against the vendored assets puts columns flush right with the search beside it.

**The 15px the shared class was losing.** bootstrap-table gives `.fixed-table-toolbar` a clearfix `::after`, and a flex container counts a pseudo-element as a flex item, so the row's `gap: 15px` applied *after* the last control and the row stopped short of the table's right edge. Flex needs no clearfix, so it is dropped. The override is scoped under `.bootstrap-table` to match the specificity of the rule it replaces (`.bootstrap-table .fixed-table-toolbar::after`), since that stylesheet is loaded per page and lands after `custom_common.css`. `custom_common.css` cache-buster bumped to `v=2.15`.

**This moves a second page, and that is intended.** `quest_user_status` is the other user of `.bt-toolbar-filter`, so its search shifts 15px right too, ending flush like every other table's. Raised with the maintainer as a change nobody asked for, and confirmed as wanted rather than scoped to the student list alone.

If the script never runs, the form stays where it is server-rendered and still works: it is a plain GET form. The group filter's auto-submit script keeps working too, since `filter.form` still resolves after the form is moved.

### Testing?

`ProfileViewTests.test_profile_list__filters_sit_in_the_table_toolbar` asserts the table's `data-toolbar` names the filter form's id, and that the form carries `.bt-toolbar-filter` **as a class token** rather than as a substring of the tag, so dropping either moves the layout back loudly rather than quietly. It fails on `develop` ("the student table declares no data-toolbar").

Full suite green, `ruff check src` clean.

Measured on the `hackerspace` deck at 1150px wide, against the mark every other table sets:

| page | search right edge | toolbar right edge |
|---|---|---|
| tags (search only, the convention) | 1135 | 1135 |
| student list, now | search 1063, columns flush at **1135** | 1135 |
| quest user status, now | **1135** | 1135 |

Real `hackerspace` dev deck, `/profiles/list/all/`, logged in as the deck owner.

**Before**: filters on their own row, column button alone on the next row.

before

**After**: one row, filters and search pulled right, column button flush at the table's edge.

after

At phone width the form stacks and the column button drops below it, which is Bootstrap's own `.form-inline` behaviour under 768px. Checked that nothing here widens the page: the horizontal scroll at 400px comes from the table itself and the app's `.row-offcanvas`, both unchanged.

### Anything Else?

Follows on from #2199 / #2081, which added the group filter.

The group filter disappears entirely (rather than showing an empty or disabled control) when no group has a registration in an open semester, because the template is wrapped in `{% if block_filter_choices %}`. Not changed here; filed as #2683 with the options.

While looking for a page with both controls to compare against: **bootstrap-table never initialises on `/portfolios/`** at all, even though its markup asks for search, sorting and the column selector. Unrelated to this PR and pre-existing; filed as #2691.

### Review request
@tylerecouture

## Summary by CodeRabbit

* **Improvements**
  * Profile list filters now appear together in the table toolbar with consistent right alignment.
  * Improved toolbar spacing so filter controls align cleanly with the table search box.

* **Bug Fixes**
  * Ensured the profile list consistently connects its filter form to the table toolbar.

* **Tests**
  * Added regression coverage for filter placement, identification, and toolbar alignment.


tylerecouture (commented):
Regarding the screenshot, pull those items right, so the search is on the right matching all our other search field locations.

tylerecouture:
Done in `e511ebbf`. (Your comment was a review with no inline thread, so replying here.)

The form now takes the shared **`.bt-toolbar-filter`** class (#2115) rather than a rule of its own, which right-aligns the toolbar row. That also puts the search left of the column show/hide button, which I checked is bootstrap-table's own arrangement for a table carrying both: rendered a bare table with `data-search` + `data-show-columns` against the vendored assets, and the library puts columns flush right with the search beside it.

**It also turned up why that class was never quite flush.** It left the row 15px short of the table's right edge: bootstrap-table gives `.fixed-table-toolbar` a clearfix `::after`, and a flex container counts a pseudo-element as a flex item, so the row's `gap: 15px` was applying *after* the last control. Flex needs no clearfix, so it is dropped. It has to be scoped under `.bootstrap-table` to match the specificity of the rule it overrides, since that stylesheet is loaded per page and lands after `custom_common.css`.

Measured against the mark, on the `hackerspace` deck at 1150px wide:

| page | search right edge | toolbar right edge |
|---|---|---|
| tags (search only, the convention) | 1135 | 1135 |
| student list, before | n/a, search was on its own row above | |
| student list, now | search 1063, columns flush at **1135** | 1135 |
| quest user status, now | **1135** | 1135 |

**Heads up that this moves a second page**: quest user status is the other user of `.bt-toolbar-filter`, and its search shifts 15px right, ending flush like every other table's. That looks like a fix rather than a regression to me, but it is a change you did not ask for, so say the word and I will scope the `::after` override to the student list alone.

after

`src/profile_manager` green (165 tests) and `ruff check src` clean. `custom_common.css` cache-buster bumped to `v=2.15`. The test now also pins the class, so dropping it moves the search back left loudly rather than quietly.

Separately, while looking for a page with both controls to compare against: **bootstrap-table never initialises on `/portfolios/`** at all (the table stays raw, `.bootstrap-table` wrapper count is 0), even though its markup asks for search, sorting and the column selector. That is unrelated to this PR and pre-existing; I will file it.

---

tylerecouture:
@coderabbitai good catch, valid and fixed in `e995e74e`. (Replying here rather than on the thread: the review-threads API has been rate limited for a while.)

`self.assertIn('bt-toolbar-filter', form.group(0))` searched the whole `<form>` tag for a substring, so it was also satisfied by a longer class that merely starts the same way, or by the name appearing in some other attribute. It now reads the class attribute and checks for the exact whitespace-separated token:

```python
classes = re.search(r"""\sclass=["']([^"']*)["']""", form.group(0))
self.assertIsNotNone(classes, 'the filter form carries no class attribute')
self.assertIn('bt-toolbar-filter', classes.group(1).split())
```

Verified the new assertion is genuinely stricter rather than just differently worded, by renaming the class to `bt-toolbar-filter-wide`: the old one passed, the new one fails with `'bt-toolbar-filter' not found in ['form-inline', 'bt-toolbar-filter-wide']`.

`src/profile_manager` green (165 tests) and `ruff check src` clean.

---

tylerecouture:
Maintainer has accepted the 15px shift, so the `::after` override stays as it is and `quest_user_status` gets the flush alignment too. No code change needed; the description now records the decision instead of asking the question.

Nothing outstanding on this PR: green and up to date with `develop` (head `1f7d47b7`, 6/6), CodeRabbit's class-token finding fixed in `e995e74e`, and no open questions from me.

---

PR #2688, opened 2026-09-11
fix: stop a teacher's quick reply turning up in another quest's box (#2685)

### LINKED ISSUE #2685: Quick reply draft text bug
Responses (draft?) in quick reply are getting placed into OTHER quest quick replies when the page reloads and the list has changed, for example when a quest is submitted, that text now appears in a different quick reply for some reason.

tylerecouture:
### What?

A teacher's quick reply box on the approvals page now starts empty on every page load, and each row's box is its own control rather than a copy of one shared form.

Three changes, all in the approvals page's quick reply:

- the form is **unbound**, so nothing that was posted is rendered back into the boxes
- each row gets **its own form instance with its own DOM ids**, via `auto_id`
- the reply box carries **`autocomplete="off"`**, so the browser does not refill it

Closes #2685

### Why?

There is no draft saving here to remove: nothing stores a quick reply anywhere. What there is, is that every row rendered the *same* `SubmissionQuickReplyForm` instance, bound to `request.POST`. That one fact produced two defects.

**1. A posted reply was rendered into every box at once.** On a real deck with 30 submissions awaiting approval, one reply came back in all 30 boxes. That is the before screenshot below, and it is a real way a reply appears under a quest it was not written for.

**2. All 30 boxes had the same DOM ids.** `id="id_comment_text"` and `id="id_award"` each appeared 30 times on the page. That is invalid HTML and it breaks every label's `for`.

### How?

`approvals()` builds one unbound form per submission and hangs it on the submission, so the template renders `{{ s.quick_reply_form | crispy }}`. `auto_id` makes the ids `id_quick_reply_<submission id>_comment_text`. The field **names** are deliberately unchanged, since `ApproveView` reads the reply back from `comment_text` and `award`.

**The award list is fetched once for the page, not once per row.** `Badge.objects.all_manually_granted()` runs a prerequisite count per badge, so a naive form-per-row was measured at **750 queries** for a page of 30. `SubmissionQuickReplyForm.build_award_choices()` fetches the options once and the new `award_choices=` argument hands the same pairs to every row. Assigning `.queryset` would not do: `ModelChoiceField` copies what it is given and the copy carries no result cache, so each row would re-query. `award_choices` sets the select's options only, which is all a form being *rendered* needs; the form `ApproveView` builds to *clean* a submitted award leaves it out and gets the queryset that validates against.

Net effect on the real page is **fewer** queries than before, because the shared list is also not re-evaluated per render:

| | queries for `/quests/approvals/submitted/all/` |
|---|---|
| `develop` (at the time of measuring) | 166 |
| this PR | 139 |

### Testing?

Four new tests in `ApprovalsViewTest`. **Three fail on the code this replaces**:

- `test_approvals__each_quick_reply_box_carries_its_own_ids` (`'id="id_comment_text"' unexpectedly found`)
- `test_approvals__a_posted_reply_is_not_echoed_into_every_box` (`3 != 0 : 'ECHOED-REPLY' unexpectedly found`)
- `test_approvals__quick_reply_boxes_do_not_autocomplete`

The fourth, `test_approvals__award_list_is_fetched_once_for_the_whole_page`, passes either way: it is not a reproduction but a guard on this PR's own design, since form-per-row is the obvious implementation and the expensive one. Confirmed it bites, by dropping the shared list and re-running it: `queries grew from 88 to 263 over 7 extra rows, which suggests the award list is being rebuilt for each one`.

Full suite on the branch merged up to date with `develop`: `Ran 2976 tests ... OK`, plus `ruff check src`, `manage.py check` and `makemigrations --check --dry-run` all clean.

This branch carries a merge of `develop` after #2690 landed: both changed the same block in `tab_quests_approvals.html`, and the resolution keeps this PR's per-row form together with #2690's `next` field, which are complementary.

Real `hackerspace` dev deck, `/quests/approvals/submitted/all/` as the deck owner, 30 submissions awaiting approval. The same page rendered from a POST carrying one reply, first three rows opened.

**Before**: the one reply is in every box.

before

**After**: every box is empty.

after

Measured off the rendered page rather than eyeballed:

| | before | this PR |
|---|---|---|
| reply boxes on the page | 30 | 30 |
| boxes holding the posted reply | **30** | **0** |
| elements matching `#id_comment_text` | **30** | **0** |
| distinct box ids | **1** | **30** |
| boxes with `autocomplete="off"` | 0 | 30 |

### What the browser actually does, measured

An earlier version of this description claimed the duplicate ids were "what leaves a browser nothing to tell the boxes apart". **That was an inference, and testing it showed it is wrong.** Recording the correction here rather than quietly deleting it.

I built a page shaped like this queue (four forms, each with a `comment_text` textarea and its own `action`), typed a reply into the second box, had a submission leave the queue, and came back to the page in Chromium:

| variant | what came back |
|---|---|
| shared ids | restored into the **correct** box |
| distinct ids | restored into the **correct** box |
| distinct ids + `autocomplete="off"` | **nothing restored** |

So Chromium keys form restoration on the form and the field *name*, not the `id`, and the names are all still `comment_text` by design. Two conclusions:

- **`autocomplete="off"` is the only one of the three changes that affects this at all**, and it is not redundant with the ids.
- **The distinct ids are a correctness fix** (valid HTML, working labels), not the thing that stops a reply appearing in the wrong box.

I also could not get Chromium to restore into the *wrong* box under any variant, including when the row typed into was the one that left the queue. Restoration only happens on back-navigation, never on reload. So the browser half of #2685 is still unreproduced; what is reproduced and fixed here is defect 1 above.

**`autocomplete="off"` stays**, confirmed with the maintainer, on the narrower justification in its code comment: a reply is about one specific submission and is sent the moment a button is pressed, so a half-typed one reappearing when a teacher returns to the queue is never useful, and could go out unnoticed with the next Approve.

Caveat: Chromium only. There is no Firefox build in this environment, and its form-history behaviour differs.

### Anything Else?

**The student-side box is untouched.** `SubmissionQuickReplyFormStudent` on the submission page is a different form and renders once, so neither defect applies to it.

### Review request
@tylerecouture


tylerecouture:
Both docstring nitpicks done in `7efa3e6` (they were in the review body with no inline threads, so replying here).

- `SubmissionQuickReplyForm.__init__` now documents the forwarded `*args` / `**kwds` and its `Returns: None`, matching how `QuestForm.__init__` in the same file documents its pass-through arguments.
- `approvals()` now has `Args:` for `request` (including which querystring params it reads: `?page`, `?sort`, `?order`, `?q`), `quest_id` and `template`, a `Returns:`, and a line on each submission carrying its own unbound `SubmissionQuickReplyForm`.

`ApprovalsViewTest` green (15 tests) and `ruff check src` clean on the new commit. No behaviour change in this commit.

---

tylerecouture:
Maintainer confirmed: **`autocomplete="off"` stays.** That is what this branch already does, so no code change; `test_approvals__quick_reply_boxes_do_not_autocomplete` keeps it honest. The description now records the decision instead of offering to drop it.

Nothing outstanding on this PR: green and up to date with `develop` (head `ac8e5486`, 6/6), CodeRabbit's docstring findings fixed in `7efa3e6`, and no open questions from me. The branch also carries a merge of `develop` after #2690 landed, resolving a real conflict in `tab_quests_approvals.html` where both PRs edited the reply form block: the resolution keeps this PR's per-row form together with #2690's `next` field. Full suite on the merged branch `Ran 2976 tests ... OK`.

---

PR #2690, opened 2026-09-11
fix: post approvals normally instead of in the background (#2689)

### LINKED ISSUE #2687: Stacking messages at the top, don't clear (due to AJAX quest approvals?)
THis will probably be resolved with 2689 automatically

Image

### LINKED ISSUE #2689: Remove the AJAX approvals
The page loads so much faster now, we don't need it and it causes too many other little buggy issues.

tylerecouture:
### What?

Approving, returning, commenting on or skipping a submission from the approvals page now posts the form and reloads the page, instead of posting in the background and patching the page in place.

- the script that intercepted the approve and return buttons is gone from `quest_approval.html`
- `ApproveView` has one response per outcome, with no ajax branches, and `dispatch` no longer sniffs `X-Requested-With`
- the `quests:ajax_approve` route, which existed only for that script, is gone
- each reply form carries `next`, so a decision returns the teacher to the tab they were working through

Closes #2689
Closes #2687

### Why?

You asked for this in #2689, and expected #2687 to fall out of it. It does, and here is the mechanism, since I checked rather than assumed.

Django clears a message once it has been displayed, and displaying happens when a page is rendered. The background post never rendered a page: it fetched the messages as a fragment and the script appended them to `#messages-container`. Nothing removed the previous one, so they accumulated, one per decision, for as long as the teacher stayed on the page. Measured on a real deck, approving three submissions in a row took the alert count from 1 to 4:

```
on load:             1 alert
after approving 61:  2 alerts
after approving 60:  3 alerts
after approving 1:   4 alerts
```

With the page reloading, it stays at one message at a time, and the text changes with each decision:

```
on load:             1 alert
after approving 4:   2 alerts   "Submission of Screenshots approved for ..."
after approving 5:   2 alerts   "Submission of Who owns your creations? approved for ..."
after approving 6:   2 alerts   "Submission of Who owns your creations? approved for ..."
```

(The standing one is the deck's current-student limit banner, which is not a Django message and is supposed to persist.)

### How?

**`next`, because the redirect target matters now.** `form_valid` redirected to `quests:approvals`, which resolves to the my-groups Submitted tab. That was invisible while approve and return never followed the redirect, but with the page reloading, a teacher working through "All", or the In Progress or Flagged tab, would be thrown back to the default tab after every single decision. I hit this in verification: my first run reported "no submissions left" after one approval, because the deck owner is not the teacher for any group and the redirect had landed on an empty queue.

Each reply form now posts `next` with the page it came from. `ApproveView.safe_next()` returns it when it stays on this host, using the same `url_has_allowed_host_and_scheme` guard `CategoryPublish` already uses, and `None` otherwise, so a crafted value cannot turn this into an open redirect. A post without `next` (the full staff form on the submission page) still lands on `quests:approvals`, as before.

`safe_next()` is also what carries the destination across a form that does not validate: `form_invalid` sends the teacher to the submission page to correct it, and the corrected post is a fresh one carrying only what that page puts in it, so the page renders the checked value as a hidden field. Because `safe_next()` returns `None` for an off-host value, such a value is neither followed nor written into the page.

**What is deliberately left alone:** the accordion rows still load their bodies over ajax (`ajax_approval_info` and friends). That is what makes the page load quickly in the first place, and it is not what this issue is about.

### Testing?

Six new tests, plus one rewritten.

**Fail on `develop`:**

- `test_approvals__approve_buttons_post_their_form_normally` (`'ajax_submission_approve' unexpectedly found in the following response`)
- `test_approvals__there_is_no_ajax_only_approve_route` (`NoReverseMatch not raised`)

**Cover the `next` handling, which is new here:**

- `test_approve__returns_to_the_tab_the_decision_was_made_from`
- `test_approve__will_not_be_redirected_off_this_host` (an off-site `next` is ignored, not followed)
- `test_approve__a_corrected_resubmission_still_returns_to_that_tab`
- `test_approve__an_off_host_next_is_not_carried_into_the_correction`

The last two came out of CodeRabbit's review, which spotted that `next` did not survive `form_invalid`. Both fail without the fix; the first with `Couldn't find 'name="next" value="/quests/approvals/submitted/all/"' in the following response`.

`test_approve__ajax_all_button_types` becomes `test_approve__every_button_type_acts_and_returns_to_the_approvals_page`: the same four-button coverage, asserting the redirect rather than a 200 from the ajax branch. `test_approve__invalid_form_ajax_returns_400` is deleted along with the branch it pinned; `test_approve__invalid_form_renders_submission_page` already covers what every caller now gets.

**I could not make a server-side test fail for the stacking itself.** The stacking is what the *browser* was left holding, and the server's response to a plain post was always correct. The two tests above pin the thing that actually changed, and the before/after below is the stacking measured in a real browser.

Full suite `Ran 2972 tests ... OK`, plus `ruff check src`, `manage.py check` and `makemigrations --check --dry-run` all clean.

Real `hackerspace` dev deck, `/quests/approvals/submitted/all/` as the deck owner, approving three submissions one after another.

**Before** (`develop`): three approval messages stacked, pushing the queue down the page.

before

**After**: one message, for the decision just made. Note the "All" tab is still selected, which is the `next` doing its job.

after

### Anything Else?

**One rough edge during the deploy itself.** A teacher who already has the approvals page open is holding the old script, which posts to a url this removes. For the few minutes until they reload, their approve and return buttons will do nothing visible (the background post 404s, so the success handler never runs). Nothing is lost and a reload fixes it, but worth knowing if you deploy this mid-morning.

**The page now reloads on every decision**, which is the trade you asked for. On the dev deck the approvals page serves in about 0.7s with 30 rows on it (the row bodies still arrive over ajax after that). If it turns out to drag when working through a long queue, the fix is to make the page cheaper rather than to bring the background post back.

### Review request
@tylerecouture


tylerecouture:
All three findings were real; fixed in `dd82781f`.

**1. `next` was lost across a form that does not validate.** Confirmed: `form_invalid` renders `submission.html`, whose staff approve form had no `next`, so the corrected post fell back to the default approvals tab. Deciding where "back" is now lives in one place, `ApproveView.safe_next()`: `form_valid` redirects to it, and `form_invalid` passes it to the submission page to render as a hidden field. It returns `None` for anything that does not stay on this host, so an off-host value is neither followed nor written into the page.

Two tests for it, both of which fail without the change:
- `test_approve__a_corrected_resubmission_still_returns_to_that_tab` (`Couldn't find 'name="next" value="/quests/approvals/submitted/all/"' in the following response`)
- `test_approve__an_off_host_next_is_not_carried_into_the_correction`

**2. `dispatch` docstring.** Now documents `request`, `*args`, `**kwargs`, the delegated `HttpResponse`, and both `Http404` cases.

**3. Docstrings describing removed behaviour.** You caught a real leftover I missed: `post()`'s `Returns:` still described the `JsonResponse` and the ajax 400. Corrected, along with the four test docstrings, which now state the invariant in the present tense (why a page that never reloads leaves messages stacked) rather than recounting what the code used to do.

`src/quest_manager` and `src/questions` green (746 tests), `ruff check src` clean. Full suite running.

---

Issue #2729, opened 2026-09-16
One blocking quest empties a student's Available tab with nothing said, while every hidden quest stays startable from the map


tylerecouture:
### What

`QuestQuerySet.block_if_needed()` is what makes a blocking quest block, and the way it does it is total:

```python
def block_if_needed(self, user=None):
    blocking_quests = self.filter(blocking=True)
    if user:
        blocking_subs_in_progress = QuestSubmission.objects.all_not_completed(user=user, blocking=False).filter(quest__blocking=True)
    else:
        blocking_subs_in_progress = False

    if blocking_quests or blocking_subs_in_progress:
        return blocking_quests
    else:
        return self
```

If **any** quest with `blocking=True` is available to a student, or they have one in progress, their Available tab returns only blocking quests. Every other quest they qualify for vanishes from the list.

Two things make this land badly:

* **Nothing says so.** The tab does not explain that a quest is holding everything else back; the other quests are simply gone, which reads as the site losing them.
* **They are still startable.** `Quest.is_available()` never consults blocking, so every one of those quests opens normally from the quest map or its own URL. A student who knows the map can walk straight past the block, and a teacher looking at the same deck sees a student doing a quest that "isn't available to them".

That combination is why a missing quest gets reported as a cache bug: an empty-looking tab plus quests that demonstrably still work is exactly what a stale prerequisite cache looks like from the outside.

Confirmed on a seeded dev deck. A student with two available quests, one of them marked blocking:

```
In the Available tab: 1 quest(s).

Startable but MISSING from the tab: 1 quest(s)
       1  Welcome to ByteDeck!    dropped by: not crowded out by a blocking quest
```

### Worth deciding

1. **Tell the student.** A line on the Available tab naming the blocking quest ("Finish *Read this first* before your other quests appear") turns a disappearance into an instruction. The Hide feature has the same shape and the tab already lists hidden quests, so there is a pattern to follow.
2. **Decide whether starting one should be refused.** Either the block is real, in which case `Quest.is_available()` should honour it and the map link should say why, or it is only a nudge, in which case the tab hiding everything is too strong. Right now it is neither: a hard block on the list and no block at all anywhere else.
3. `Quest.objects.get_available(user, remove_hidden=True, **blocking=True**)` takes a `blocking` parameter it never reads. Whatever is decided above, that parameter should either do something or go.

[...]

PR #2731, opened 2026-09-16
feat: say which quest is holding a student's other quests out of the tab (#2729)

tylerecouture:
### What?

Two things, both about making a blocking quest visible as a blocking quest.

1. The Available tab names the quest that is keeping the student's other quests out of it, and links to it:

   > Your other quests are on hold until you finish **Weekly Journal**. They come back to this tab as soon as it is done.

2. The row's status-icon column marks a blocking quest with a **raised hand** instead of the exclamation triangle it used before, so the row and the notice say the same thing.

Part of #2729. The blocking rules themselves are untouched, for the reason in **Anything Else**.

### Why?

While a quest with `blocking=True` is available to a student, or they have one in progress, `QuestQuerySet.block_if_needed()` reduces the whole tab to blocking quests only. Every other quest they qualify for vanishes from the list.

Two things made that land badly:

* **Nothing said so.** The quests were simply gone, which reads as the tab losing them. The one marker that existed was the row icon, and an exclamation triangle is what this site puts on any caution, so it said "careful" rather than "this one is holding your other quests back".
* **They stayed startable.** `Quest.is_available()` never consults blocking, so all of those quests still open from the quest map or their own URL. A tab that looks empty next to quests that demonstrably still work is exactly what a stale prerequisite cache looks like from the outside, which is why this gets reported as a cache bug.

When a blocking quest is already *in progress* it is not in the tab either, so the tab is completely empty and the old generic line said "You have no new quests available, but you can find some quests you have already started in your 'In Progress' tab above" without connecting the two.

### How?

`quest_list()` collects the quests responsible into a `blocking_quests` context entry: the blocking quests still in `available_quests` (which is already past `block_if_needed`, so a blocking quest still in it is one holding the others back) plus any blocking quest the student has in progress.

Only for a student on the course-registered path. Staff see every published quest unfiltered, and a student with no current course goes through `get_available_without_course()`, which does not block; in both cases nothing is being held back and the notice would be untrue.

The template shows it whether or not the tab still has anything in it, and the generic "no new quests available" branch now gives way to it rather than saying something else alongside. A plain `alert alert-info`, so no new CSS.

The icon change is `fa-exclamation-triangle` to `fa-hand-paper-o`, in the quest tab and in the Shared Library quest list, which carries the same marker. The tooltip follows the notice's wording but stays audience-neutral ("all other quests are on hold until this one is finished"), because staff see the quest tab too and the Library list is browsed by teachers looking at someone else's content.

### Testing?

`BlockingQuestNoticeTests`, 10 tests. Five fail on `develop`:

```
FAIL: test_quest_list__names_the_blocking_quest_that_is_available
FAIL: test_quest_list__names_a_blocking_quest_the_student_has_in_progress
FAIL: test_quest_list__names_every_blocking_quest
FAIL: test_quest_list__reads_as_one_quest_when_only_one_blocks
FAIL: test_quest_list__the_generic_empty_message_gives_way_to_the_notice
AssertionError: False is not true : Couldn't find 'Your other quests are on hold until you finish' in the following response
```

Two more pin the row icon, asserting the row's own markup (`icon-spacing fa fa-fw fa-hand-paper-o`) so they cannot be satisfied by the notice's copy of the same symbol, and that an ordinary quest carries none.

The other three guard what must not change: no notice when nothing is blocking (so it cannot become wallpaper), none for staff, and none for a student with no current course.

Full suite green on the branch (3095 tests, after merging `develop` in), plus `ruff check src` and `makemigrations --check --dry-run` clean.

From a seeded `hackerspace` dev deck, signed in as a student with two quests available to them, one of which is marked blocking. Each pair reads the same database, so only the code differs.

**A blocking quest is available.** Before: their other quest is gone from the tab, with nothing to say why, and the row carries a warning triangle.

Available tab on develop, other quest silently missing

After:

Available tab with the notice naming the blocking quest

**The row's status icons**, close up. Before and after:

status icons on develop: undo and warning triangle
status icons on this branch: undo and raised hand

**The same student once they start that blocking quest**, which is the harsher case: the tab is now completely empty. Before:

empty Available tab on develop

After:

empty Available tab with the notice

The deck was put back as it was afterwards.

### Anything Else?

`develop` is merged in as of `be0340f47`. The conflict was that this branch and #2727 each appended a test class to the end of `quest_manager/tests/test_views.py`, so both are kept, side by side; `views.py` merged cleanly.

This says what is happening; it does not decide whether what is happening is right. Two things stay open on #2729 because they are behaviour decisions, not fixes:

* `Quest.is_available()` does not honour the block, so the map link still works. Either the block is real, and starting one should be refused with a reason, or it is only a nudge, in which case hiding the whole tab is too strong. Right now it is a hard block on the list and no block anywhere else.
* `Quest.objects.get_available(user, remove_hidden=True, blocking=True)` takes a `blocking` parameter it never reads. Whichever way the above goes, that should either do something or go.

### Review request
@tylerecouture

## Summary by CodeRabbit

- **New Features**
  - Added an “on hold” notice to the Available Quests tab when another quest must be completed first.
  - Blocking quests are listed with links to their details, along with guidance that other quests will remain on hold until completion.
  - The notice supports both single and multiple blocking quests.

- **Improvements**
  - Updated blocking-quest indicators with a hand icon and clearer “on hold” wording.
  - The generic “no new quests available” message is hidden when blocking quests are present.


tylerecouture:
@coderabbitai review

Issue #2559, opened 2026-08-20
Student file answers accept SVG/HTML, giving stored XSS against the marker on same-origin media


tylerecouture:
Found in a full audit of the submission questions feature. Related to the old #1022 (closed).

### The problem
`RestrictedFileFormField.validate_file` (`src/utilities/fields.py:323-340`) is the only server-side gate on a student's file answer, and it accepts when `file.content_type in self.content_types`, where `content_type` is the client-declared multipart Content-Type (spoofable). No spoofing is even required: `IMAGE_MIME_TYPES` includes `image/svg+xml` (`fields.py:25`), and the default `allowed_file_type` is `all` (`src/questions/models.py:75`), which accepts any file including `.html`. There is no extension deny-list and no `Content-Disposition: attachment`.

The stored file is linked for the marker in `comment_answers.html` -> `question_file.html` as a plain inline link (`<a href="{{ file.url }}" target="_blank">`). Media is served same-origin under `MEDIA_URL='/media/'` in the default / self-hosted (non-S3) deployment. `SESSION_COOKIE_HTTPONLY` does not stop an XSS payload issuing authenticated same-origin requests as the marker.

### Impact
On a default or self-hosted (same-origin `/media`) deck, a student answering a File Upload question uploads `pwn.svg` containing an inline `<script>` (accepted as `image/svg+xml` with no spoofing; under the default `all` a plain `pwn.html` works too). When a teacher marks the submission and clicks the file link (opens in a new tab, served inline same-origin), the script executes in the deck's origin under the teacher's authenticated session: stored XSS letting an unprivileged student act as a staff marker (or deck owner/superuser). On S3/CDN production the media is a separate origin, which mitigates same-origin session theft there, but self-hosted default deployments are exposed.

### Suggested fix
Reject script-capable types regardless of declared `content_type`: add a server-side extension/type deny-list (at minimum `.svg`, `.html`/`.htm`/`.xhtml`/`.xml`) in `RestrictedFileFormField` or the question submission form, and/or serve user-uploaded media with `Content-Disposition: attachment` (or from a sandboxed cookieless origin). The client-declared content_type MIME check cannot be relied on for security. Note this also protects comment attachments, which share the same field.

- Claude Code (`Bytedeck - multiple submission types`)

tylerecouture:
[...]

Issue #2427, opened 2026-08-14
Comment attachments are silently lost when a submission fails validation


tylerecouture:
### Description

The same problem #2165 fixed for question file answers still applies to the submission form's main **Attach files** field.

When a student attaches files to their submission comment and the submit fails validation (a blank required question, a missing XP value on a custom-XP quest), the page re-renders with the errors. Browsers never repopulate a file input, so the "Attach files" box comes back empty and the uploaded files are gone. Nothing on the page says so: the student either re-attaches them (if they notice) or submits again without the files they thought they had sent.

### Steps to reproduce

1. Give a quest a required submission question (or turn on "XP can be entered by students").
2. As a student, open the quest submission page, attach a file under **Attach files**, and leave the required answer blank.
3. Submit for approval. The page comes back with the validation error, and the attachment is gone.

### Expected

Attachments that passed their own validation are kept across the failed submit, the way question file answers now are (#2165): saved on the submission's draft comment, listed on the re-rendered page so the student can see they survived, and published with the comment when the submission finally goes through. A file rejected on its own merits (over the 16MB limit) is not kept, so its error still applies on the retry.

Follow-up to #2165 / #2425.

- Claude Code (Bytedeck - multiple submission types)

Issue #2692, opened 2026-09-12
SVG answer previews expose the file's storage URL, which is same-origin on non-S3 decks


tylerecouture:
`src/questions/templates/questions/snippets/comment_answers.html` previews a script-capable SVG answer through an `<img>`:

```html
{% if answer.response_file|is_displayable_svg %}
  ``&lt;img class="question-media" src="{{ answer.response_file.url }}"·       alt="{{ answer.response_file|filename|striptags }}"&gt;``
{% endif %}
```

Inside an `<img>` a browser runs no script, which is what the comment there says and it is correct. The issue is that rendering it puts `response_file.url` in the page, and **navigating** to that URL is a different matter: the file is served as `image/svg+xml`, and script inside an SVG does run when the SVG is the document rather than an image.

## Why the origin matters

- **Production (S3):** `MEDIA_URL` is `https://{AWS_S3_CUSTOM_DOMAIN}/...`, so an SVG opened directly runs on the media domain, not on the deck's. It has no access to a deck session cookie. Low risk.
- **Any deck not using S3:** `MEDIA_URL` is `/media/`, served from the deck's own origin. An SVG opened directly runs **same-origin**, with whatever session the viewer has.

`SECURE_CONTENT_TYPE_NOSNIFF` does not help here: `image/svg+xml` is the file's honest content type, and nosniff only stops a browser guessing a different one.

## The path

A student uploads a crafted SVG as a `file_upload` answer. Their teacher opens the submission; the preview renders harmlessly, but the URL is now in the page, one "open image in new tab" away, and can also simply be sent to the teacher. Opening it executes the SVG's script against the teacher's session.

## Not introduced by any open PR

This is on `develop` already, from the SVG preview added in #2559. CodeRabbit raised it as an *outside diff range* comment on PR #2622, which is why it has no review thread to resolve: #2622 does not touch that branch of the template (its only change there is the Add to Portfolio button in the non-script-capable branch). Filing it here so it is not lost when that PR merges.

## Options

- drop the `<img>` preview for SVG, leaving the download link that #2559 already routes through `Content-Disposition: attachment`
- serve answer media through a view that sets `Content-Disposition: attachment` for script-capable types, so the storage URL is never the thing rendered
- render a sanitized raster derivative for the preview
- sanitize the SVG on upload

[...]

PR #2733, opened 2026-09-16
fix: hand a script-capable submission attachment over as a download (#2726)

### LINKED ISSUE #2726: Serve submission comment attachments as downloads, so a web file cannot run its script in the marker's session
Follow-up to the hotfix that lets the submission comment's **Attach files** box take web files again (SVG, HTML, XML, MHTML). Refusing them, which #2559 did site-wide, broke whole cohorts of existing quests that ask students to hand in a page or a vector drawing, so the types are accepted again while the graceful version of the restriction is worked out here.

### What is exposed meanwhile

An attachment on a comment is linked at its storage URL and opens in the page:

```html
<a href="{{ document.docfile.url }}" target="_blank">{{ document.docfile|filename }}</a>
```

Served inline from the app's own origin, an SVG or HTML file runs whatever script it carries in the session of whoever follows the link, which is normally the teacher marking the work. That is stored XSS with the student as the author, which is what #2559 was about.

### The graceful version

A question's file answer already has the shape to copy (`questions.views.answer_file_download`, #2559): the file is stored as usual, and a script-capable one is handed over with `Content-Disposition: attachment` rather than linked directly, so the browser saves it instead of running it. An SVG is still shown as a picture, which is the one way a browser will not run it, so a graphic design answer can still be marked at a glance.

Worth doing for comment attachments too:

* a download view for `Document`, permission-checked the way the comment itself is
* `comments/comments.html` and `quest_manager/snippets/draft_attachments.html` linking script-capable attachments through it rather than at the storage URL
* decide what an SVG attachment should do: rendering it as an image is both safe and more useful than a download

Worth checking at the same time whether anything else links a `Document` at its storage URL (the portfolio "Add to Portfolio" path is the obvious candidate).

tylerecouture:
### What?

A submission comment's SVG, HTML, XML and MHTML attachments are handed to the viewer as downloads instead of being opened in the page. Everything else is linked as before.

The follow-up to #2728, split out so that one could merge on its own: refusing those types broke quests already in use, and students on them could not hand work in at all, so it should not have waited on this.

Closes #2726

### Why?

Attachments are linked at their storage url and open in the page:

```html
<a href="{{ document.docfile.url }}" target="_blank">{{ document.docfile|filename }}</a>
```

Served inline from the app's own origin, an SVG or HTML file runs whatever script it carries in the session of whoever follows the link, normally the teacher marking the work. That is stored XSS with the student as its author, and CodeRabbit flagged it on #2728 (CWE-79).

**The exposure is not created by #2728.** Files of these types are already stored on decks from before the site-wide refusal landed, and they are linked at their storage urls today. This closes it for those as well as for new ones.

### How?

The shape a question's file answer already uses (`questions.views.answer_file_download`, #2559):

* `Document.is_script_capable` answers from the **stored file**, not from whatever rule was in force when it was uploaded. Those rules change, and a change to them must not quietly start linking a stored HTML attachment at its storage url again.
* `Document.get_download_url()` returns the download view for those files and the storage url for everything else, so an image a teacher is marking still opens in a tab.
* `comments:document_download` responds with `Content-Disposition: attachment`.
* The templates that linked `document.docfile.url` link `get_download_url` instead: `comments/comments.html`, and `quest_manager/snippets/draft_attachments.html` for the list of files saved with a draft.

The view checks who is asking. A url carrying a row id is guessable in a way a storage path is not, so an open one would let a classmate walk the ids and pull down everyone's work: that would have been a weakening, introduced by a change meant to harden. It answers to staff, to whoever wrote the comment, to the student whose submission it is (so a teacher's attached example is still theirs to open), and to any signed-in member of the deck on the thread of an announcement **they can see**: `Announcement.objects.get_for_students()`, the same queryset the announcements list shows a student, so a draft, one held back for a later release, an archived one and an expired one are all out.

### Testing?

`DocumentDownloadViewTests`, 17 tests, and **100% statement and branch coverage** of both changed modules:

```
src/comments/models.py     129      0     32      0   100%
src/comments/views.py       50      0     12      0   100%
```

What they pin:

* an SVG and an HTML attachment come back with `Content-Disposition: attachment`
* an image is **not** served this way (404), so marking a photo is unchanged
* `get_download_url()` returns the view for a script-capable file and the storage url for anything else
* the rendered comment thread carries the download url and **not** the storage url, since that template is where the link a teacher clicks actually comes from
* each arm of the access check: the submission's own student, a classmate (404), staff, an announcement everyone can read, an attachment on no comment at all (404), and anonymous (redirected to login)
* an announcement a student cannot see is not theirs to fetch either: draft, unreleased, archived and expired, as four subtests

Full suite green on the branch as it stands (3155 tests), plus `ruff check src`, `manage.py check` and `makemigrations --check --dry-run`.

N/A. The only visible difference is that following the link to an SVG or HTML attachment saves it rather than opening it, which the `Content-Disposition` assertions cover. Nothing about the page's markup changes beyond the href.

### Anything Else?

The branch was cut from `staging` and retargeted to `develop` once #2728 was merged to `develop` rather than through `staging`, so this sits on the same line as the change it follows.

`develop` has been merged in twice, and each time the draft-attachment list was the file to settle:

* at `68769e081`, develop had moved that list into `quest_manager/snippets/draft_attachments.html` (#2427) while this branch had changed the link in the old inline markup. Develop's snippet include is kept, and the snippet's own link points at `get_download_url`.
* at `fe0d54f4b`, develop had rebuilt the same list as the rows of the attachments control (#2749): a `list-group-item` per file, with the name behind a paperclip and the Delete button on the right. Those rows are kept, and the one thing this PR wants from that markup, `get_download_url` in place of the storage url, is applied to them.

Two things this deliberately does not do:

* **It does not hide the stored file.** Media is public on S3 in production, so anyone already holding a storage url can still fetch the object; what changes is that the app no longer hands that url out. That is the same bar #2559 set for question answers, and moving media to a private bucket is a much larger piece of work: filed as #2754.
* **It does not render an SVG attachment as a picture.** A question's file answer does that (an `` will not run an SVG's script), and it would be nicer than a download for a graphic design quest. Worth doing, but it is a display change rather than a security one, so it is not riding along here.

### Review request
@tylerecouture

## Summary by CodeRabbit

* **New Features**
  * Script-capable attachments, such as SVG and HTML files, are now downloaded instead of opened directly in the browser.
  * Added permission-controlled download links for eligible attachments.
  * Draft and comment attachments now use safer download handling.

* **Bug Fixes**
  * Reduced the risk of attachment scripts running in a user’s session.
  * Unauthorized, unavailable, or inaccessible attachment downloads now return appropriate errors.
  * Attachments from unpublished, expired, archived, or otherwise unavailable announcements cannot be downloaded.


tylerecouture:
The announcement finding from the latest review was right, and it is fixed in 9cbb2ed22 (replying here rather than on a thread, since it came in as an outside-the-diff note).

`_may_download()` treated every announcement thread as the whole deck's. Not every announcement is: `announcements.views.list` shows a student `Announcement.objects.get_for_students()`, which leaves out drafts, ones held back for a later release, archived ones and expired ones. This view's url carries a guessable row id, so the open branch would have been a way around exactly what that page withholds, in a change whose whole point is that an id is guessable where a storage path is not.

The branch now asks that same queryset. Staff, the comment's author and the student whose submission it is are unaffected, since they are answered before it.

The old announcement test was making a draft announcement (baker's default), which is what let the hole through unnoticed, so the fixture is now explicitly a released, visible one, and `test_document_download__an_announcement_a_student_cannot_see_is_not_theirs_to_fetch` covers all four of the withheld states as subtests. Each of them fails without the fix. `comments/views.py` and `comments/models.py` are still at 100% statement and branch coverage, and the full suite is green (3155 tests).

---

tylerecouture:
Latest review, both points (again as a comment, since both came in outside the diff):

**The docstring nitpick is fixed** in 5235182ef. "Anyone at all" reads as open to the world; an announcement's attachments are open to anyone signed in to the deck, and `login_required` sends an anonymous visitor to the login page. The docstring says that now.

**The public-storage finding is real, but it is not this PR's to fix**, and this PR's description has said so since it opened, under "It does not hide the stored file". Production runs `USE_S3=1` with `PublicMediaStorage` and a `public-read` ACL, so the object behind any attachment is retrievable by anyone holding its URL, with or without this view. What this PR changes is which url the app hands out, which is what closes the guessable-id hole it opens by existing. Making the object itself private is a different piece of work: it means either serving all media through Django or moving to a private bucket with signed urls, it affects every uploaded file rather than these four types, and it would want its own testing on a deck with S3 configured. The same limit applies to `questions.views.answer_file_download` (#2559), which set this bar first.

Filed as #2754 so it is tracked rather than lost, with the options and their costs, and cross-referenced from #2692 (one specific way a storage url still reaches a page).

CI is green on 5235182ef, and the full suite is green locally (3155 tests).

---

